PR #10, opened 2020-08-12
Improve ranking battle reward

ipdae:
- Drop consumable materials when PVP finished.

### DIFF
--- a/.Lib9c.Tests/Action/RankingBattleTest.cs
+++ b/.Lib9c.Tests/Action/RankingBattleTest.cs
@@ -13,18 +13,13 @@ namespace Lib9c.Tests.Action
     using Nekoyume.TableData;
     using Xunit;
 
-    public class RankingBattleTest : IDisposable
+    public class RankingBattleTest
     {
         private TableSheetsState _tableSheetsState;
 
         public RankingBattleTest()
         {
             _tableSheetsState = TableSheetsImporter.ImportTableSheets();
-        }
-
-        public void Dispose()
-        {
-            _tableSheetsState = null;
         }
 
         [Fact]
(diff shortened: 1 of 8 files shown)


.Lib9c.Tests/Action/RankingBattleTest.cs
```diff
@@ -0,0 +1,99 @@
...
+    public class RankingBattleTest : IDisposable
+    {
+        private TableSheetsState _tableSheetsState;
+
+        public RankingBattleTest()
+        {
+            _tableSheetsState = TableSheetsImporter.ImportTableSheets();
+        }
+
+        public void Dispose()
```
longfin, line 25: I guess it's okay that we skip this `Dispose()`. (`null` assigning doesn't affect other test cases.)

.Lib9c.Tests/Lib9c.Tests.csproj
```diff
@@ -28,6 +28,12 @@
 
   <ItemGroup>
     <Content Include="Data/**/*.*" CopyToOutputDirectory="Always" />
+    <Content Update="Data\WeeklyRewardSheet.csv">
```
longfin, line 31: Are these `<Content>`s needed?

ipdae:
mic check

Lib9c/TableData/TableSheets.cs
```diff
@@ -109,7 +110,15 @@ public static TableSheets FromTableSheetsState(TableSheetsState tableSheetsState
             var tableSheets = new TableSheets();
             tableSheets.InitializeWithTableSheetsState(tableSheetsState);
 
-            Cache.Add(tableSheetsState, tableSheets);
+            // Avoid Exception when run parallel tests
```
longfin, line 113: It seems that we should reconsider the current table sheets caching strategy 😢

PR #11, opened 2020-08-13
Bug fix/consider table patch when unlock world

boscohyun:
## Problem

1. After patching the table related to the world, it is not reflected in the avatar's world information.

### DIFF
--- a/Lib9c/Model/WorldInformation.cs
+++ b/Lib9c/Model/WorldInformation.cs
@@ -402,4 +402,12 @@ namespace Nekoyume.Model
             _worlds[worldId] = new World(world, unlockedAt);
         }
     }
+
+    [Serializable]
+    public class FailedToUnlockWorldException : Exception
+    {
+        public FailedToUnlockWorldException(string message) : base(message)
+        {
+        }
+    }
 }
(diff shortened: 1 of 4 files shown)


Lib9c/Model/WorldInformation.cs
```diff
@@ -351,55 +351,74 @@ public bool TryGetLastClearedStageId(out int stageId)
...
                 return;
             }
 
-            var world = _worlds[worldId];
             _worlds[worldId] = new World(world, unlockedAt);
         }
     }
+
+    [Serializable]
+    public class FailedToUnlockWorldException : Exception
```
longfin, line 418: I guess we need to add `protected FailedToUnlockWorldException(SerializationInfo info, StreamingContext context) : base(info, context)` for .NET Serializable. (see also https://stackoverflow.com/questions/94488/what-is-the-correct-way-to-make-a-custom-net-exception-serializable)

longfin:
@boscohyun The build was [failed](https://github.com/planetarium/lib9c/pull/11/checks?check_run_id=983111574) due to interface changing. I guess we should apply changed interfaces to test project too.

PR #19, opened 2020-08-19
Introduce FungibleAssetValue

longfin:
This PR introduces `FungibleAssetValue` from [libplanet 0.10.0-dev.20200818120248](https://www.nuget.org/packages/Libplanet/0.10.0-dev.20200818120248). (it had been introduced by https://github.com/planetarium/libplanet/pull/954)

### DIFF
--- a/Lib9c/Action/ActionBase.cs
+++ b/Lib9c/Action/ActionBase.cs
@@ -12,6 +12,7 @@ using Libplanet;
 using Libplanet.Action;
 using Serilog;
 using Nekoyume.Model.State;
+using Libplanet.Assets;
 #if UNITY_EDITOR || UNITY_STANDALONE
 using UniRx;
 #else
@@ -36,7 +37,7 @@ namespace Nekoyume.Action
         private struct AccountStateDelta : IAccountStateDelta
         {
             private IImmutableDictionary<Address, IValue> _states;
-            private IImmutableDictionary<(Address, Currency), BigInteger> _balances;
+            private IImmutableDictionary<(Address, Currency), (BigInteger, BigInteger)> _balances;
 
             public IImmutableSet<Address> UpdatedAddresses => _states.Keys.ToImmutableHashSet();
 
@@ -50,7 +51,7 @@ namespace Nekoyume.Action
 
             public AccountStateDelta(
                 IImmutableDictionary<Address, IValue> states,
-                IImmutableDictionary<(Address, Currency), BigInteger> balances
+                IImmutableDictionary<(Address, Currency), (BigInteger, BigInteger)> balances
             )
             {
                 _states = states;
@@ -63,9 +64,16 @@ namespace Nekoyume.Action
                     kv => new Address(kv.Key.EncodeAsByteArray()),
                     kv => kv.Value
                 );
+
+                (BigInteger, BigInteger) Decode(List value) =>
+                    (
+                        value.ElementAt(0).ToBigInteger(),
+                        value.ElementAt(1).ToBigInteger()
+                    );
+
                 _balances = balances.Cast<Dictionary>().ToImmutableDictionary(
                     record => (record["address"].ToAddress(), CurrencyExtensions.Deserialize((Dictionary)record["currency"])),
-                    record => record["amount"].ToBigInteger()
+                    record => Decode((List)record["amount"])
                 );
             }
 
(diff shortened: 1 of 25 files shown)


Lib9c/Action/ActionBase.cs
```diff
@@ -88,86 +98,94 @@ public AccountStateDelta(byte[] bytes)
...
 
+                var nextAmount = GetBalance(recipient, value.Currency) + value;
+
                 return new AccountStateDelta(
                     _states,
-                    _balances.SetItem((recipient, currency),
-                        GetBalance(recipient, currency) + amount)
+                    _balances.SetItem(
+                        (recipient, value.Currency),
+                        (nextAmount.MajorUnit, nextAmount.MinorUnit)
```
dahlia, line 127: `FungibleAssetValue` is serializable.  Isn't it okay to use the `nextAmount` value in itself instead of a pair of `(nextAmount.MajorUnit, nextAmount.MinorUnit)`?

dahlia (commented):
I believe @moreal also should review this, because he made <https://github.com/planetarium/libplanet/pull/950>.

Lib9c/Action/ActionBase.cs
longfin: Right, but it also relies on bencodex's serialization. so I need to write some translation codes like [it](https://github.com/planetarium/lib9c/pull/19/files#diff-01e3b7cdd708218e5c5c45ec04ab9196R256).😢

dahlia: Pair enough.  So would it be useful for this case if we change `FungibleAssetValue.RawValue`'s visibility from `internal` to `public`?

longfin: I guess... yes maybe.

dahlia: Made a pull request for that: <https://github.com/planetarium/libplanet/pull/962>.

Lib9c/Action/BalanceDoesNotExistsException.cs
```diff
@@ -1,4 +1,5 @@
 using Libplanet;
+using Libplanet.Assets;
```
boscohyun, line 2:
```suggestion
```

Lib9c/Action/ActionBase.cs
```diff
@@ -88,86 +96,94 @@ public AccountStateDelta(byte[] bytes)
...
             public IAccountStateDelta TransferAsset(
                 Address sender,
                 Address recipient,
-                Currency currency,
-                BigInteger amount,
+                FungibleAssetValue value,
                 bool allowNegativeBalance = false)
             {
-                if (amount <= 0)
+                if (value.Sign < 0)
```
ipdae, line 136: Is this the intended change?

Lib9c/Model/State/GoldBalanceState.cs
```diff
@@ -1,32 +1,32 @@
...
+        public readonly FungibleAssetValue Gold;
 
-        public GoldBalanceState(Address address) : this(address, 0)
-        {
-        }
-
-        public GoldBalanceState(Address address, BigInteger gold) : base(address) =>
-            this.gold = gold;
+        public GoldBalanceState(Address address, FungibleAssetValue gold) : base(address) =>
+            this.Gold = gold;
```
ipdae, line 15:
```suggestion
            Gold = gold;
```

Lib9c/Action/ActionBase.cs
longfin: Yes, but `value.Sign < 0` isn't same as `amount <= 0` because it allows `0`. so I'll fix it.

Lib9c/Model/State/GoldCurrencyState.cs
```diff
@@ -5,6 +5,7 @@
 using Bencodex;
 using Bencodex.Types;
 using Libplanet;
+using Libplanet.Assets;
```
boscohyun, line 8:
```suggestion
```

Lib9c/Action/BalanceDoesNotExistsException.cs
longfin: It's needed because `Currency` was moved from `Libplanet` to `Libplanet.Assets`.

Lib9c/Model/State/GoldCurrencyState.cs
longfin: Same as https://github.com/planetarium/lib9c/pull/19#discussion_r474360464

Lib9c/Lib9c.csproj
```diff
@@ -15,7 +15,7 @@
   <ItemGroup>
     <PackageReference Include="DecimalMath.DecimalEx" Version="1.0.1" />
     <PackageReference Include="JetBrains.Annotations" Version="2019.1.3" />
-    <PackageReference Include="Libplanet" Version="0.10.0-dev.20200806084644" />
+    <PackageReference Include="Libplanet" Version="0.10.0-dev.20200818120248" />
```
dahlia, line 18: [Libplanet 0.10.0-dev.20200821052050][1] has released!

[1]: https://www.nuget.org/packages/Libplanet/0.10.0-dev.20200821052050

Lib9c/Action/ActionBase.cs
dahlia: [Libplanet 0.10.0-dev.20200821052050][1] has released!

[1]: https://www.nuget.org/packages/Libplanet/0.10.0-dev.20200821052050

moreal (approved):

PR #23, opened 2020-08-25
Implement IRenderer<T> interface

dahlia:
As the upstream Libplanet changed the way to render actions (https://github.com/planetarium/libplanet/pull/963), I made a patch for lib9c to follow up these changes.

I am going to send a pull request to catch up the same API changes to NineChronicles.Standalone repository as well.

### DIFF
--- a/Lib9c/Action/ActionRenderer.cs
+++ b/Lib9c/Action/ActionRenderer.cs
@@ -1,5 +1,8 @@
 using System;
 using Libplanet;
+using Libplanet.Action;
+using Libplanet.Blockchain.Renderers;
+using Libplanet.Blocks;
 using static Nekoyume.Action.ActionBase;
 #if UNITY_EDITOR || UNITY_STANDALONE
 using UniRx;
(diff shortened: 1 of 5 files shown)


longfin (approved):

riemannulus (approved):

limebell (approved):

earlbread (approved):

Lib9c/Action/ActionRenderer.cs
```diff
@@ -10,25 +13,124 @@
 
 namespace Nekoyume.Action
 {
-    public class ActionRenderer
+    using Action = PolymorphicAction<ActionBase>;
```
moreal, line 16: I believe it may cause confused it with `System.Action`. How about using another alias name like `ActionType` or a better one not confused?

dahlia: @moreal Renamed to `NCAction`: 43f06784c15bad1538451a5e96288bdb01d4c571.

moreal (approved):

ipdae (approved):

PR #28, opened 2020-08-27
Fix bug/Not equal updated addresses in buy action

boscohyun:
### Problem

1. In the Buy action, the updated addresses between the rehearsal and the main run are different, so the TX is invalidated.


longfin (approved):

longfin:
I guess we need the unittest for `Buy` action. 😅

dahlia (approved):

moreal (approved):

limebell (approved):

PR #32, opened 2020-08-28
Add new WeeklyArenaState.GetArenaInfos() method

boscohyun:
### Needs

1. Get a certain number of ranking information from the starting rank.


Lib9c/Model/State/WeeklyArenaState.cs
```diff
@@ -112,7 +112,34 @@ private void ResetOrderedArenaInfos()
         }
 
         /// <summary>
-        /// 인자로 넘겨 받은 `avatarAddress`를 기준으로 상위와 하위 범위에 해당하는 랭킹 정보를 얻습니다.
+        /// Get arena rank information.
+        /// </summary>
+        /// <param name="firstRank"></param>
+        /// <param name="count"></param>
+        /// <returns></returns>
```
limebell, line 119:
```suggestion
        /// <param name="firstRank">Some text describes this parameter</param>
        /// <param name="count">Some text describes this parameter</param>
        /// <returns>A list of tuples that contains <c>int</c> and <c>ArenaInfo</c>.</returns>
```

ipdae (approved):

Lib9c/Model/State/WeeklyArenaState.cs
```diff
@@ -112,7 +112,34 @@ private void ResetOrderedArenaInfos()
...
+        public List<(int rank, ArenaInfo arenaInfo)> GetArenaInfos(
+            int firstRank = 1,
+            int? count = null)
+        {
+            if (firstRank < 1 ||
+                firstRank > OrderedArenaInfos.Count)
+            {
+                throw new ArgumentException(
```
longfin, line 127:
```suggestion
                throw new ArgumentOutOfRangeException(
```

Lib9c/Model/State/WeeklyArenaState.cs
```diff
@@ -112,7 +112,34 @@ private void ResetOrderedArenaInfos()
...
+        /// </summary>
+        /// <param name="firstRank"></param>
+        /// <param name="count"></param>
+        /// <returns></returns>
+        public List<(int rank, ArenaInfo arenaInfo)> GetArenaInfos(
+            int firstRank = 1,
+            int? count = null)
+        {
+            if (firstRank < 1 ||
+                firstRank > OrderedArenaInfos.Count)
```
longfin, line 125: it's a quite more readable.
```suggestion
            if (!(0 < firstRank && firstRank <= OrderedArenaInfos.Count))
```

longfin (approved):
It would be better if we have a testcase for `.GetArenaInfo()`.

Lib9c/Model/State/WeeklyArenaState.cs
boscohyun: Oh, too, when I checked the range, I preferred to write it in that structure that is visually more readable.
However, I have previously accepted the opinion that it is better to have the object to be examined on the left side.
The latter feels like a convention, but I think the visually readable structure is better.
In the future, I will write it in a structure that is visually easy to read.

riemannulus (approved):

boscohyun:
> It would be better if we have a testcase for `.GetArenaInfo()`.

@longfin Can you please check the unit test codes?

.Lib9c.Tests/Model/State/WeeklyArenaStateTest.cs
```diff
@@ -110,5 +110,111 @@ public void ArenaInfoGetRewardCount(int minScore, int maxScore, int expected)
...
+        [Theory]
+        [InlineData(100, 1, 100)]
+        [InlineData(100, 51, 50)]
+        [InlineData(10, 1, 50)]
+        [InlineData(10, 1, 1)]
+        [InlineData(10, 6, 50)]
+        [InlineData(10, 6, 1)]
+        public void GetArenaInfos_By_FirstRank_And_Count(int infoCount, int firstRank, int count)
+        {
+            var expected = Math.Min(infoCount - firstRank + 1, count);
```
longfin, line 123: I guess it would be better if specifying expected value directly (instead of calculation) like other fixture(e.g. `infoCount`) for readability of this test case.

.Lib9c.Tests/Model/State/WeeklyArenaStateTest.cs
```diff
@@ -110,5 +110,111 @@ public void ArenaInfoGetRewardCount(int minScore, int maxScore, int expected)
...
+            {
+                upperRange = targetRank - 1;
+            }
+
+            if (targetRank + lowerRange >= infoCount)
+            {
+                lowerRange = infoCount - targetRank;
+            }
+
+            var expected = Math.Min(upperRange + lowerRange + 1, infoCount);
```
longfin, line 216: I guess it would be better if specifying expected value directly (instead of calculation) like other fixture(e.g. infoCount) for readability of this test case.

longfin (approved):
I left a comment, but it seems to be okay as it is now.

boscohyun:
> I left a comment, but it seems to be okay as it is now.

@longfin 
I also put the expected value myself at first. However, I made a mistake that was wrong with the expected value.
That's why I struggled between letting the expected value be entered and viewed directly and easily adding the number of different inline cases.
As a result, I thought it would be better to easily add various inputs rather than visually seeing the expected values.
But after confirming your suggestion, I thought again.
This means that unit tests are important to test a number of different cases, but it is also important to check a given expected value for a given input value.
What do you think?
This dialogue will be helpful when writing unit tests in the future.

longfin:
> > I left a comment, but it seems to be okay as it is now.
> 
> @longfin
> I also put the expected value myself at first. However, I made a mistake that was wrong with the expected value.
> That's why I struggled between letting the expected value be entered and viewed directly and easily adding the number of different inline cases.
> As a result, I thought it would be better to easily add various inputs rather than visually seeing the expected values.
> But after confirming your suggestion, I thought again.
> This means that unit tests are important to test a number of different cases, but it is also important to check a given expected value for a given input value.
> What do you think?
> This dialogue will be helpful when writing unit tests in the future.

I agree that it is easy to make mistakes when new values are added, but But in general, this doesn't happen very often, and I thought it wasn't a big issue because it could also be validated through written test case as well. so, rather than reducing mistakes when adding, it's important to make it easier to recognize the intended value.

longfin (approved):

PR #37, opened 2020-09-01
Separate BlockRenderer from ActionRenderer

earlbread:
This separates `BlockRenderer` from `ActionRenderer` and changes to use `BlockRenderer` to update `ActivationSet`.


dahlia (approved):

earlbread:
I changed the way `ActivationSet` is updated to not render. Please take a look. @planetarium/libplanet

Lib9c/BlockPolicy.cs
```diff
@@ -60,7 +66,10 @@ public static IBlockPolicy<PolymorphicAction<ActionBase>> GetPolicy(int minimumD
 #endif
         }
 
-        public static ActionRenderer GetRenderer() => ActionRenderer;
+        public static IRenderer<NCAction> GetRenderer(bool actionRender) =>
```
dahlia, line 69: Returning it as an abstract type `IRenderer<T>` instead of a concrete type (`BlockRenderer` or `ActionRenderer`) makes client code difficult to use its concrete methods like `EveryBlock()` or `EveryRender()`…

How about making `BlockRenderer` and `ActionRenderer` fields `public`, instead of having `bool actionRender` flag?

earlbread: That would be better. Changed it.

Lib9c/BlockPolicy.cs
```diff
@@ -10,32 +10,38 @@
...
 
-            ActionRenderer
-                .EveryUnrender(ActivatedAccountsState.Address)
-                .Subscribe(UpdateActivationSet);
+            BlockRenderer
+                .EveryReorg()
+                .Subscribe(_ => UpdateActivationSet());
         }
 
+        public static Func<IValue> ActivatedAccountsStateGetter { get; set; }
```
longfin, line 43: We'll change this class to non-static at #40 . is it okay to add it?

Lib9c/BlockPolicy.cs
```diff
@@ -60,8 +66,6 @@ public static IBlockPolicy<PolymorphicAction<ActionBase>> GetPolicy(int minimumD
 #endif
         }
 
-        public static ActionRenderer GetRenderer() => ActionRenderer;
```
dahlia, line 63: It might be better to have a method to get two kinds of renderers like:

```suggestion
        public static IEnumerable<IRenderer<ActionBase>> GetRenderers() => new[] { BlockRenderer, ActionRenderer };
```

… so that client code can pass it to the `BlockChain<ActionBase>()` constructor:

```csharp
new BlockChain<ActionBase>(…, renderers: BlockPolicy.GetRenderers());
```

earlbread:
I combined this PR with #40. Please review this again.

longfin (approved):

moreal (approved):

Lib9c/BlockPolicySource.cs
```diff
@@ -0,0 +1,95 @@
...
+        public Func<IValue> ActivatedAccountsStateGetter { get; set; }
+
+        public IImmutableSet<Address> ActivatedAccounts { get; private set; }
+
+        public void UpdateActivationSet(IValue state)
+        {
+            ActivatedAccounts = new ActivatedAccountsState((Dictionary)state).Accounts;
+        }
+
+        // FIXME 남은 설정들도 설정화 해야 할지도?
```
riemannulus, line 54: Maybe we need translate this comment 🙂

PR #39, opened 2020-09-02
WIP: Serialize int values into Bencodex.Types.Integer instead of Text

boscohyun:
- Fix serialization of `int` to `Bencodex.Types.Integer` from `Bencodex.Types.Text` in `StateExtensions`

### DIFF
--- a/Lib9c/Model/State/StateExtensions.cs
+++ b/Lib9c/Model/State/StateExtensions.cs
@@ -88,17 +88,13 @@ namespace Nekoyume.Model.State
 
         #endregion
 
-        //FIXME (Text) 대신 (Integer) 로 직렬화해야함
         #region Integer
 
-        public static IValue Serialize(this int number) =>
-            (Text)number.ToString(CultureInfo.InvariantCulture);
+        public static IValue Serialize(this int number) => (Integer) number;
 
-        public static IValue Serialize(this int? number) =>
-            Serialize(Serialize, number);
+        public static IValue Serialize(this int? number) => Serialize(Serialize, number);
 
-        public static int ToInteger(this IValue serialized) =>
-            int.Parse(((Text)serialized).Value, CultureInfo.InvariantCulture);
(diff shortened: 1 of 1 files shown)


Lib9c/Model/State/StateExtensions.cs
```diff
@@ -88,17 +88,13 @@ public static HashSet<T> ToHashSet<T>(this IValue serialized, Func<IValue, T> de
 
         #endregion
 
-        //FIXME (Text) 대신 (Integer) 로 직렬화해야함
```
longfin, line 91: Is it okay that we don't care about this comment now?

longfin (changes requested):
Tests seemed to [fail](https://github.com/planetarium/lib9c/pull/39/checks?check_run_id=1059385011#step:6:484). 😢

Lib9c/Model/State/StateExtensions.cs
boscohyun: I think this project was already in a state where we didn't have to worry about this comment.
However, since this change required a chain reset, I was worried about when to do it.
So, I am looking forward to putting up PR first and merging them at the right time.

boscohyun:
@libplanet @planetarium/nine-chronicles-engineers 

image

Is it possible to make it possible to assign `Bencodex.Types.Integer` to `Bencodex.Types.IKey`?
If that's difficult, I'll convert `Bencodex.Types.Integer` back to `Bencodex.Types.Text` and put it in `IKey`. If this is possible.

ipdae:
`IKey` can't set `Integer` type

ipdae (commented):
in my opinion, it would be better to add a test case.

boscohyun:
> in my opinion, it would be better to add a test case.

I agree. I've come to see that there are many more changes this PR should cover. I will proceed slowly.

dahlia (approved):

limebell:
Some tests seems to fail 😥

boscohyun:
Close stale pr.

PR #42, opened 2020-09-03
Divide TableSheets 2nd

### LINKED ISSUE #41: Delete TableSheetsState dependency
- #36 follow up
- delete TableSheetsState when generate genesis block

ipdae:
- resolve #41 
- Use abstracted class instead of sheets


Lib9c/Action/PatchTableSheet.cs
```diff
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Immutable;
+using System.Net.Http.Headers;
```
longfin, line 3: I guess we don't need this `using`.

longfin (approved):

Lib9c/Action/PatchTableSheet.cs
ipdae: 😅

longfin (approved):

boscohyun (approved):

Unengine (approved):

PR #47, opened 2020-09-07
Fix notification condition on equipment inventory.

Unengine:
- Fix condition of notification in inventory.
- Rings can be equipped more than one, so case is seperated when checking equipments in inventory.


boscohyun (approved):

ipdae (commented):
I think we need to add a condition to player equipment slot is open.

ipdae (approved):

PR #51, opened 2020-09-09
Use LoggedRenderers

limebell:
According to this change, a `logger` should be given when creating `BlockPolicySource` instance. (i.e. `Log.Logger`, `Logger.None`)

### DIFF
--- a/.Lib9c.Tests/BlockPolicyTest.cs
+++ b/.Lib9c.Tests/BlockPolicyTest.cs
@@ -8,7 +8,6 @@ namespace Lib9c.Tests
     using Libplanet.Assets;
     using Libplanet.Blockchain;
     using Libplanet.Blockchain.Policies;
-    using Libplanet.Blockchain.Renderers;
     using Libplanet.Blocks;
     using Libplanet.Crypto;
     using Libplanet.Store;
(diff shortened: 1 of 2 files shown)


dahlia (commented):
If an app uses only `BlockRenderer` events won't be logged.  Would it be okay?

limebell:
> If an app uses only `BlockRenderer` events won't be logged. Would it be okay?

If the app uses only `BlockRenderer` then no actions are rendered. So it is natural that the actions are not be logged.

dahlia:
> If the app uses only `BlockRenderer` then no actions are rendered. So it is natural that the actions are not be logged.

IMHO, however, it would be more useful if we log `RenderBloc()`/`RenderReorg()` events as well as `{Render,Unrender}Action{,Error}()` events… 🤔

limebell:
> IMHO, however, it would be more useful if we log `RenderBloc()`/`RenderReorg()` events as well as `{Render,Unrender}Action{,Error}()` events… 🤔

I've amended your suggestion! Please review again. @dahlia

dahlia (approved):

earlbread (approved):

PR #59, opened 2020-09-11
Define exceptions in HackAndSlash

ipdae:
- Use defined exceptions instead of  `InvalidOperationException`

### DIFF
--- a/Lib9c/Model/WorldInformation.cs
+++ b/Lib9c/Model/WorldInformation.cs
@@ -240,24 +240,16 @@ namespace Nekoyume.Model
             return true;
         }
 
-        public bool TryUpdateWorld(WorldSheet.Row worldRow, out World world)
+        public void UpdateWorld(WorldSheet.Row worldRow)
         {
-            if (worldRow is null ||
-                !_worlds.ContainsKey(worldRow.Id))
-            {
-                world = default;
-                return false;
-            }
-
             var originWorld = _worlds[worldRow.Id];
-            world = new World(
+            var world = new World(
                 worldRow,
                 originWorld.UnlockedBlockIndex,
                 originWorld.StageClearedBlockIndex,
                 originWorld.StageClearedId
             );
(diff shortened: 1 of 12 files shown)


longfin (approved):

Lib9c/Model/WorldInformation.cs
```diff
@@ -240,24 +240,16 @@ public bool TryAddWorld(WorldSheet.Row worldRow, out World world)
             return true;
         }
 
-        public bool TryUpdateWorld(WorldSheet.Row worldRow, out World world)
+        public void UpdateWorld(WorldSheet.Row worldRow)
         {
-            if (worldRow is null ||
```
boscohyun, line 245: Why did you remove the `null` check?

Lib9c/Model/WorldInformation.cs
```diff
@@ -397,6 +389,21 @@ public bool TryGetLastClearedStageId(out int stageId)
...
 
+        public void AddNewWorld(WorldSheet.Row worldRow, long unlockedAt, WorldSheet worldSheet)
+        {
+            var worldId = worldRow.Id;
+            if (IsStageCleared(worldRow.StageBegin - 1))
+            {
+                var world = new World(worldRow);
+                _worlds.Add(worldId, world);
+                UnlockWorld(worldId, unlockedAt, worldSheet);
+            }
```
boscohyun, line 400: If it was to unlock while adding `world`, it would have been better to have a detailed name like `AddNewWorldAndUnlock()`. However, I can guess because there is `unlockedAt` in the argument.

boscohyun (approved):

Lib9c/Model/WorldInformation.cs
ipdae: `UpdateWorld` called only in `HackAndSlash`, and null check before call `UpdateWorld`.

PR #60, opened 2020-09-11
Introduce AuthorizedMinersState

longfin:
This PR allows adding restriction for block miners as `AuthorizedMinersState` on the genesis block.

### DIFF
--- a/Lib9c/BlockPolicySource.cs
+++ b/Lib9c/BlockPolicySource.cs
@@ -58,24 +58,29 @@ namespace Nekoyume.BlockChain
 
         public IImmutableSet<Address> ActivatedAccounts { get; private set; }
 
+        public AuthorizedMinersState AuthorizedMinersState { get; set; }
+
         public void UpdateActivationSet(IValue state)
         {
             ActivatedAccounts = new ActivatedAccountsState((Dictionary)state).Accounts;
         }
 
         // FIXME 남은 설정들도 설정화 해야 할지도?
-        public IBlockPolicy<PolymorphicAction<ActionBase>> GetPolicy(int minimumDifficulty)
+        public IBlockPolicy<NCAction> GetPolicy(int minimumDifficulty)
         {
             ActivatedAccounts = ActivatedAccounts?.Clear();
 #if UNITY_EDITOR
             return new DebugPolicy();
 #else
-            return new BlockPolicy<PolymorphicAction<ActionBase>>(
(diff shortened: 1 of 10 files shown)


Lib9c/BlockPolicySource.cs
```diff
@@ -103,5 +108,29 @@ private void UpdateActivationSet()
...
+                return null;
+            }
+
+            bool targetBlock =
+                0 < block.Index && block.Index <= AuthorizedMinersState.ValidUntil &&
+                block.Index % AuthorizedMinersState.Interval == 0 && 
+                block.Miner is Address miner;
+            bool minedByAuthorities = AuthorizedMinersState.Miners.Contains(miner);
+            
+            if (targetBlock && !minedByAuthorities)
```
dahlia, line 125: Due to short-circuit evaluation, there seems a case that `miner` is uninitialized.  Though the whole logic would work as intended whether `miner` is initialized or not, it's quite complicated at first glance.  It would be better if the whole condition is rewritten into nested `if` blocks IMHO.

dahlia (approved):

Lib9c/BlockPolicySource.cs
```diff
@@ -103,5 +108,29 @@ private void UpdateActivationSet()
...
+        {
+            if (AuthorizedMinersState is null)
+            {
+                return null;
+            }
+
+            bool targetBlock =
+                0 < block.Index && block.Index <= AuthorizedMinersState.ValidUntil &&
+                block.Index % AuthorizedMinersState.Interval == 0 && 
+                block.Miner is Address miner;
```
riemannulus, line 122: how about...
```suggestion
            bool isBlockIndexValid = 0 < block.Index && block.Index <= AuthorizedMinersState.ValidUntil 
            bool isBlockInterval = block.Index % AuthorizedMinersState.Interval == 0 
            bool isMinerValid = block.Miner is Address miner;
            bool isTargetBlock = isBlockIndexValid && isBlockInterval && isMinerValid
```

riemannulus (approved):

Lib9c/BlockPolicySource.cs
longfin: I extracted `block.Miner` checking as separated `if` clause. PTAL 🙇

dahlia (approved):

moreal (approved):

PR #61, opened 2020-09-11
Publish Lib9c to NuGet

moreal:
It makes Lib9c available on NuGet and users can use Lib9c as a package.

You can see GitHub Actions logs and the package on NuGet.

https://github.com/moreal/lib9c/runs/1101605761?check_suite_focus=true
https://www.nuget.org/packages/Lib9c

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -9,8 +9,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
-      with:
-        submodules: recursive
     - uses: actions/setup-dotnet@v1
       with:
         dotnet-version: '3.1.301'
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,35 @@
+name: publish
+
+on:
+  push:
+    branches: main
+
+jobs:
+  nuget:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


longfin (approved):

ipdae (approved):

earlbread (approved):

.github/workflows/publish.yml
```diff
@@ -0,0 +1,35 @@
...
+  push:
+    branches: main
+
+jobs:
+  nuget:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        submodules: recursive
```
limebell, line 13: We don't have any submodules in `Lib9c` yet. Let's leave it as default. (which is `false`)

riemannulus (approved):

.github/workflows/publish.yml
moreal: I amended it in 32193c13f273b578d1d39b4f02fabe0399c7ab3b.

dahlia (approved):

limebell (approved):

boscohyun (approved):

longfin:
@moreal Could you rebase it?

moreal:
> 
> 
> @moreal Could you rebase it?

@longfin I rebased. If there is no issue, it seems able to merge.

PR #65, opened 2020-09-13
Introduce Lib9c.Tools

riemannulus:
- For now, only exists `create` command. it creates genesis block and activation keys.
(eg. `create -g TableCSV -d GoldDistribution.csv -a 10`)


Lib9c.Tools/Program.cs
```diff
@@ -0,0 +1,14 @@
...
+using Lib9c.Tools.SubCommend;
+
+namespace Lib9c.Tools
+{
+    [HasSubCommands(typeof(Create), Description = "Create commands")]
+    class Program
+    {
+        static void Main(string[] args) => CoconaLiteApp.Run<Program>(args);
+
+        public void Info() => Console.WriteLine("Show information");
```
moreal, line 12: This line seems not fully implemented yet, in code. If it is, how about leaving *FIXME* comment?

Lib9c.Tools/SubCommend/Create.cs
```diff
@@ -0,0 +1,37 @@
...
+                out List<ActivationKey> activationKeys,
+                activationKeyCount);
+            GoldDistribution[] goldDistributions = GoldDistribution
+                .LoadInDescendingEndBlockOrder(goldDistributedPath);
+            Block<PolymorphicAction<ActionBase>> block = BlockHelper.MineGenesisBlock(
+                tableSheets,
+                goldDistributions,
+                pendingActivationStates.ToArray());
+            
+            Utils.ExportBlock(block, "genesis-block");
```
longfin, line 33: How about moving export functions to this class? I guess it doesn't seem to be needed from the other class.

longfin (approved):

Lib9c.Tools/Utils.cs
```diff
@@ -0,0 +1,64 @@
...
+        public static Dictionary<string, string> ImportSheets(string dir)
+        {
+            var sheets = new Dictionary<string, string>();
+            var files = Directory.GetFiles(dir, "*.csv", SearchOption.AllDirectories);
+            foreach (var filePath in files)
+            {
+                var fileName = Path.GetFileName(filePath);
+                if (fileName.EndsWith(".csv"))
+                {
+                    fileName = fileName.Split(".csv")[0];
```
moreal, line 36: I wonder why to split it with *.csv*. If the purpose was to remove the extension, *.csv*, then how about using [`String.Substring`][string-substring] or [`String.Remove`][string-remove]? It is not a comment for performance, there seems [some benchmark posted in StackOverflow by user][some-benchmark].

[some-benchmark]: https://stackoverflow.com/a/54896713
[string-remove]: https://docs.microsoft.com/en-us/dotnet/api/system.string.remove?view=netcore-3.1
[string-substring]: https://docs.microsoft.com/en-us/dotnet/api/system.string.substring?view=netcore-3.1

Lib9c.Tools/Program.cs
moreal, line 12: This line seems not fully implemented yet, in code. If it is, how about leaving *FIXME* comment?

Lib9c.Tools/Utils.cs
earlbread: How about using `Path.GetFileNameWithoutExtension(filePath);`?

earlbread (approved):

Lib9c.Tools/Utils.cs
riemannulus: I'll use `Path.GetFileNameWithoutExtension(filePath);`

Lib9c.Tools/SubCommend/Create.cs
riemannulus: amended by 6c42228

Lib9c.Tools/Program.cs
riemannulus: amended by 8e33b90

Lib9c.Tools/Utils.cs
riemannulus: amended by 9ac8f6e

longfin (approved):

riemannulus:
@moreal I'll amended review all. please check out once more!

moreal (approved):

PR #68, opened 2020-09-15
Activate admin address if `isActivateAdminAddress` is true

riemannulus:
- Activate admin address if `isActivateAdminAddress` is true

### DIFF
--- a/.Lib9c.Tools/SubCommend/Create.cs
+++ b/.Lib9c.Tools/SubCommend/Create.cs
@@ -18,7 +18,7 @@ namespace Lib9c.Tools.SubCommend
         public void Genesis(
             [Option('g')] string gameConfigDir,
             [Option('d')] string goldDistributedPath,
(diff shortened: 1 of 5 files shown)


.Lib9c.Tools/SubCommend/Create.cs
```diff
@@ -18,7 +18,7 @@ public class Create
         public void Genesis(
             [Option('g')] string gameConfigDir,
             [Option('d')] string goldDistributedPath,
-            [Option('a')] int activationKeyCount)
+            [Option('c')] int activationKeyCount)
```
limebell, line 21: Is this change necessary? If so, how about separate commit for this change?

riemannulus: Maybe it was my fault. I'll revert it. 😓

earlbread (approved):

limebell (approved):

PR #74, opened 2020-09-16
Fix required block index in ItemEnhancement

boscohyun:
- It reflects the need to use 1 block to enhance equipment.

### DIFF
--- a/Lib9c/Action/ItemEnhancement.cs
+++ b/Lib9c/Action/ItemEnhancement.cs
@@ -20,6 +20,8 @@ namespace Nekoyume.Action
     [ActionType("item_enhancement")]
     public class ItemEnhancement : GameAction
     {
+        public const int RequiredBlockCount = 1;
+
         public static readonly Address BlacksmithAddress = Addresses.Blacksmith;
 
         public Guid itemId;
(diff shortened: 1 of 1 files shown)


longfin (approved):

Lib9c/Action/ItemEnhancement.cs
```diff
@@ -268,6 +268,9 @@ public override IAccountStateDelta Execute(IActionContext context)
             enhancementEquipment.Unequip();
 
             enhancementEquipment = UpgradeEquipment(enhancementEquipment);
+
+            var requiredBlockIndex = ctx.BlockIndex + 1;
```
limebell, line 272: How about defining a new constant such as `requiredBlocks` and set the value as 1 for further changes?

ipdae (commented):
It seems that there is no need to increase the required block value because the situation from the time when item enhancement is performed until the action is executed is viewed as 1 block.

boscohyun:
> It seems that there is no need to increase the required block value because the situation from the time when item enhancement is performed until the action is executed is viewed as 1 block.

@ipdae As you might think, it works fine in the end. But I thought it was a necessary change in the process. Because it is an element of the design of equipment enhancement.

Unengine (approved):

PR #80, opened 2020-09-18
Throw exceptions in RankingBattle action

boscohyun:
The title is the content.


ipdae (commented):
It would be nice to add a test case.

Lib9c/Action/RankingBattle.cs
```diff
@@ -78,31 +74,27 @@ public override IAccountStateDelta Execute(IActionContext context)
...
-                    context,
-                    "Aborted as the avatar state of the opponent ({OpponentAddress}) was failed to load.",
-                    EnemyAddress
-                );
+                throw new FailedLoadStateException($"Aborted as the avatar state of the opponent ({EnemyAddress}) was failed to load.");
             }
 
             var weeklyArenaState = states.GetWeeklyArenaState(WeeklyArenaAddress);
 
             //FIXME 오류던지게 고쳐야함
```
limebell, line 82: We can remove this FIXME :)

Lib9c/Action/NotEnoughFungibleAssetValueException.cs
```diff
@@ -0,0 +1,57 @@
...
+    [Serializable]
+    public class NotEnoughFungibleAssetValueException : Exception
+    {
+        public NotEnoughFungibleAssetValueException(string message) : base(message)
+        {
+        }
+
+        public NotEnoughFungibleAssetValueException(string require, string current)
+            : this(
+                $"Aborted as the signer's balance ({current}) is insufficient to pay entrance fee/stake ({require}).")
```
limebell, line 18: How about this for consistency?
```suggestion
                $"Aborted as the signer's balance is insufficient to pay entrance fee/stake: {current} < {require}.")
```

limebell (approved):

Unengine (approved):

ipdae (approved):

Lib9c/Action/RankingBattle.cs
```diff
@@ -78,36 +73,39 @@ public override IAccountStateDelta Execute(IActionContext context)
...
 
             if (!arenaInfo.Active)
             {
-                FungibleAssetValue agentBalance = states.GetBalance(ctx.Signer, states.GetGoldCurrency());
+                FungibleAssetValue agentBalance = default;
+                try
+                {
+                    agentBalance = states.GetBalance(ctx.Signer, states.GetGoldCurrency());
+                }
+                catch (InvalidOperationException)
```
longfin, line 105: (it isn't critical, but just checking)

Does `states.GetBalance()` throw `InvalidOperationException` ? according to [docs](https://docs.libplanet.io/master/api/Libplanet.Action.IAccountStateDelta.html#Libplanet_Action_IAccountStateDelta_GetBalance_Libplanet_Address_Libplanet_Currency_), it doesn't seem to throw any exceptions.

cc @planetarium/libplanet

boscohyun: Found during testing. If I put an address that has never received `Balance` in `states.GetBalance()` as the first parameter, `InvalidOperationException ` occurred.

PR #88, opened 2020-09-23
Add tests to prove actions are idempotent

moreal:
I tried to prove actions' idempotency by executing every action twice and comparing two. If you thought it is not enough, then please leave your opinion. 🙏 

### Actions

- ActivateAccount
- AddActivatedAccount
- ChargeActionPoint
- CombinationConsumable
- CombinationEquipment
- CreateAvatar
- CreatePendingActivation
- HackAndSlash
- InitializeStates
- ItemEnhancement
- RankingBattle
- RapidCombination
- RedeemCode
- TransferAsset

### DIFF
--- a/.Lib9c.Tests/Action/ActionExecutionUtils.cs
+++ b/.Lib9c.Tests/Action/ActionExecutionUtils.cs
@@ -1,0 +1,46 @@
+namespace Lib9c.Tests.Action
+{
+    using System.Collections.Immutable;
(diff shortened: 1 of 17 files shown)


dahlia:
It seems you mean determinism rather than idempotency.

moreal:
Temporary it will be closed because another pull requests to validate actions are deterministic has opened now or will be opened.

PR #92, opened 2020-09-24
Bug fix/make list deterministic

boscohyun:
- When serializing an object that implements `IState`, it sorts the List<T> fields that the object contains.
- Change type of `DeletedAvatarState.deletedAt` from `DateTimeOffset` to `long`.

### DIFF
--- a/Lib9c/Model/Item/Consumable.cs
+++ b/Lib9c/Model/Item/Consumable.cs
@@ -31,7 +31,10 @@ namespace Nekoyume.Model.Item
         public override IValue Serialize() =>
             new Dictionary(new Dictionary<IKey, IValue>
             {
-                [(Text) "stats"] = new List(Stats.Select(s => s.Serialize())),
+                [(Text) "stats"] = new List(Stats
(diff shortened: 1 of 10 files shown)


Lib9c/Model/Item/Consumable.cs
```diff
@@ -31,7 +31,10 @@ public Consumable(Dictionary serialized) : base(serialized)
         public override IValue Serialize() =>
             new Dictionary(new Dictionary<IKey, IValue>
             {
-                [(Text) "stats"] = new List(Stats.Select(s => s.Serialize())),
+                [(Text) "stats"] = new List(Stats
+                    .OrderBy(i => i.StatType)
```
ipdae, line 35: Is any `IComparer<StatType>` defined?

ipdae: i found this. https://docs.microsoft.com/ko-kr/dotnet/api/system.icomparable?view=netcore-2.0

ipdae (approved):

longfin (approved):

PR #94, opened 2020-09-24
Fix bug in WeeklyArenaState

boscohyun:
- Consider the case where there is no `ArenaInfo` registered in the `WeeklyArenaState`.

### DIFF
--- a/Lib9c/Model/State/WeeklyArenaState.cs
+++ b/Lib9c/Model/State/WeeklyArenaState.cs
@@ -100,6 +100,11 @@ namespace Nekoyume.Model.State
             int firstRank = 1,
             int? count = null)
         {
+            if (OrderedArenaInfos.Count == 0)
+            {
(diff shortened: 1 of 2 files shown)


Lib9c/Model/State/WeeklyArenaState.cs
```diff
@@ -100,6 +100,11 @@ private void ResetOrderedArenaInfos()
             int firstRank = 1,
             int? count = null)
         {
+            if (OrderedArenaInfos.Count == 0)
+            {
+                return new List<(int rank, ArenaInfo arenaInfo)>();
+            }
+
```
longfin, line 107: I guess it would be better if we have a regression test case for it.

longfin (approved):

Lib9c/Model/State/WeeklyArenaState.cs
boscohyun: Added it!

longfin (approved):

PR #112, opened 2020-09-29
Bug fix/make list deterministic

boscohyun:
- Add little unit test for `Bencodex.Types.List`.
- Use `Sheet.OrderedList` when use any `Sheet` in `foreach`.
- Sort materiali ids In `ItemEnhancement.Execute()`.
- Sort some lists in `RankingBattle` action.
- Sort `AvatarState.combinationSlotAddresses` once in `AvatarState` constructor.

### DIFF
--- a/Lib9c/TableData/Sheet.cs
+++ b/Lib9c/TableData/Sheet.cs
@@ -3,11 +3,15 @@ using System.Collections;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+#if UNITY_EDITOR
 using System.Text;
+#endif
 using Bencodex.Types;
 using JetBrains.Annotations;
 using Nekoyume.Model.State;
+#if UNITY_EDITOR
 using Serilog;
+#endif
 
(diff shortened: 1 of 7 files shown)


longfin (approved):

Unengine (approved):

Lib9c/TableData/Sheet.cs
```diff
@@ -3,11 +3,15 @@
...
 using System.Linq;
+#if UNITY_EDITOR
 using System.Text;
+#endif
 using Bencodex.Types;
 using JetBrains.Annotations;
 using Nekoyume.Model.State;
+#if UNITY_EDITOR
 using Serilog;
+#endif
```
ipdae, line 14: is it correct changes?

boscohyun: Yes, this is the right change.
These two using statements are used below with the same preprocessor.

ipdae (approved):

PR #114, opened 2020-09-29
Sort quest args for deterministic

ipdae:
- Sort quest list, quest reward for deterministic

### DIFF
--- a/Lib9c/Model/Quest/QuestReward.cs
+++ b/Lib9c/Model/Quest/QuestReward.cs
@@ -14,7 +14,10 @@ namespace Nekoyume.Model.Quest
 
         public QuestReward(Dictionary<int, int> map)
         {
-            ItemMap = map;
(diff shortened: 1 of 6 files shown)


limebell (approved):

boscohyun (approved):

Lib9c/Model/Quest/QuestReward.cs
```diff
@@ -14,7 +14,10 @@ public class QuestReward : IState
 
         public QuestReward(Dictionary<int, int> map)
         {
-            ItemMap = map;
+            ItemMap = map
+                .OrderBy(kv => kv.Key)
```
longfin, line 18: I guess this `.OrderBy()` doesn't affect result because we convert it through `.ToDictionary()` directly.

longfin (approved):

PR #118, opened 2020-10-06
Add ValidatingActionRenderer

limebell:
`ValidatingActionRenderer<T>` extends `IActionRenderer<T>` where throws `InvalidRenderException` when render events occurred in unexpected order.

Additionally, I moved renders' namespace from `Nekoyume.Action` to `Lib9c.Renderer`. (which is breaking change)


earlbread (approved):

Lib9c/Renderer/ValidatingActionRenderer.cs
```diff
@@ -0,0 +1,364 @@
...
+        {
+            var state = RenderState.Ready;
+            RenderRecord<T>.Reorg reorgState = null;
+            RenderRecord<T>.Block blockState = null;
+            long previousActionBlockIndex = -1L;
+            var records = new List<RenderRecord<T>>(Records.Count);
+
+            Exception BadRenderExc(string message) => new InvalidRenderException(records, message);
+
+#pragma warning disable S2589
```
riemannulus, line 106: Maybe we need comment about `S2589`

dahlia (approved):

longfin (approved):

PR #124, opened 2020-10-08
Validate when reorg ends

limebell:
Validate function must be passed to `ValidatingActionRenderer`.

### DIFF
--- a/Lib9c/Renderer/ValidatingActionRenderer.cs
+++ b/Lib9c/Renderer/ValidatingActionRenderer.cs
@@ -5,7 +5,6 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using Libplanet.Action;
-using Libplanet.Blockchain.Policies;
 using Libplanet.Blockchain.Renderers;
 using Libplanet.Blocks;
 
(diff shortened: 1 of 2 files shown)


limebell:
Validate function became not to be passed to its constructor, it's now built in. Instead, `BlockChain` property should be set before any reorg rendering.

earlbread (approved):

Lib9c/Renderer/ValidatingActionRenderer.cs
```diff
@@ -92,51 +89,105 @@ public override void RenderBlockEnd(Block<T> oldTip, Block<T> newTip)
...
-                expectedActions.AddRange(tx.Actions.Cast<IAction>());
+                repeat = !block.PreviousHash.Equals(branchpoint.Hash);
+                expectedUnrenderedActions.AddRange(
+                    block.Transactions.SelectMany(t => t.Actions).Cast<IAction>());
+                if (!(chain.Policy.BlockAction is null))
+                {
+                    expectedUnrenderedActions.Add(chain.Policy.BlockAction);
+                }
+
+                block = block.PreviousHash is null ? chain.Genesis : store.GetBlock<T>(block.PreviousHash.Value);
```
longfin, line 130: I guess `block` is already `chain.Genesis` if `block.PreviousHash is null`. 😕

Lib9c/Renderer/ValidatingActionRenderer.cs
```diff
@@ -92,51 +89,105 @@ public override void RenderBlockEnd(Block<T> oldTip, Block<T> newTip)
...
-                expectedActions.Add(_policy.BlockAction);
+                repeat = !block.PreviousHash.Equals(branchpoint.Hash);
+                var actions = block.Transactions.SelectMany(t => t.Actions).Cast<IAction>().ToList();
+                if (!(chain.Policy.BlockAction is null))
+                {
+                    actions.Add(chain.Policy.BlockAction);
+                }
+
+                expectedRenderedActions = actions.Concat(expectedRenderedActions).ToList();
+                block = block.PreviousHash is null ? chain.Genesis : store.GetBlock<T>(block.PreviousHash.Value);
```
longfin, line 145: I guess `block` is already `chain.Genesis` if `block.PreviousHash is null`. 😕

Lib9c/Renderer/ValidatingActionRenderer.cs
limebell: You are right. Amended in [`9e44d67`](https://github.com/planetarium/lib9c/pull/124/commits/9e44d67e40b808fda503a4f58e86a3bf96dae280).

longfin (approved):

PR #151, opened 2020-10-15
Filter illegal ActivateAccount on BlockPolicy

longfin:
This PR adds a filter for `BlockPolicy` to block txs containing illegal `ActivateAccount` action.

### DIFF
--- a/Lib9c/BlockPolicySource.cs
+++ b/Lib9c/BlockPolicySource.cs
@@ -98,6 +98,8 @@ namespace Nekoyume.BlockChain
             {
                 // It can be caused during `Swarm<T>.PreloadAsync()` because it doesn't fill its 
                 // state right away...
+                // FIXME It should be removed after fix that Libplanet fills its state on IBD.
+                // See also: https://github.com/planetarium/lib9c/pull/151#discussion_r506039478
                 return true;
             }
         }
--- a/Lib9c/BlockPolicySource.cs
+++ b/Lib9c/BlockPolicySource.cs
@@ -69,14 +69,15 @@ namespace Nekoyume.BlockChain
             BlockChain<NCAction> blockChain
         )
         {
-            if (transaction.Actions.Count == 1 &&
-                transaction.Actions.First().InnerAction is ActivateAccount)
-            {
-                return true;
-            }
-
             try
             {
+                if (transaction.Actions.Count == 1 &&
+                    transaction.Actions.First().InnerAction is ActivateAccount aa)
+                {
+                    return blockChain.GetState(aa.PendingAddress) is Dictionary rawPending &&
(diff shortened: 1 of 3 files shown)


earlbread (approved):

Lib9c/BlockPolicySource.cs
```diff
@@ -69,14 +69,15 @@ public IBlockPolicy<NCAction> GetPolicy(int minimumDifficulty)
...
-            if (transaction.Actions.Count == 1 &&
-                transaction.Actions.First().InnerAction is ActivateAccount)
-            {
-                return true;
-            }
-
             try
             {
+                if (transaction.Actions.Count == 1 &&
+                    transaction.Actions.First().InnerAction is ActivateAccount aa)
```
riemannulus, line 75: Why we need `transaction.Actions.Count == 1` statement?

longfin: I don't have deep thought, but it can be fair because in the normal cases, ActivateAccount should be only one and first action per account.

Lib9c/BlockPolicySource.cs
```diff
@@ -69,14 +69,15 @@ public IBlockPolicy<NCAction> GetPolicy(int minimumDifficulty)
...
+                {
+                    return blockChain.GetState(aa.PendingAddress) is Dictionary rawPending &&
```
limebell, line 77: Can this throw `IncompleteBlockStatesException`? If so, it is okay to return true on the case where it actually fails to verify but the state is still not calculated?

longfin: > Can this throw IncompleteBlockStatesException?

Yes.

> If so, it is okay to return true on the case where it actually fails to verify but the state is still not calculated?

It can be matter. `IncompleteBlockStatesException` will be thrown during IBD. but there is no method to workaround at this time. (See also: https://github.com/planetarium/lib9c/pull/89)

I guess, the way we can solve this problem is by dropping legacy block state store and features for state downloading and forcing state calculation even if IBD too.

limebell: > I guess, the way we can solve this problem is by dropping legacy block state store and features for state downloading and forcing state calculation even if IBD too.

Great. Can we add FIXME on here or make an issue referencing your suggestion?

Lib9c/BlockPolicySource.cs
riemannulus: What happens if someone make a typo when they type the first activation key?

longfin: In that case, the first activation key will be ignored and user should challenge with other key.

riemannulus (approved):

PR #152, opened 2020-10-15
fix: replace zio abstraction with another.

moreal:
There was an issue to resolve the path of the JSON file because of the Zio's file system abstraction policy. So I replaced it with `System.IO.Abstractions`. And I also fixed the problem that can't load states from JSON successfully.

### DIFF
--- a/.Lib9c.Tests/JsonStatesLoaderTest.cs
+++ b/.Lib9c.Tests/JsonStatesLoaderTest.cs
@@ -5,9 +5,11 @@ namespace Lib9c.Tests
     using System.IO;
     using System.IO.Abstractions;
     using System.IO.Abstractions.TestingHelpers;
+    using System.Text;
     using Bencodex;
     using Bencodex.Types;
     using Libplanet;
+    using Org.BouncyCastle.Utilities.Encoders;
     using Xunit;
 
     public class JsonStatesLoaderTest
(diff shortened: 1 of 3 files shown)


limebell:
`Lib9c.Tests.JsonStatesLoaderTest.Load` fails

moreal:
> 
> 
> `Lib9c.Tests.JsonStatesLoaderTest.Load` fails

I fixed it! @limebell

.Lib9c.Tests/JsonStatesLoaderTest.cs
```diff
@@ -1,23 +1,40 @@
...
         public void Load()
         {
-            IFileSystem fileSystem = new MemoryFileSystem();
             var codec = new Codec();
             Text barValue = (Text)"bar";
-            fileSystem.WriteAllText("/foo", $"{{\"foo\": \"{ByteUtil.Hex(codec.Encode(barValue))}\"}}");
+
+            byte[] Base64Encode(byte[] data)
+            {
+                var encoder = new Base64Encoder();
```
longfin, line 25: How about using [`Convert.ToBase64String()`](https://docs.microsoft.com/ko-kr/dotnet/api/system.convert.tobase64string?view=netcore-3.1#System_Convert_ToBase64String_System_Byte___) instead of BouncyCastle's?

longfin (approved):

.Lib9c.Tests/JsonStatesLoaderTest.cs
moreal: I amended it in 3e75a3b .

.Lib9c.Tests/JsonStatesLoaderTest.cs
```diff
@@ -1,23 +1,28 @@
...
     using System;
+    using System.Collections.Generic;
     using System.IO;
+    using System.IO.Abstractions;
+    using System.IO.Abstractions.TestingHelpers;
+    using System.Text;
     using Bencodex;
     using Bencodex.Types;
     using Libplanet;
+    using Org.BouncyCastle.Utilities.Encoders;
```
longfin, line 12:
```suggestion
```

PR #158, opened 2020-10-16
Reject blocks with too many transactions in BlockPolicy

limebell:
It checks number of transactions in blocks to append, reject if its count exceeds the maximum.

### DIFF
--- a/Lib9c/BlockPolicy.cs
+++ b/Lib9c/BlockPolicy.cs
@@ -4,6 +4,7 @@ using Libplanet.Blockchain.Policies;
 using Libplanet.Blocks;
 using Libplanet.Tx;
 using System;
+using System.Linq;
 using Lib9c;
 using Libplanet;
 using Nekoyume.Model.State;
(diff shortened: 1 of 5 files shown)


longfin (approved):

earlbread (approved):

Lib9c/BlockPolicy.cs
```diff
@@ -31,13 +34,21 @@ public class BlockPolicy : BlockPolicy<NCAction>
         {
             _minimumDifficulty = minimumDifficulty;
             _difficultyBoundDivisor = difficultyBoundDivisor;
+            _maximumTransactions = Math.Max(0, maximumTransactions);
```
riemannulus, line 37: Why don't we just use [`uint`](https://docs.microsoft.com/en-us/dotnet/api/system.uint32?view=netcore-3.1) type?

limebell: 😲

longfin: In .NET, I prefer `int` to `uint` in the interface of library because it isn't CLS-compliant. ([see also](https://stackoverflow.com/questions/1828575/why-should-i-write-cls-compliant-code)).

limebell: 😲

moreal (approved):

dahlia (approved):

PR #161, opened 2020-10-19
Restrict size of transaction batch at mining

limebell:
Title says it all.

### DIFF
--- a/Lib9c/Miner.cs
+++ b/Lib9c/Miner.cs
@@ -21,7 +21,9 @@ namespace Nekoyume.BlockChain
 
         public Address Address { get; }
 
-        public async Task<Block<PolymorphicAction<ActionBase>>> MineBlockAsync(CancellationToken cancellationToken)
+        public async Task<Block<PolymorphicAction<ActionBase>>> MineBlockAsync(
+            int txBatchSize,
+            CancellationToken cancellationToken)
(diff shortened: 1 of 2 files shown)


Lib9c/Miner.cs
```diff
@@ -21,7 +21,9 @@ public class Miner
 
         public Address Address { get; }
 
-        public async Task<Block<PolymorphicAction<ActionBase>>> MineBlockAsync(CancellationToken cancellationToken)
+        public async Task<Block<PolymorphicAction<ActionBase>>> MineBlockAsync(
+            CancellationToken cancellationToken,
+            int txBatchSize)
```
longfin, line 26: How about switching positions? (many of async methods receive `CancellationToken` as the last parameter)

```suggestion
            int txBatchSize,
            CancellationToken cancellationToken)
```

earlbread (approved):

longfin (approved):

Lib9c/Miner.cs
limebell: Amended! [`da3438c`](https://github.com/planetarium/lib9c/pull/161/commits/da3438c77a7375cc91399007b8bb58eca48fbcc0)

longfin (approved):

PR #168, opened 2020-10-20
Add private-key option for 9c-tools genesis

longfin:
This PR adds `--private-key` option to `9c-tools genesis` to provide signer key for genesis block.

### DIFF
--- a/Lib9c/BlockHelper.cs
+++ b/Lib9c/BlockHelper.cs
@@ -1,7 +1,5 @@
 using System.Collections.Generic;
 using System.Collections.Immutable;
-using System.Linq;
-using Bencodex.Types;
 using Libplanet;
 using Libplanet.Action;
 using Libplanet.Assets;
@@ -27,7 +25,8 @@ namespace Nekoyume
             IImmutableSet<Address> activatedAccounts = null,
             bool isActivateAdminAddress = false,
             IEnumerable<string> credits = null,
-            int maximumTransactions = 100
+            int maximumTransactions = 100,
+            PrivateKey privateKey = null
         )
         {
             if (!tableSheets.TryGetValue(nameof(GameConfigSheet), out var csv))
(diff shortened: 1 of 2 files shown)


Lib9c/BlockHelper.cs
```diff
@@ -38,9 +37,12 @@ public static class BlockHelper
             var redeemCodeListSheet = new RedeemCodeListSheet();
             redeemCodeListSheet.Set(tableSheets[nameof(RedeemCodeListSheet)]);
 
-            // FIXME Must use a separate key for the mainnet.
-            var minterKey = new PrivateKey();
-            var ncg = new Currency("NCG", 2, minterKey.ToAddress());
+            if (privateKey is null)
+            {
+                privateKey = new PrivateKey();
```
dahlia, line 42: When the `--private-key` option is omitted, how about showing the password prompt through tty rather than creating a random one?

longfin: 🆗 I'll try

longfin: Instead of password prompt, I turned `--private-key` option to required. (IMHO it also help prevent mistake)

earlbread (approved):

moreal (approved):

Lib9c/BlockHelper.cs
```diff
@@ -27,7 +25,8 @@ public static class BlockHelper
             IImmutableSet<Address> activatedAccounts = null,
             bool isActivateAdminAddress = false,
             IEnumerable<string> credits = null,
-            int maximumTransactions = 100
+            int maximumTransactions = 100,
+            PrivateKey privateKey = null
```
dahlia, line 29: Shouldn't this parameter be required and get rid of default value?

longfin: `privateKey` for `BlockHelper.MineGenesisBlock()` is still optional. I turned only `--private-key` option for `9c-tools genesis` because `BlockHelper.MineGenesisBlock()` can be used in other situations (e.g. unittest)...

riemannulus (approved):

dahlia (approved):

PR #192, opened 2020-11-05
Fix tx patchtable sub-command

longfin:
- Use stderr instead of stdout for printing debug messages
- Add line ending to hex output

P.S. this code is cheap too.


limebell (approved):

ipdae (approved):
this approve is expensive

PR #198, opened 2020-11-10
Reduce GetState, SetState call count in IAction.Execute

ipdae:
- Delete SetState with no changes.
- Use `TryGetAvatarState` instead of `TryGetAgentAvatarStates`.

### DIFF
--- a/Lib9c/Action/AccountStateDeltaExtensions.cs
+++ b/Lib9c/Action/AccountStateDeltaExtensions.cs
@@ -155,8 +155,32 @@ namespace Nekoyume.Action
             out AvatarState avatarState
         )
         {
-            avatarState = states.GetAvatarState(avatarAddress);
-            return !(avatarState is null) && avatarState.agentAddress == agentAddress;
+            avatarState = null;
+            var value = states.GetState(avatarAddress);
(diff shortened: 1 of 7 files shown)


Lib9c/Action/AccountStateDeltaExtensions.cs
```diff
@@ -148,6 +148,17 @@ public static AvatarState GetAvatarState(this IAccountStateDelta states, Address
...
 
+        public static bool TryGetAvatarState(
+            this IAccountStateDelta states,
+            Address agentAddress,
+            Address avatarAddress,
+            out AvatarState avatarState
+        )
+        {
+            avatarState = states.GetAvatarState(avatarAddress);
+            return !(avatarState is null) && avatarState.agentAddress == agentAddress;
```
longfin, line 159: Can we do this address checking(`avatarState.agentAddress == agentAddress`) *before* full deserization by `GetAvatarState()`? it will be better if so.

ipdae: ok, i'll try

longfin (approved):

PR #199, opened 2020-11-10
Feature/make costume item tradable

boscohyun:
- Add `Costume` to `ShopItem`.
- Add `Sell2` action.
- Fix `SellTest` and `Inventory`


.Lib9c.Tests/Model/Item/ShopItemTest.cs
```diff
@@ -0,0 +1,55 @@
...
+            var beforeDeserialized = new ShopItem((Dictionary)beforeSerialized);
+
+            var serialized = shopItem.Serialize();
+            var deserialized = new ShopItem((Dictionary)serialized);
+
+            Assert.Equal(beforeSerialized, serialized);
+            Assert.Equal(beforeDeserialized, deserialized);
+        }
+    }
+}
```
ipdae, line 52: missing eol

Lib9c/Action/AccountStateDeltaExtensions.cs
```diff
@@ -199,13 +199,8 @@ public static AvatarState GetAvatarState(this IAccountStateDelta states, Address
...
             if (!agentState.avatarAddresses.ContainsValue(avatarAddress))
             {
-                Log.Error(
-                    "The avatar {0} does not belong to the agent {1}.",
-                    avatarAddress.ToHex(),
-                    agentAddress.ToHex()
-                );
-
-                return false;
+                throw new AgentStateNotContainsAvatarAddressException(
```
ipdae, line 202: i think this change is incorrect. (throw exception in `TryGetAvatarStates`)

ipdae (changes requested):
test is broken

.Lib9c.Tests/Action/SellTest.cs
```diff
@@ -90,35 +95,66 @@ public void Execute()
...
+            var productCount = 0;
+            foreach (var nonFungibleItem in items)
             {
-                BlockIndex = 0,
-                PreviousStates = _initialState,
-                Rehearsal = false,
-                Signer = _agentAddress,
-                Random = new ItemEnhancementTest.TestRandom(),
-            });
+                var sellAction = new Sell2
```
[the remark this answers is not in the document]
hskim881028: I agree to create new test class. And <pre><code>{SellTest.cs}</code></pre> is going to revert

Lib9c/Action/AccountStateDeltaExtensions.cs
hskim881028: I guess.. 
this line seems to have thrown an exception when there is no AvatarAddress in AgentState.

Lib9c/Model/Item/ShopItem.cs
```diff
@@ -62,12 +103,12 @@ public override bool Equals(object obj)
...
+            return Equals((ShopItem) obj);
         }
 
         public override int GetHashCode()
         {
             return ProductId.GetHashCode();
         }
     }
-}
+}
```
riemannulus, line 114:
```suggestion
}

```

Lib9c/Action/AccountStateDeltaExtensions.cs
boscohyun: Oh right. It would be correct not to throw an exception inside `TryXxx()`.

Lib9c/Action/AccountStateDeltaExtensions.cs
```diff
@@ -199,13 +199,8 @@ public static AvatarState GetAvatarState(this IAccountStateDelta states, Address
...
+                    $"The avatar {avatarAddress.ToHex()} does not belong to the agent {agentAddress.ToHex()}.");
```
boscohyun, line 203:
```suggestion
                Log.Error(
                    "The avatar {0} does not belong to the agent {1}.",
                    avatarAddress.ToHex(),	
                    agentAddress.ToHex()	
                );	

                return false;
```

Lib9c/Action/AttachmentActionResult.cs
```diff
@@ -36,14 +37,37 @@ protected AttachmentActionResult()
...
+
+            if (costume != null)
+            {
+                innerDictionary.Add((Text) "costume", costume.Serialize());
+            }
+            
+            return new Dictionary(innerDictionary);
+        }
+        
+        public virtual IValue SerializeBackup1() =>
```
ipdae, line 70: is this any test case for backward compatibility?

PR #212, opened 2020-11-18
Fix runtime deserialization of `Chest`

boscohyun:
- The title says it all.
- In fact, we don't use the `Chest` object. However, it is cumbersome to remove it, so temporary measures are taken to avoid causing an error.

### DIFF
--- a/.Lib9c.Tests/Model/Item/ChestTest.cs
+++ b/.Lib9c.Tests/Model/Item/ChestTest.cs
@@ -1,0 +1,33 @@
+namespace Lib9c.Tests.Model.Item
+{
+    using System.Collections.Generic;
+    using System.Linq;
+    using Nekoyume.Model.Item;
+    using Nekoyume.TableData;
+    using Xunit;
+
+    public class ChestTest
+    {
+        private readonly MaterialItemSheet.Row _chestRow;
+
+        public ChestTest()
+        {
+            var dict = TableSheetsImporter.ImportSheets();
+            var tableSheets = new TableSheets(dict);
+            _chestRow = tableSheets.MaterialItemSheet.OrderedList.FirstOrDefault(row =>
+                row.ItemSubType == ItemSubType.Chest);
+        }
+
+        [Fact]
+        public void Serialization()
+        {
+            Assert.NotNull(_chestRow);
+
+            var chest = new Chest(_chestRow, new List<RedeemRewardSheet.RewardInfo>());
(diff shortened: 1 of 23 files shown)


ipdae (approved):

moreal (approved):

limebell (changes requested):
Can we have unit test serializing and deserializing `Chest` instance? And we need `GetObjectData(SerializationInfo, StreamingContext)` method either.

Lib9c/Model/Item/Chest.cs
```diff
@@ -32,6 +33,10 @@ public Chest(Dictionary serialized) : base(serialized)
             }
         }
 
+        public Chest(SerializationInfo info, StreamingContext context) : base(info, context)
+        {
```
limebell, line 37: Should deal with `Rewards` field here.

limebell (changes requested):

boscohyun:
> Can we have unit test serializing and deserializing `Chest` instance? And we need `void GetObjectData(SerializationInfo, StreamingContext)` method either.

Thanks. I Added `ChestTest`.
And can you explain the reason we need `void GetObjectData(SerializationInfo, StreamingContext)` method?

Unengine (approved):

.Lib9c.Tests/Model/Item/ChestTest.cs
```diff
@@ -0,0 +1,33 @@
...
+        [Fact]
+        public void Serialization()
+        {
+            Assert.NotNull(_chestRow);
+
+            var chest = new Chest(_chestRow, new List<RedeemRewardSheet.RewardInfo>());
+            var serialized = chest.Serialize();
+            var deserialized = new Chest((Bencodex.Types.Dictionary)serialized);
+
+            Assert.Equal(chest, deserialized);
```
limebell, line 30: Can you please add test about .NET serialization? Because it contains `[Serializable]`.
ref) https://docs.microsoft.com/ko-kr/dotnet/standard/serialization/

limebell (changes requested):
I think `GetObjectData` override method is missing. Please see https://docs.microsoft.com/ko-kr/dotnet/api/system.runtime.serialization.iserializable.getobjectdata?view=net-5.0

boscohyun:
@limebell For all `Model.Item.Xxx` classes, I added runtime serialization features and wrote unit tests.

limebell (approved):

PR #234, opened 2020-12-03
Implement/IEquippableItem interface

boscohyun:
- Add `IEquippableItem` interface.
- Implement 'IEquippableItem` from `Equipment` and `Costume`
- Bug fix in `Inventory.HasItem(Guid)`: Since this method is only being used for unit tests, I think it is a change that does not affect the chain.

### DIFF
--- a/Lib9c/Model/Item/Costume.cs
+++ b/Lib9c/Model/Item/Costume.cs
@@ -10,6 +10,7 @@ namespace Nekoyume.Model.Item
     [Serializable]
     public class Costume : ItemBase, INonFungibleItem, IEquippableItem
     {
+        // FIXME: Do not use anymore please!
         public bool equipped = false;
         public string SpineResourcePath { get; }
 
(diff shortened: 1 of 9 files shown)


Lib9c/Model/Item/Costume.cs
```diff
@@ -8,12 +8,13 @@
 namespace Nekoyume.Model.Item
 {
     [Serializable]
-    public class Costume : ItemBase, INonFungibleItem
+    public class Costume : ItemBase, INonFungibleItem, IEquippableItem
     {
         public bool equipped = false;
```
ipdae, line 13: this field can be private

boscohyun: @ipdae We already use this field in `AvatarState`, `Sell2`, `Sell3`..😭

ipdae: 😅 Could you please add a comment this field not to use anymore?

boscohyun: And.. I find some week logic https://github.com/boscohyun/lib9c/blob/e3ea312f2f806672db7dc7fef5d112bb98e8d3c4/Lib9c/Model/State/AvatarState.cs#L522

ipdae (approved):

Unengine (approved):

PR #227, opened 2020-11-30
Features/mimisbrunnr

hskim881028:
Added actions and sheets for  mimisbrunnr

PR #236, opened 2020-12-04
Add filter in getting cleared stage id.

Unengine:
Filter hard mode stage ID in `TryGetLastStageId`.


Lib9c/GameConfig.cs
```diff
@@ -40,6 +40,8 @@ public static class GameConfig
         public const int ArenaScoreDefault = 1000;
         public const int ArenaChallengeCountMax = 5;
 
+        public const int HardStageBeginId = 101;
```
boscohyun, line 43: https://github.com/planetarium/lib9c/pull/227 PR contains [this line](https://github.com/hskim881028/lib9c/blob/4ab035a7f544804ae08c4a0729d0b3c82546d8b3/Lib9c/GameConfig.cs#L20)

Unengine: I filtered with `World.Id`, but is it better to filter with `Stage.Id`?

ipdae (approved):

hskim881028 (approved):
Previously, if you had cleared any of the Stages in Mimirsbrunnr, you only get mimisbrunnr stage id. So you divided the method.

Unengine (commented):
@boscohyun LGTM!

PR #245, opened 2020-12-15
Filter RequiredBlockIndex on equipment notification.

Unengine:


ipdae (changes requested):
test is broken

ipdae (approved):

.Lib9c.Tests/Model/ItemNotificationTest.cs
```diff
@@ -143,9 +143,44 @@ public void EquipTwoRings()
...
+                .Where(r => r.ItemSubType == ItemSubType.Ring && r.Grade == 1);
+            foreach (var row in rows)
+            {
+                var guid = Guid.NewGuid();
+                var equipment = (Equipment)ItemFactory.CreateItemUsable(row, guid, 1, 0);
+                inventory.AddItem(equipment);
+            }
+
+            hasNotification = inventory.HasNotification(avatarState.level, 0);
+            Assert.False(hasNotification);
```
boscohyun, line 183: Shouldn't it be'True' after a ring is added to your inventory?

Unengine: `var equipment = (Equipment)ItemFactory.CreateItemUsable(row, guid, 1, 0);`
As you can see, its `RequiredIndex` is 1, so it should be false.

boscohyun: Now I see.
How about making it easier to read using the `Theory` and `InlineData` attributes?
ex.
```
[Theory]
[InlineData(0, 0, true)]
[InlineData(0, 1, false)]
[InlineData(...)]
public void RequiredBlockIndexTest(long blockIndex, long requiredBlockIndex, bool hasNotification)
{ /* */ }
```

Unengine: Seems like a good idea!

boscohyun (approved):

PR #248, opened 2020-12-17
Make TransferAsset cannot have same sender and recipient

limebell:
This is applied after block 380000 (now 374631 at 2020.12.17 14:10 KST)


junorouse (approved):
🚢

longfin (approved):

PR #249, opened 2020-12-17
Add SelectV3() and  testcode

hskim881028:
- Added Select method version 3
- Modified to properly receive Ranked Arena reward items

### DIFF
--- a/.Lib9c.Tests/Model/RankingSimulatorTest.cs
+++ b/.Lib9c.Tests/Model/RankingSimulatorTest.cs
@@ -188,12 +188,10 @@ namespace Lib9c.Tests.Model
             {
                 level = level,
             };
-            avatarState.worldInformation.ClearStage(
-                1,
-                GameConfig.RequireClearedStageLevel.ActionsInRankingBoard,
-                1,
+            avatarState.worldInformation = new WorldInformation(
+                0,
                 _tableSheets.WorldSheet,
-                _tableSheets.WorldUnlockSheet);
+                GameConfig.RequireClearedStageLevel.ActionsInRankingBoard);
 
             var simulator = new RankingSimulator(
                 _random,
--- a/.Lib9c.Tests/Model/RankingSimulatorTest.cs
+++ b/.Lib9c.Tests/Model/RankingSimulatorTest.cs
@@ -167,5 +167,71 @@ namespace Lib9c.Tests.Model
             var enemyPlayer = (EnemyPlayer)e.Character;
             Assert.Equal(row2.Stat, enemyPlayer.Stats.OptionalStats.DEF);
         }
+
+        [Theory]
+        [InlineData(1, 1000)]
+        [InlineData(55, 1000)]
+        [InlineData(85, 1000)]
+        [InlineData(120, 1000)]
+        [InlineData(160, 1000)]
+        [InlineData(200, 1000)]
+        public void CheckToReceiveAllRewardItems(int level, int simulationCount)
+        {
+            _tableSheets.WeeklyArenaRewardSheet = _tableSheets.WeeklyArenaRewardSheet;
+            var avatarState = new AvatarState(
+                default,
+                default,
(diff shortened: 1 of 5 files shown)


Lib9c/Action/CombinationEquipment3.cs
```diff
@@ -293,7 +293,7 @@ Equipment equipment
                 new EquipmentItemOptionSheet.Row[0];
             try
             {
-                optionRows = optionSelector.SelectV2(subRecipe.MaxOptionLimit);
```
ipdae, line 296: is this result of v2 and v3 are same?

Lib9c/Battle/Simulator.cs
```diff
@@ -68,7 +68,7 @@ MaterialItemSheet materialItemSheet
             {
                 try
                 {
-                    var data = itemSelector.SelectV2(1).First();
```
ipdae, line 71: is this result of v2 and v3 are same?

.Lib9c.Tests/Model/RankingSimulatorTest.cs
```diff
@@ -167,5 +167,71 @@ public void ConstructorWithCostume()
...
         }
+
+        [Theory]
+        [InlineData(1, 1000)]
+        [InlineData(55, 1000)]
+        [InlineData(85, 1000)]
+        [InlineData(120, 1000)]
+        [InlineData(160, 1000)]
+        [InlineData(200, 1000)]
+        public void CheckToReceiveAllRewardItems(int level, int simulationCount)
```
ipdae, line 178: In my opinion, it would be better to check the order of randomly selected values. (like [WeightedSelectorTest.SelectV2Multiple](https://github.com/planetarium/lib9c/blob/1e3912bd78b634a1e2760c30f291a325a08a8c44/.Lib9c.Tests/Model/WeightedSelectorTest.cs#L65))

Lib9c/Action/CombinationEquipment3.cs
hskim881028: If the sum of the `option_ratio` parameters in the `EquipmentItemSubRecipeSheet` was not 1, the result of v2 and v3 would have been different. However, the sum of the option ratio parameters in the `EquipmentItemSubRecipeSheet` is 1.

Lib9c/Battle/Simulator.cs
hskim881028: If the sum of the `item_ratio` parameters in the `StageSheet` was not 1, the result of v2 and v3 would have been different. However, the sum of the option ratio parameters in the `StageSheet` is 1.

.Lib9c.Tests/Model/RankingSimulatorTest.cs
hskim881028: That's a good idea. I Will add test case.

.Lib9c.Tests/Model/RankingSimulatorTest.cs
```diff
@@ -167,5 +167,71 @@ public void ConstructorWithCostume()
...
+                default)
+            {
+                level = level,
+            };
+            avatarState.worldInformation.ClearStage(
+                1,
+                GameConfig.RequireClearedStageLevel.ActionsInRankingBoard,
+                1,
+                _tableSheets.WorldSheet,
+                _tableSheets.WorldUnlockSheet);
```
boscohyun, line 196:
```suggestion
            avatarState.worldInformation = new WorldInformation(
                0,
                _tableSheets.WorldSheet,
                GameConfig.RequireClearedStageLevel.ActionsInRankingBoard);
```

`WorldInformation.ClearStage()` works only when the previous stage of the stage received as an argument is cleared. So it's easy to write it like this.

boscohyun (approved):

Unengine (approved):

ipdae (approved):

PR #253, opened 2020-12-22
Add unit tests for `SimplePriorityQueue`

boscohyun:
- The title says it all.


ipdae (approved):
it would be better expected result const for equals check

hskim881028 (approved):

PR #264, opened 2021-01-07
Divide SetRewardV2 for Backward compatibility

ipdae:
- Divide SetReward, SetRewardV2 for backward compatibility

### DIFF
--- a/Lib9c/Battle/Simulator.cs
+++ b/Lib9c/Battle/Simulator.cs
@@ -66,6 +66,47 @@ namespace Nekoyume.Battle
             {
                 try
                 {
+                    var data = itemSelector.SelectV2(1).First();
+                    if (materialItemSheet.TryGetValue(data.ItemId, out var itemData))
+                    {
+                        var count = random.Next(data.Min, data.Max + 1);
+                        for (var i = 0; i < count; i++)
+                        {
+                            var item = ItemFactory.CreateMaterial(itemData);
+                            if (reward.Count < maxCount)
+                            {
+                                reward.Add(item);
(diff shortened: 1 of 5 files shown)


hskim881028 (approved):

Lib9c/Battle/Simulator.cs
```diff
@@ -62,6 +62,47 @@ MaterialItemSheet materialItemSheet
         {
             var reward = new List<ItemBase>();
 
+            while (reward.Count < maxCount)
+            {
+                try
+                {
+                    var data = itemSelector.SelectV2(1).First();
```
boscohyun, line 69: Isn't it correct to use `itemSelector.SelectV2()` in `SetRewardV2()`?

ipdae: nope, `SetRewardV2` to use in `SelectV3` is correct. because `SetReward` is being used for older actions.(like `HackAndSlash, HackAndSlash2, HackAndSlash3`)

Unengine (approved):

Lib9c/Battle/Simulator.cs
boscohyun: Oh I was wondering why there is no more logic to use `itemSelector.Select()` in `SetReward()` method.
Looking at the log, it was confirmed that `itemSelector.Select()` was uploaded to `itemSelector.SelectV2()` without upgrading `SetReward()` in the f58f998 commit.

boscohyun (approved):

longfin (approved):

PR #266, opened 2021-01-08
Implement crafting custom equipment for test.

Unengine:
You can add options by adding item option rows in the constructor of `optionRows` variable.

### DIFF
--- a/Lib9c/Action/CombinationEquipment.cs
+++ b/Lib9c/Action/CombinationEquipment.cs
@@ -317,5 +317,35 @@ namespace Nekoyume.Action
 
             return optionIds;
         }
+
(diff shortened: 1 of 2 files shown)


Lib9c/Action/CombinationEquipment.cs
```diff
@@ -317,5 +317,35 @@ Equipment equipment
 
             return optionIds;
         }
+
+        public static HashSet<int> AddOption(
```
ipdae, line 321: it would be better move to `CreateAvatar`.

ipdae (approved):

hskim881028 (approved):

PR #329, opened 2021-03-08
Store migration tool

riemannulus:
SSIA


riemannulus:
https://github.com/planetarium/lib9c/runs/2057181883?check_suite_focus=true check success.

.github/scripts/run-migration.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
```
dahlia, line 2: To exit immediately with non-zero status when any error occurs:

```suggestion
set -e

```

See also the [docs about GNU Bash's `set` builtin][1].

[1]: https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html

.github/scripts/run-migration.sh
```diff
@@ -0,0 +1,12 @@
...
+echo "Build Lib9c.Tools"
```
dahlia, line 3: Informational log messages should go to the standard error (rather than the standard output):

```suggestion
echo "Build Lib9c.Tools" > /dev/stderr
```

.github/workflows/main.yml
```diff
@@ -44,6 +44,7 @@ jobs:
         pushd _benchmarks_snapshot
         7z x "$tmpfile"
         popd
+        .github/scripts/run-migration.sh
```
dahlia, line 47: As the directory names like `_benchmarks_snapshot` & `_benchmarks_snapshot_migrate` are hard-coded into the script, it seems better to be here rather than having a separated script file.  How about that?

area363 (approved):

.github/workflows/main.yml
riemannulus: https://github.com/planetarium/lib9c/pull/329/commits/3c17c3439ba85fab2aeebe35073733023cebe298 Merged separated file into main.yml

riemannulus:
PTAL! 🙏  @dahlia

dahlia (approved):

PR #331, opened 2021-03-09
Introduce BuyMultiple

Unengine:
- Implemented new action, `BuyMultiple`.
- Based on `Buy4` action.

### DIFF
--- a/.Lib9c.Tests/Action/BuyMultipleTest.cs
+++ b/.Lib9c.Tests/Action/BuyMultipleTest.cs
@@ -1,0 +1,367 @@
+﻿namespace Lib9c.Tests.Action
+{
+    using System;
+    using System.Collections;
+    using System.Collections.Generic;
+    using System.Linq;
+    using System.Numerics;
+    using Libplanet;
+    using Libplanet.Action;
+    using Libplanet.Assets;
+    using Libplanet.Crypto;
+    using Nekoyume;
+    using Nekoyume.Action;
+    using Nekoyume.Model;
+    using Nekoyume.Model.Item;
+    using Nekoyume.Model.Mail;
+    using Nekoyume.Model.State;
+    using Serilog;
+    using Xunit;
+    using Xunit.Abstractions;
+
+    public class BuyMultipleTest
+    {
+        private readonly Address _sellerAgentAddress;
+        private readonly Address _sellerAvatarAddress;
+        private readonly Address _buyerAgentAddress;
+        private readonly Address _buyerAvatarAddress;
+        private readonly AvatarState _buyerAvatarState;
+        private readonly TableSheets _tableSheets;
+        private readonly GoldCurrencyState _goldCurrencyState;
+        private IAccountStateDelta _initialState;
+
+        public BuyMultipleTest(ITestOutputHelper outputHelper)
+        {
+            Log.Logger = new LoggerConfiguration()
+                .MinimumLevel.Verbose()
+                .WriteTo.TestOutput(outputHelper)
+                .CreateLogger();
+
+            _initialState = new State();
+            var sheets = TableSheetsImporter.ImportSheets();
+            foreach (var (key, value) in sheets)
+            {
+                _initialState = _initialState
+                    .SetState(Addresses.TableSheet.Derive(key), value.Serialize());
+            }
+
+            _tableSheets = new TableSheets(sheets);
+
+            var currency = new Currency("NCG", 2, minters: null);
+            _goldCurrencyState = new GoldCurrencyState(currency);
+
+            _sellerAgentAddress = new PrivateKey().ToAddress();
+            var sellerAgentState = new AgentState(_sellerAgentAddress);
+            _sellerAvatarAddress = new PrivateKey().ToAddress();
+            var rankingMapAddress = new PrivateKey().ToAddress();
+            var sellerAvatarState = new AvatarState(
+                _sellerAvatarAddress,
+                _sellerAgentAddress,
+                0,
+                _tableSheets.GetAvatarSheets(),
+                new GameConfigState(),
+                rankingMapAddress)
+            {
+                worldInformation = new WorldInformation(
+                    0,
+                    _tableSheets.WorldSheet,
+                    GameConfig.RequireClearedStageLevel.ActionsInShop),
+            };
+            sellerAgentState.avatarAddresses[0] = _sellerAvatarAddress;
+
+            _buyerAgentAddress = new PrivateKey().ToAddress();
+            var buyerAgentState = new AgentState(_buyerAgentAddress);
+            _buyerAvatarAddress = new PrivateKey().ToAddress();
+            _buyerAvatarState = new AvatarState(
+                _buyerAvatarAddress,
+                _buyerAgentAddress,
+                0,
+                _tableSheets.GetAvatarSheets(),
+                new GameConfigState(),
+                rankingMapAddress)
+            {
+                worldInformation = new WorldInformation(
+                    0,
+                    _tableSheets.WorldSheet,
+                    GameConfig.RequireClearedStageLevel.ActionsInShop),
+            };
+            buyerAgentState.avatarAddresses[0] = _buyerAvatarAddress;
+
+            var equipment = ItemFactory.CreateItemUsable(
+                _tableSheets.EquipmentItemSheet.First,
+                Guid.NewGuid(),
+                0);
+
+            var consumable = ItemFactory.CreateItemUsable(
+                _tableSheets.ConsumableItemSheet.First,
+                Guid.NewGuid(),
+                0);
+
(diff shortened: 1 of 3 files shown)


hskim881028 (commented):
최근에 buy4 랑 buy1이 서로 스왑했다.
그래서 buy1이 최신코드니 맞는지 확인 부탁한다.

ipdae:
Is there any other reason for creating a new action instead of changing the argument of `Buy`? I think it would be better `Buy` supports bulk operation.

Lib9c/Action/BuyMultiple.cs
```diff
@@ -0,0 +1,325 @@
...
+
+namespace Nekoyume.Action
+{
+    [Serializable]
+    [ActionType("buy_multiple")]
+    public class BuyMultiple : GameAction
+    {
+        public Address buyerAvatarAddress;
+        public Address sellerAgentAddress;
+        public Address sellerAvatarAddress;
```
ipdae, line 25: i think seller can multiple addresses by product.

Lib9c/Action/BuyMultiple.cs
```diff
@@ -0,0 +1,325 @@
...
+        public IEnumerable<Guid> productIds;
+        public BuyerResult buyerResult;
+        public SellerResult sellerResult;
```
ipdae, line 28: i think sellerResult can multiple by seller.

boscohyun (changes requested):
`sellerAgentAddress` and `sellerAvatarAddress` must be bundled with `productId`.

Lib9c/Action/BuyMultiple.cs
```diff
@@ -0,0 +1,325 @@
...
+                }
+                nonFungibleItem.Update(context.BlockIndex);
+
+                var buyerResultToAdd = new Buy.BuyerResult
+                {
+                    shopItem = shopItem,
+                    itemUsable = shopItem.ItemUsable,
+                    costume = shopItem.Costume
+                };
+                var buyerMail = new BuyerMail(buyerResultToAdd, ctx.BlockIndex, ctx.Random.GenerateRandomGuid(), ctx.BlockIndex);
```
ipdae, line 261: I think we can reduce the buyerMail be 1.

ipdae (changes requested):
you must change seller address info field(sellerAgent, sellerAvatar) to collection.  because the address to be changed must be specified in rehearsal mode.

Lib9c/Action/BuyMultiple.cs
```diff
@@ -0,0 +1,353 @@
...
+    [Serializable]
+    [ActionType("buy_multiple")]
+    public class BuyMultiple : GameAction
+    {
+        public Address buyerAvatarAddress;
+        public Address sellerAgentAddress;
+        public Address sellerAvatarAddress;
+        public IEnumerable<Guid> productIds;
+        public BuyerResult buyerResult;
+        public SellerResult sellerResult;
```
boscohyun, line 28: Each product(shop item) has its own `sellerAgentAdress` and `sellerAvatarAddress`, `sellerResult`. How about writing it like below?
```
class BuyInfo
{
  Address agentAddress;
  Address avatarAddress;
  Guid productId;
}

List<BuyInfo> buyInfos;
List<SellerResult> sellerResults;

Execute()
{
  // buy process for `buyInfos`
  foreach (BuyInfo buyInfo in buyInfos)
  {
    // buy

    // add seller result to `sellerResults`
  }
}
```

Lib9c/Action/ActionBase.cs
```diff
@@ -295,18 +295,6 @@ protected string GetSignerAndOtherAddressesHex(IActionContext ctx, params Addres
             return sb.ToString();
         }
 
-        protected IAccountStateDelta LogError(IActionContext context, string message, params object[] values)
```
ipdae, line 298: why delete this method?

Lib9c/Action/BuyMultiple.cs
```diff
@@ -0,0 +1,401 @@
...
+            public int errorCode = 0;
+
+            public PurchaseResult()
+            {
+
+            }
+
+            public PurchaseResult(Bencodex.Types.Dictionary serialized) : base(serialized)
+            {
+                errorCode = (int) ((Integer) serialized["errorCode"]).Value;
```
ipdae, line 77:
```suggestion
                errorCode = serialized["errorCode"]).ToInteger();
```

Lib9c/Action/BuyMultiple.cs
```diff
@@ -0,0 +1,401 @@
...
+            }
+
+            public override IValue Serialize() =>
+#pragma warning disable LAA1002
+                new Bencodex.Types.Dictionary(new Dictionary<IKey, IValue>
+                {
+                    [(Text) "errorCode"] = (Integer) errorCode,
```
ipdae, line 84:
```suggestion
                    [(Text) "errorCode"] = errorCode.Serialize(),
```

Lib9c/Action/BuyMultiple.cs
```diff
@@ -0,0 +1,401 @@
...
+            if (ctx.Rehearsal)
+            {
+                states = states
+                    .SetState(buyerAvatarAddress, MarkChanged)
+                    .SetState(ctx.Signer, MarkChanged);
+
+                foreach (var info in purchaseInfos)
+                {
+                    var sellerAgentAddress = info.sellerAgentAddress;
+                    states = states.SetState(sellerAgentAddress, MarkChanged)
```
ipdae, line 168: sellerAvatarAddress is missing.

ipdae (changes requested):

Lib9c/Action/ActionBase.cs
Unengine: This is unexpected. I will rollback it!

.Lib9c.Tests/Action/BuyMultipleTest.cs
```diff
@@ -0,0 +1,670 @@
...
+                    Buy = true,
+                    Price = 30,
+                    ContainsInInventory = false,
+                },
+            };
+        }
+
+        [Theory]
+        [MemberData(nameof(GetExecuteMemberData))]
+        public void Execute(params ShopItemData[] productDatas)
```
ipdae, line 124: If you use `Theory`, it seems that you will run the test three times while going through the `ShopItemData` one by one, this is not correct test purchasing multiple items in `BuyMultiple`.

ipdae (commented):
I think we need add to test case for `BuyMuiltiple.PlainValue` SerializeDotnetAPI

.Lib9c.Tests/Action/BuyMultipleTest.cs
Unengine: The test runs just once with three `ShopItemData`. I think the test case is just not enough, so i will add some more.

.Lib9c.Tests/Action/BuyMultipleTest.cs
```diff
@@ -363,5 +550,20 @@ public PriceData(Currency currency)
...
+        {
+            public ItemType ItemType { get; set; }
+
+            public Guid ItemId { get; set; }
+
+            public BigInteger Price { get; set; }
+
+            public long RequiredBlockIndex { get; set; }
+
+            public bool Buy { get; set; }
```
hskim881028, line 665: What does this variable mean?

Unengine: If this variable is set to `false`, this product will be registered to the `ShopState`, but won't buy on the actual action.

Lib9c/Action/BuyMultiple.cs
```diff
@@ -0,0 +1,396 @@
...
+    {
+        public Address buyerAvatarAddress;
+        public IEnumerable<PurchaseInfo> purchaseInfos;
+        public BuyerResult buyerResult;
+        public SellerResult sellerResult;
+
+        public const int FAILED_LOADING_STATE = 1;
+        public const int ITEM_DOES_NOT_EXIST = 2;
+        public const int SHOPITEM_EXPIRED = 3;
+        public const int INSUFFICIENT_BALANCE = 4;
```
boscohyun, line 31:
```suggestion
        public const int PURCHASE_FAILED_LOADING_STATE = 1;
        public const int PURCHASE_FAILED_ITEM_DOES_NOT_EXIST = 2;
        public const int PURCHASE_FAILED_SHOPITEM_EXPIRED = 3;
        public const int PURCHASE_FAILED_INSUFFICIENT_BALANCE = 4;
```
These fields can be grouped by several ways as above.

Unengine: Seems like a good idea, and I grouped with `ERROR_CODE`.

ipdae (approved):

PR #346, opened 2021-03-24
Feature/introduce sharded shop state

ipdae:
- resolve #342 

before
```
2021-03-19 12:01:06 [Information] Get ShopState. Count: 16684 Time: 00:00:02.2284485
2021-03-19 12:01:06 [Information] Get ShopMatState. Type: Weapon Count: 4288 Time: 00:00:00.4246772
2021-03-19 12:01:07 [Information] Get ShopMatState. Type: Armor Count: 3368 Time: 00:00:00.3510505
2021-03-19 12:01:07 [Information] Get ShopMatState. Type: Belt Count: 2788 Time: 00:00:00.3126463
2021-03-19 12:01:07 [Information] Get ShopMatState. Type: Necklace Count: 2645 Time: 00:00:00.2485241
2021-03-19 12:01:07 [Information] Get ShopMatState. Type: Ring Count: 2373 Time: 00:00:00.3570137
2021-03-19 12:01:08 [Information] Get ShopMatState. Type: Food Count: 1201 Time: 00:00:00.1886100
2021-03-19 12:01:08 [Information] Get ShopMatState. Type: FullCostume Count: 17 Time: 00:00:00.0010733
2021-03-19 12:01:08 [Information] Get ShopMatState. Type: HairCostume Count: 0 Time: 00:00:00.0002881
2021-03-19 12:01:08 [Information] Get ShopMatState. Type: EarCostume Count: 0 Time: 00:00:00.0002679
2021-03-19 12:01:08 [Information] Get ShopMatState. Type: EyeCostume Count: 0 Time: 00:00:00.0003441
2021-03-19 12:01:08 [Information] Get ShopMatState. Type: TailCostume Count: 0 Time: 00:00:00.0001856
2021-03-19 12:01:08 [Information] Get ShopMatState. Type: Title Count: 4 Time: 00:00:00.0005781
2021-03-19 12:01:08 [Information] Get ShopMatState Total Time: 00:00:01.8872124
```
after
```
2021-03-23 17:01:58 [Information] Start "ShardShopStateByCategoryWithGuId" on BlockIndex: #1039424
2021-03-23 17:02:00 [Information] Initialize ShardedShopStates
2021-03-23 17:02:00 [Information] Initialize Finish
2021-03-23 17:02:00 [Information] Start Shard ShopState.Products
2021-03-23 17:02:01 [Information] Shard ShopState.Products Finish
2021-03-23 17:02:01 [Information] Start Set ShardedShopStates
[...]

PR #354, opened 2021-03-31
merge buy & buy multiple action

ipdae:
based on #346


Lib9c/Action/Buy.cs
```diff
@@ -148,159 +269,174 @@ public override IAccountStateDelta Execute(IActionContext context)
...
+                Address sellerAvatarAddress = purchaseInfo.sellerAvatarAddress;
+
+                purchaseResults.Add(purchaseResult);
+                if (!states.TryGetState(shardedShopAddress, out Bencodex.Types.Dictionary shopStateDict))
+                {
+                    purchaseResult.errorCode = ERROR_CODE_FAILED_LOADING_STATE;
+                    continue;
+                }
+                sw.Stop();
+                Log.Verbose("{AddressesHex}Buy Get ShopState: {Elapsed}", addressesHex, sw.Elapsed);
```
Unengine, line 293: This seems to be unmeasured on looping.

Lib9c/Action/Buy.cs
```diff
@@ -148,159 +269,174 @@ public override IAccountStateDelta Execute(IActionContext context)
...
+                sw.Restart();
+
+                Log.Verbose(
+                    "{AddressesHex}Execute Buy; buyer: {Buyer} seller: {Seller}",
+                    addressesHex,
+                    buyerAvatarAddress,
+                    sellerAvatarAddress);
+
+                // 상점에서 구매할 아이템을 찾는다.
```
Unengine, line 302: Is it still OK to keep comments in Korean?

area363 (approved):

Lib9c/Action/Buy.cs
ipdae: I think it is correct to loop because the address varies depending on the item type and itemId.

hskim881028 (approved):

PR #356, opened 2021-04-02
Delete new sell error case in buy multiple

ipdae:
delete use case from new sell(sell4).

### DIFF
--- a/Lib9c/Action/BuyMultiple.cs
+++ b/Lib9c/Action/BuyMultiple.cs
@@ -318,18 +318,6 @@ namespace Nekoyume.Action
                 productDict = (Dictionary)productDict.Remove(productIdSerialized);
                 shopStateDict = shopStateDict.SetItem("products", productDict);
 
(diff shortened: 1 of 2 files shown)


Lib9c/Action/BuyMultiple.cs
```diff
@@ -60,28 +96,51 @@ public PurchaseInfo(Bencodex.Types.Dictionary serialized)
...
 #pragma warning restore LAA1002
+            public int CompareTo(PurchaseInfo other)
+            {
+                int cmp = productId.CompareTo(other.productId);
+                if (cmp != 0)
+                {
+                    return cmp;
+                }
+
+                return 0;
```
moreal, line 107: What's difference between this way and returning `cmp` in direct?

```
                return cmp;
```

moreal (approved):

area363 (approved):

area363 (approved):

PR #415, opened 2021-05-14
Add tradableFungibleItem info in seller result

hskim881028:
SSIA


boscohyun (approved):

ipdae (approved):

boscohyun:
We need to fix the same things in the `SellCancellation` action.

Issue #416, opened 2021-05-17
Invalid update tradable item when sell cancel


ipdae:
updating wrong items required block index.

PR #417, opened 2021-05-17
fix inventory tradable item methods

ipdae:
- #416
- Select same required block index & item count from inventory when sell cancel.
- Check required block index when remove tradable item.

### DIFF
--- a/Lib9c/Action/SellCancellation.cs
+++ b/Lib9c/Action/SellCancellation.cs
@@ -186,14 +186,10 @@ namespace Nekoyume.Action
                 throw new InvalidShopItemException($"{addressesHex}Tradable Item is null.");
             }
 
-            if (avatarState.inventory.TryGetTradableItem(
-                    tradableItem.TradableId,
-                    tradableItem.RequiredBlockIndex,
-                    itemCount,
-                    out var tradableItems))
+            if (!backwardCompatible)
             {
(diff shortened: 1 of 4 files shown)


Lib9c/Action/SellCancellation.cs
```diff
@@ -186,13 +186,13 @@ public override IAccountStateDelta Execute(IActionContext context)
...
 
-            if (avatarState.inventory.TryGetTradableItems(
+            if (avatarState.inventory.TryGetTradableItem(
                     tradableItem.TradableId,
                     tradableItem.RequiredBlockIndex,
                     itemCount,
                     out var tradableItems))
             {
-                ITradableItem tradableItemInInventory = (ITradableItem) tradableItems.First().item;
+                ITradableItem tradableItemInInventory = (ITradableItem) tradableItems.item;
```
boscohyun, line 195:
```suggestion
                ITradableItem tradableItemInInventory = (ITradableItem) tradableItem.item;
```

boscohyun (commented):
When modifying the `RequiredBlockIndex` of `(ITradableItem) Inventory.Item.Item` obtained by `Inventory.TryGetTradableItems()` in `SellCancellation`, you need to cut and modify only `count`, not all items, and keep the rest.

boscohyun (approved):

PR #425, opened 2021-05-24
Check RequiredBlockIndex in ChargeActionPoint, RapidCombination

ipdae:
SSIA


hskim881028:
We need to change the target from development to v100045.

hskim881028 (approved):

PR #454, opened 2021-06-10
Patch/authorized miner rounding

greymistcube:
This PR attempts to do the following:

- Speed up the production of authorized blocks.
- Mitigate authorized miners competing each other.

The former is hopefully achieved by making authorized miners to only mine proof transactions. This should let the authorized miners bypass most of the computationally heavy evaluating phase and let them start working on "actual mining". It is hard to tell what the full ramifications might be, but I think it is worth giving it a go.

One immediate caveat is that, for every `50` blocks, no transactions get processed. This guarantees no player transactions are processed and the only action that gets processed is the policy block action. This means players may have to wait additional block confirmation for their transactions to be processed. However, there are several existing trends to consider:

- Unconfirmed transaction propagation is independent of block mining.
- There are plenty blocks without any transaction.
- Average mining time for authorized miners already far exceeds twice the target interval.

We don't really know how much of this is due to slow block propagation and how much is due to having to mine a "heavy" block. I believe the change is safe enough to test it on the network, where the worst case scenario would result in different congestion around authorized blocks, and perhaps more reorgs. Whether this change results in good or bad, either way, we may gain a valuable insight.

The implementation of the latter is rather *sketchy*, so should be applied with caution. 😶 
Naïve rounding is implemented by forcing an authorized miner to wait and announce only at its allotted time slot. By forcing authorized minors to announce at different times, hopefully the first announced block gets propagated fast enough so that the miners on the network can get ahead far enough that there will be no real competition between authorized minors. The caveats are the following:

- Depending on the circumstances, delay can be rather large. It is maximally set to target interval * number of authorized miners.
- Delay amount chosen is rather arbitrary. Optimally, it would be better if we know how much time difference between two blocks with the same index is large enough to mostly eliminate a competition between the two.

### DIFF
--- a/Lib9c/Miner.cs
+++ b/Lib9c/Miner.cs
@@ -47,51 +47,46 @@ namespace Nekoyume.BlockChain
             var txs = new HashSet<Transaction<PolymorphicAction<ActionBase>>>();
             var invalidTxs = txs;
 
-            Transaction<PolymorphicAction<ActionBase>> authProof = null;
             Block<PolymorphicAction<ActionBase>> block = null;
             try
             {
-                if (AuthorizedMiner)
+                if (!AuthorizedMiner)
                 {
-                    authProof = StageProofTransaction();
-                }
-                block = await _chain.MineBlock(
-                    Address,
-                    DateTimeOffset.UtcNow,
-                    cancellationToken: cancellationToken,
-                    maxTransactions: maxTransactions,
-                    append: false);
-
-                if (authProof is Transaction<PolymorphicAction<ActionBase>> proof &&
-                    !block.Transactions.Contains(proof))
-                {
-                    // For any reason, if the proof tx is not contained mine a new block again
-                    // without any transactions except for the proof tx.
-                    block = Block<PolymorphicAction<ActionBase>>.Mine(
-                        block.Index,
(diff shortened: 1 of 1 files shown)


Lib9c/Miner.cs
```diff
@@ -47,37 +47,47 @@ public Transaction<PolymorphicAction<ActionBase>> StageProofTransaction()
...
-                    append: false);
-
-                if (authProof is Transaction<PolymorphicAction<ActionBase>> proof &&
-                    !block.Transactions.Contains(proof))
+                else
                 {
-                    // For any reason, if the proof tx is not contained mine a new block again
-                    // without any transactions except for the proof tx.
+                    Transaction<PolymorphicAction<ActionBase>> authProof = StageProofTransaction();
                     block = Block<PolymorphicAction<ActionBase>>.Mine(
```
longfin, line 65: It isn't from this PR, but appending with this `block` will fail because it doesn't have a state root hash, (see also: https://github.com/planetarium/libplanet/issues/1345) 😢

greymistcube:
As per request, rounding has been removed. See eac0fd5191e0cadcfc2073608e59d3a038039173.

longfin (approved):

PR #455, opened 2021-06-13
Revise Monster Collection

longfin:
This PR adapts the versions of some actions and introduces new code to match the new concept of Monster Collection.

- Replaced `MonsterCollect0` (action_type: `monster_collect`) to `MonsterCollect` (action_type: 'monter_collect2')
  - `CancelMonsterCollect` unified into `MonsterCollect` (Setting the level to 0 cancels it.)
- Replaced `ClaimMonsterCollectionReward0` (action_type: `claim_monster_collection_reward`) to `ClaimMonsterCollectionReward` (action_type: `claim_monster_collection_reward2`)
- Replaced `MonsterCollectionState0` to `MonsterCollectionState`

There are many changes, so it is recommended to focus on the newly added class rather than the replaced class.


Lib9c/Action/ClaimMonsterCollectionReward.cs
```diff
@@ -41,54 +41,52 @@ public override IAccountStateDelta Execute(IActionContext context)
...
+            if (monsterCollectionState.ReceivedBlockIndex > 0)
             {
-                throw new RequiredBlockIndexException(
-                    $"{collectionAddress} is not available yet; it will be available after {Math.Max(monsterCollectionState.StartedBlockIndex, monsterCollectionState.ReceivedBlockIndex) + MonsterCollectionState.RewardInterval}");
+                int previousStep = (int)Math.DivRem(
+                    monsterCollectionState.ReceivedBlockIndex - monsterCollectionState.StartedBlockIndex,
+                    MonsterCollectionState.RewardInterval,
+                    out _
+                );
+                step -= previousStep;
```
ipdae: IMO, it would be better to create method for this logic. because, It looks like it will be used in the NineChronicles.Headless API too.

ipdae (approved):

Lib9c/Action/ClaimMonsterCollectionReward.cs
```diff
@@ -14,25 +14,24 @@
...
     {
         public Address avatarAddress;
-        public int collectionRound;
         public override IAccountStateDelta Execute(IActionContext context)
         {
             IAccountStateDelta states = context.PreviousStates;
-            Address collectionAddress = MonsterCollectionState.DeriveAddress(context.Signer, collectionRound);
-
+            Address collectionAddress = MonsterCollectionState.DeriveAddress(context.Signer);
+            Address inventoryAddress = avatarAddress.Derive(LegacyInventoryKey);
```
ipdae, line 25: it must be add `worldInformation address, questList address` too. because this action might be the first time called `AvatarState.SerializeV2()`.

ipdae (changes requested):

Lib9c/Action/MonsterCollect.cs
```diff
@@ -37,52 +36,45 @@ public override IAccountStateDelta Execute(IActionContext context)
...
-            int currentLevel = 1;
-            MonsterCollectionRewardSheet monsterCollectionRewardSheet = states.GetSheet<MonsterCollectionRewardSheet>();
             if (states.TryGetState(monsterCollectionAddress, out Dictionary stateDict))
             {
-                monsterCollectionState = new MonsterCollectionState(stateDict);
-
-                if (monsterCollectionState.ExpiredBlockIndex < context.BlockIndex)
+                var existingStates = new MonsterCollectionState(stateDict);
+                int previousLevel = existingStates.Level;
+                // 락업 확인
```
boscohyun, line 53:
```suggestion
                // Check collection level and required block index
```

boscohyun (commented):
How about renaming `MonsterCollectStateTest.cs` to `MonsterCollectionStateTest.cs`

Lib9c/Action/ClaimMonsterCollectionReward.cs
longfin: - QuestList can be fair enough, but WorldInformation doesn't seem to relate to this action.
- As you said, we should care about first state migration, but this code will remain and produces meaningless diffs after migration. is there any way to check the avatar state already migrated?

longfin:
@ipdae @boscohyun PTAL

I've applied for reviews and introduced a new exception(MonsterCollectionExistingClaimableException) to address the issue of changing claimable collections. I heard about @Namyujeong that in this case, we should block these changes.

ipdae (approved):
It would be nice if tests for AvatarState.Serialize version-specific behavior were added.

PR #467, opened 2021-06-22
Proof of concept: See candidate chains' generation number first

dahlia:
As long as we have authorized miners who authorize a chain every 50 blocks, comparing blocks before the last authorized block is unnecessary for determining the canonical chain.  This patch tries an experimental approach to compare chains' *generation number* first, and then compare total difficulties among longest generations.

For example, suppose we have 4 competing chains (tips):

| Chains:          | Chain A | Chain B | Chain C |
|------------------|---------|---------|---------|
| Tip index        | 10001   | 10002   | 9999    |
| Last authorized  | 10000   | 10000   | 9950    |
| Generation       | 200     | 200     | 199     |
| Total difficulty | 50010   | 50009   | 50011   |

Under the existing rule ([`Libplanet.Blockchain.TotalDifficultyComparer`](https://github.com/planetarium/libplanet/blob/0.11.1/Libplanet/Blockchain/TotalDifficultyComparer.cs)), the chain C is determined to be canon due to its total difficulty, even though other two candidates have more authorized blocks.

However, under the new rule (`Lib9c.CanonicalChainComparer`), the chain A is determined to be canon instead, because the new rule eliminates the chain C first due to its inferior generation number (200 vs 199), and then compares chain A & B's total difficulties.

### DIFF
--- a/.Lib9c.Tests/CanonicalChainComparerTest.cs
+++ b/.Lib9c.Tests/CanonicalChainComparerTest.cs
@@ -1,0 +1,71 @@
+namespace Lib9c.Tests
+{
+    using System;
+    using System.Linq;
+    using System.Numerics;
+    using Libplanet;
+    using Libplanet.Action;
+    using Libplanet.Blockchain;
+    using Libplanet.Blocks;
+    using Nekoyume.Action;
+    using Nekoyume.Model.State;
+    using Xunit;
+
+    public class CanonicalChainComparerTest
+    {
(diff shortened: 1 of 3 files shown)


longfin (approved):

dahlia:
Added minimum unit tests, and as a result I found a bug as well, which I fixed.  Also rebased and retargeted to the branch rc-v100051.

longfin (approved):

PR #469, opened 2021-06-22
Migration legacy shop state

ipdae:
- unregister shop item and return avatar inventory from legacy shop state.
- There are a total of 1507 target avatar addresses(seller), and it takes about 3 minutes to run.


Unengine (approved):

Lib9c/Action/MigrationLegacyShop.cs
```diff
@@ -0,0 +1,83 @@
...
+                    {
+                        avatarState.inventory.AddItem(shopItem.ItemUsable);
+                    }
+                    if (!(shopItem.Costume is null))
+                    {
+                        avatarState.inventory.AddItem(shopItem.Costume);
+                    }
+                    shopState.Unregister(shopItem);
+                }
+                states = states.SetState(avatarAddress, avatarState.Serialize());
```
hskim881028, line 64: The `shopState` is set at the end, is there any problem if the action fails while running?

Lib9c/Action/MigrationLegacyShop.cs
```diff
@@ -0,0 +1,83 @@
...
+
+                    if (!(shopItem.ItemUsable is null))
+                    {
+                        avatarState.inventory.AddItem(shopItem.ItemUsable);
+                    }
+                    if (!(shopItem.Costume is null))
+                    {
+                        avatarState.inventory.AddItem(shopItem.Costume);
+                    }
+                    shopState.Unregister(shopItem);
```
boscohyun, line 62: How about unregister `shopItem` from `shopState` only when succeed to migration like below.
```suggestion
                    var migrated = false;
                    if (!(shopItem.ItemUsable is null))
                    {
                        avatarState.inventory.AddItem(shopItem.ItemUsable);
                        migrated = true;
                    }
                    if (!(shopItem.Costume is null))
                    {
                        avatarState.inventory.AddItem(shopItem.Costume);
                        migrated = true;
                    }
                    
                    if (migrated)
                    {
                        shopState.Unregister(shopItem);
                    }
```

ipdae: Can `AddItem` fail?

Lib9c/Action/MigrationLegacyShop.cs
```diff
@@ -0,0 +1,83 @@
...
+            var groupBy = shopState.Products.Values
+                .Where(s => SellerAvatarAddresses.Contains(s.SellerAvatarAddress))
+                .GroupBy(p => p.SellerAvatarAddress)
+                .ToList();
+
+            Log.Debug($"Found {groupBy.Count} Seller Avatar.");
+            foreach (var group in groupBy)
+            {
+                var avatarAddress = group.Key;
+                var avatarState = states.GetAvatarState(avatarAddress);
```
longfin, line 42: how about caching `avatarState` for the same `avatarAddress`?

Lib9c/Action/MigrationLegacyShop.cs
boscohyun: I think the `shopItem` should remained in `shopState` when `shopItem.ItemUsable` and `shopItem.Costume` are `null`.

ipdae: If both is null, it can't register shop. so I think this case can be ignored.

Lib9c/Action/MigrationLegacyShop.cs
ipdae: i try caching `avatarState`, but cache does not work. because `GetState` is called only once per avatar address.

Lib9c/Action/MigrationLegacyShop.cs
boscohyun: Your words are right. I intended not to even consider the registration logic of `ShopState` within the context of this action.
I wanted the `ShopItem` registered in `ShopState` to not have to know that either `ItemUsable` or `Consumable` must not be `null` when determining whether the `MigrationLegacyShop` action is appropriate.

boscohyun (approved):

hskim881028 (approved):

PR #473, opened 2021-06-23
Add new sheets and unit tests

boscohyun:
### Add dummy csv files for new sheets

1. EnhancementCostSheetV2.csv
2. EquipmentItemOptionSheetV2.csv
3. EquipmentItemSubRecipeSheetV2.csv
4. ItemRequirementSheet.csv

### Add new sheets

1. EnhancementCostSheetV2
2. EquipmentItemOptionSheetV2
3. EquipmentItemSubRecipeSheetV2
4. ItemRequiementSheet

### Add new unit tests for sheets

1. EnhancementCostSheetTest
2. EnhancementCostSheetV2Test
3. EquipmentItemOptionSheetTest
4. EquipmentItemOptionSheetV2Test
5. EquipmentItemSubRecipeSheetTest
6. EquipmentItemSubRecipeSheetV2Test
7. ItemRequiementSheetTest


ipdae (commented):
There seems to be no difference between EquipmentItemRequirementSheet and ConsumableItemRequirementSheet, it would be better to handle it in one ItemRequirementSheet.

boscohyun:
> There seems to be no difference between EquipmentItemRequirementSheet and ConsumableItemRequirementSheet, it would be better to handle it in one ItemRequirementSheet.

I applied it.

Lib9c/TableData/Cost/EnhancementCostSheetV2.cs
```diff
@@ -0,0 +1,37 @@
...
+using System.Numerics;
+using Nekoyume.Model.Item;
+
+namespace Nekoyume.TableData
+{
+    using static TableExtensions;
+    
+    public class EnhancementCostSheetV2 : Sheet<int, EnhancementCostSheetV2.Row>
+    {
+        public class Row: EnhancementCostSheet.Row
```
moreal, line 12: It is just a comment about style.

```suggestion
        public class Row : EnhancementCostSheet.Row
```

Lib9c/TableData/Item/EquipmentItemOptionSheetV2.cs
```diff
@@ -0,0 +1,24 @@
+using System.Collections.Generic;
+using static Nekoyume.TableData.TableExtensions;
+
+namespace Nekoyume.TableData
+{
+    public class EquipmentItemOptionSheetV2 : Sheet<int, EquipmentItemOptionSheetV2.Row>
+    {
+        public class Row: EquipmentItemOptionSheet.Row
```
moreal, line 8: It is just a comment about style.

```suggestion
        public class Row : EquipmentItemOptionSheet.Row
```

.Lib9c.Tests/Data/TableCSV/Item/EquipmentItemSubRecipeSheetV2.csv
```diff
@@ -0,0 +1,373 @@
+id,required_action_point,required_gold,required_block_index,material_id,material_count,material_2_id,material_2_count,material_3_id,material_3_count,option_id,option_ratio,option_1_required_block_index,option_2_id,option_2_ratio,option_2_required_block_index,option_3_id,option_3_ratio,option_3_required_block_index,option_4_id,option_4_ratio,option_4_required_block_index,min_option_limit,max_option_limit
```
moreal, line 1: I don't know the convention of the sheet in @planetarium/9c-dev team but I wonder simple one. Why `option_id` field doesn't have the number like `2` in `option_2_id` though `option_1_required_block_index` has?

Of course, I know it doesn't affect anything because the field name is just hint and the parser accesses each fields with index. I left this comment for only style.

boscohyun: I just guess there are no `option_2_id` and `option_3_id` columns when the `option_id` column added.
Or not, It seems a little habits of writer.
And now we use csv file as state so.. If we want change that yes we can via `PatchTableSheet` action.😉

boscohyun: @moreal Can you please review this pull request?

moreal (approved):

PR #481, opened 2021-06-26
Add support for previous monster collection

longfin:
This PR extends `MonsterCollectionState.DeriveAddress()` to receive collection round for previous monster collection. it can help migration due to continuing previous state.


ipdae (approved):
how about add a field to `ClaimMonsterCollectionReward` reference from  `AgentState.CollectionRound`?

PR #483, opened 2021-06-28
Introduce `ItemUsable2`

boscohyun:
- Refactor the `ItemUsable` class
- Add the `CreateItemV2()` and `CreateItemUsableV2()` method to the `ItemFactory` class


Lib9c/Action/CreateAvatar0.cs
```diff
@@ -230,7 +230,12 @@ EquipmentItemSheet equipmentItemSheet
                 row.Id > GameConfig.DefaultAvatarWeaponId))
             {
                 var itemId = random.GenerateRandomGuid();
-                avatarState.inventory.AddItem(ItemFactory.CreateItemUsable(row, itemId, default));
+                avatarState.inventory.AddItem(ItemFactory.CreateItemUsableV2(
```
ipdae, line 233: `CreateAvatar0` is deprecated action. i think this is incorrect changes.

Lib9c/Model/Item/ItemUsable.cs
```diff
@@ -182,6 +182,28 @@ public override IValue Serialize()
                     throw new SerializeFailedException($"Serialize{_serializedVersion}() is not implemented");
             }
         }
+        
+        /// <summary>
+        /// We can migrate serializedVersion(fromVersion) objects to serializedVersion(toVersion) like below.
+        /// </summary>
+        /// <param name="toVersion"></param>
+        public IValue MigrateSerialisedVersion(int toVersion)
```
ipdae, line 192: Do not use current `ItemUsable` model. please divide new class.

ipdae (changes requested):

Lib9c/Action/ClaimMonsterCollectionReward2.cs
```diff
@@ -0,0 +1,98 @@
...
+using Nekoyume.Model.Item;
+using Nekoyume.Model.Mail;
+using Nekoyume.Model.State;
+using Nekoyume.TableData;
+using static Lib9c.SerializeKeys;
+
+namespace Nekoyume.Action
+{
+    [Serializable]
+    [ActionType("claim_monster_collection_reward2")]
```
moreal, line 16: This `type_id` seems conflicted with `Lib9c.Action.ClaimMonsterCollectionReward` class' `type_id` and it will cause error. The `Lib9c.Benchmark`'s failure reason may be this. 🤔

I don't know the relation between class name `type_id` perfectfully but it seems to need the change of `Lib9c.Action.ClaimMonsterCollectionReward`'s `type_id`.

moreal (changes requested):

Lib9c/Action/ClaimMonsterCollectionReward2.cs
boscohyun: The `class name` is not changed to make the diff easier to check. And the `class` before the change is renamed to the same name as `type_id` and archived.
e.g.,
previous
```
[ActionType("some_action")]
public class SomeAction { }
```
next
```
[ActionType("some_action2")]
public class SomeAction { }

[ActionType("some_action")]
public class SomeAction1 { }
```

next
```
[ActionType("some_action3")]
public class SomeAction { }

[ActionType("some_action2")]
public class SomeAction2 { }

[ActionType("some_action")]
public class SomeAction1 { }
```

boscohyun:
This pull request is not necessary anymore.

PR #486, opened 2021-06-30
fix sell cancel

ipdae:
- fix missing plain value
- fix error msg

### DIFF
--- a/Lib9c/Action/SellCancellation.cs
+++ b/Lib9c/Action/SellCancellation.cs
@@ -59,6 +59,7 @@ namespace Nekoyume.Action
             [ProductIdKey] = orderId.Serialize(),
             [SellerAvatarAddressKey] = sellerAvatarAddress.Serialize(),
             [ItemSubTypeKey] = itemSubType.Serialize(),
+            [TradableIdKey] = tradableId.Serialize(),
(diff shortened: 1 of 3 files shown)


boscohyun (approved):

Lib9c/Action/SellCancellation.cs
```diff
@@ -59,13 +59,18 @@ public Result(BxDictionary serialized) : base(serialized)
             [ProductIdKey] = orderId.Serialize(),
             [SellerAvatarAddressKey] = sellerAvatarAddress.Serialize(),
             [ItemSubTypeKey] = itemSubType.Serialize(),
+            [TradableIdKey] = tradableId.Serialize(),
         }.ToImmutableDictionary();
 
         protected override void LoadPlainValueInternal(IImmutableDictionary<string, IValue> plainValue)
         {
             orderId = plainValue[ProductIdKey].ToGuid();
```
hskim881028, line 67: I think we should change `ProductIdKey` to `OrderIdKey`

ipdae: I agree. However, if we change the key, the internal chain needs to be reset, so I left this.

PR #487, opened 2021-07-02
Obsolete previous actions

ipdae:
- check transaction has obsoleted actions
- prevent action execution

### DIFF
--- a/Lib9c/Action/Buy0.cs
+++ b/Lib9c/Action/Buy0.cs
@@ -17,6 +17,7 @@ using Serilog;
 namespace Nekoyume.Action
 {
     [Serializable]
+    [Obsolete("Use Buy")]
     [ActionType("buy")]
     public class Buy0 : GameAction
     {
--- a/Lib9c/Action/Buy0.cs
+++ b/Lib9c/Action/Buy0.cs
@@ -17,7 +17,7 @@ using Serilog;
 namespace Nekoyume.Action
 {
     [Serializable]
-    [Obsolete("Use Buy")]
+    [ActionObsolete(2000000)]
     [ActionType("buy")]
     public class Buy0 : GameAction
     {
--- a/Lib9c/Action/Buy0.cs
+++ b/Lib9c/Action/Buy0.cs
@@ -61,7 +61,9 @@ namespace Nekoyume.Action
                         GoldCurrencyState.Address);
                 return states.SetState(ShopState.Address, MarkChanged);
             }
-            
+
+            CheckObsolete(2000000, context);
+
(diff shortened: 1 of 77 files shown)


Lib9c/BlockPolicySource.cs
```diff
@@ -87,21 +87,31 @@ bool ignoreHardcodedIndicesForBackwardCompatibility
         public IEnumerable<IRenderer<NCAction>> GetRenderers() =>
             new IRenderer<NCAction>[] { BlockRenderer, LoggedActionRenderer };
 
+        public static bool IsObsolete(Transaction<NCAction> transaction)
+        {
+            return transaction.Actions
+                .Select(action => action.InnerAction.GetType())
+                .Any(at => at.IsDefined(typeof(ObsoleteAttribute), false));
```
longfin, line 94: It can be a matter because checking obsoleted depends on block height(index). so we need to specify its validity period on an attribute to check here.

longfin (changes requested):

Lib9c/Action/Buy0.cs
```diff
@@ -60,7 +61,9 @@ public override IAccountStateDelta Execute(IActionContext context)
                         GoldCurrencyState.Address);
                 return states.SetState(ShopState.Address, MarkChanged);
             }
-            
+
+            CheckObsolete(1815592, context);
```
longfin, line 65: It isn't matched with [`ActionObsolete()`'s](https://github.com/planetarium/lib9c/pull/487/files#diff-03b0cf6730ee3e7c83a692d427111da44c44a6c0ea29d4a71211793a14d1d77cR20). is there any reason?

ipdae: This matches the CheckPermission in the IActionContext extension. Shall we change the name?

area363 (approved):

Lib9c/Action/Buy0.cs
```diff
@@ -60,7 +61,9 @@ public override IAccountStateDelta Execute(IActionContext context)
                         GoldCurrencyState.Address);
                 return states.SetState(ShopState.Address, MarkChanged);
             }
-            
+
+            CheckObsolete(2000000, context);
```
boscohyun, line 65: Why invoke this method after rehearsal?

ipdae: because `CheckObsolete` always pass in rehearsal mode(rehearsal running blockIndex 0)

ipdae:
reconfigure blockIndex when branching rc-v100062

longfin (approved):

PR #491, opened 2021-07-05
Feature/update quest list in hack and slash

boscohyun:
- Separate the `QuestList` class from `Quest.cs` file to `QuestList.cs` file.
- Add the `QuestList.UpdateList()` method.
- Add some unit tests.
- Use `QuestList.UpdateList()` from the `HackAndSlash` action.


Lib9c/Model/Quest/QuestList.cs
```diff
@@ -0,0 +1,386 @@
...
+            _quests = serialized.TryGetValue((Text) "quests", out var questsValue)
+                ? questsValue.ToList(Quest.Deserialize)
+                : new List<Quest>();
+
+            completedQuestIds = serialized.TryGetValue((Text) "completedQuestIds", out var idsValue)
+                ? idsValue.ToList(StateExtensions.ToInteger)
+                : new List<int>();
+        }
+
+        public void UpdateList(
```
ipdae, line 115: i think it would be better add version parameter when updateList called.  because check the version information at the time of the update.

Lib9c/Model/Quest/QuestList.cs
```diff
@@ -0,0 +1,386 @@
...
+        {
+            if (questSheet.Count <= _quests.Count)
+            {
+                throw new ArgumentException(
+                    $"{nameof(questSheet)}.Count({questSheet.Count}) should greater than ${_quests.Count}");
+            }
+            
+            _listVersion += 1;
+
+            for (var i = questSheet.OrderedList.Count; i > 0; i--)
```
ipdae, line 129: it seems duplicate logic in constructor. it would be better make method.

Lib9c/Model/Quest/QuestList.cs
```diff
@@ -0,0 +1,386 @@
...
+                        _quests.Add(quest);
+                        break;
+                }
+            }
+        }
+
+
+        public QuestList(Dictionary serialized)
+        {
+            _listVersion = serialized.TryGetValue((Text) "list-version", out var listVersion)
```
ipdae, line 102: For performance reasons, it is recommended to reduce the length of the key whenever possible.

Lib9c/Model/Quest/QuestList.cs
```diff
@@ -0,0 +1,418 @@
...
+        private const string _questsKeyDeprecated = "quests";
+        private const string _questsKey = "q";
+        private readonly List<Quest> _quests;
+
+        private const string _listVersionKey = "lv";
+        private int _listVersion = 1;
+        public int ListVersion => _listVersion;
+
+        private const string _completedQuestIdsKeyDeprecated = "completedQuestIds";
+        private const string _completedQuestIdsKey = "cqi";
```
ipdae, line 69: it would be better keys should be managed in `SerializeKey.cs`

.Lib9c.Tests/Model/QuestListTest.cs
```diff
@@ -74,5 +84,89 @@ public void UpdateItemTypeCollectQuestDeterministic()
...
+                patchedSheetCsvSb.AppendLine($"{990000 + i - 1},10,{questSheet.First.QuestRewardId}");
+            }
+
+            patchedSheet.Set(patchedSheetCsvSb.ToString());
+            Assert.Equal(addedQuestCount, patchedSheet.Count);
+            var previousQuestSheetCount = questSheet.Count;
+            questSheet.Set(patchedSheet);
+            Assert.Equal(previousQuestSheetCount + addedQuestCount, questSheet.Count);
+
+            questList.UpdateList(
```
ipdae, line 119: i think check the execution result should be added.

.Lib9c.Tests/Action/HackAndSlashTest.cs
```diff
@@ -230,10 +252,107 @@ public void Execute(int avatarLevel, int worldId, int stageId, bool contains, bo
             Assert.Equal(info.AvatarAddress, _avatarAddress);
         }
 
+        [Theory]
+        [InlineData(4, 200)]
+        public void Execute_With_UpdateQuestList(int worldId, int stageId)
```
ipdae, line 257: It seems can be simplified by avatars previous quest list in Execute case.

boscohyun: > It seems can be simplified by avatars previous quest list in Execute case.

@ipdae Could you please explain more?

Lib9c/Model/Quest/QuestList.cs
```diff
@@ -361,58 +389,62 @@ public CollectionMap UpdateCompletedQuest(CollectionMap eventMap)
...
                 case CombinationEquipmentQuestSheet.Row r:
                     int stageId;
                     var recipeRow = equipmentItemRecipeSheet.Values
                         .FirstOrDefault(e => e.Id == r.RecipeId);
                     if (recipeRow is null)
                     {
-                        throw new ArgumentException($"Invalid Recipe Id : {r.RecipeId}");
+                        result.quest = null;
+                        result.errorMessage = $"Invalid Recipe Id : {r.RecipeId}";
+                        break;
```
ipdae, line 439: i think this change can be different execution result in old blocks.

ipdae (changes requested):

.Lib9c.Tests/Action/HackAndSlashTest.cs
ipdae: If we add test parms to Execute for this case(worldId, stageId). we just check avatars quest list is added in next state.

Lib9c/Model/Quest/QuestList.cs
boscohyun: Yes.. I missed that point.🙃

ipdae (changes requested):
CreateQuest execution result may still be different in previous QuestList consturctor.(because it can be occur exception did not occur before). It would be better to create a separate class or method from the existing method

ipdae (approved):

PR #494, opened 2021-07-08
Require minimum 3 transactions for every block

dahlia:


Lib9c/BlockPolicy.cs
```diff
@@ -160,6 +159,24 @@ private InvalidBlockException ValidateBlock(Block<NCAction> block)
                 return null;
             }
 
+            // As a temporary approach to prevent selfish mining (again), we add a new rule
+            // disallowing blocks with less than 10 transactions.  This rule is applied since
+            // 1,870,000th block.  (Note that as of Jul 8, 2021, there are about 1,860,000+ blocks.)
+            // This rule is not applied to blocks (with proofs) made by authorized miners.
+            if (block.Transactions.Count < 10 &&
+                (IgnoreHardcodedIndicesForBackwardCompatibility || block.Index >= 1_870_000) &&
```
longfin, line 167: We may need to change this value by the timeline. 😅

dahlia: 👌🏼 Let me know when the index number needs to be changed.

limebell (approved):

kfangw (approved):

limebell (approved):

ipdae (approved):

Lib9c/BlockPolicy.cs
dahlia: @longfin As we've already passed by [1,870,000th block](https://explorer.libplanet.io/9c-main/block/?0f413ff7fabe6aa150b5989979ae5c0d46bbdc261a893a5f9243ec217ace658b), we need to change the target block index.

longfin:
I adjusted the threshold from 10 to 3 due to 9c mainnet status.

area363 (approved):

PR #507, opened 2021-07-20
Add re-register action

hskim881028:
It is an action to re-register an item that is on sale. 
The product is removed from `OrderDigestListState` and `ShardedShopStateV2` and then added again.

document : [Link](https://www.notion.so/planetarium/8967b29766714a7c8fcfae426d704cba)

### DIFF
--- a/.Lib9c.Tests/Action/ReregisterTest.cs
+++ b/.Lib9c.Tests/Action/ReregisterTest.cs
@@ -1,0 +1,359 @@
+﻿namespace Lib9c.Tests.Action
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Collections.Immutable;
+    using System.Linq;
+    using Bencodex.Types;
+    using Lib9c.Model.Order;
+    using Libplanet;
+    using Libplanet.Action;
+    using Libplanet.Assets;
+    using Libplanet.Crypto;
+    using Nekoyume;
+    using Nekoyume.Action;
+    using Nekoyume.Model;
+    using Nekoyume.Model.Item;
+    using Nekoyume.Model.State;
+    using Serilog;
+    using Xunit;
+    using Xunit.Abstractions;
+    using static SerializeKeys;
+
+    public class ReregisterTest
+    {
+        private const long ProductPrice = 100;
+        private readonly Address _agentAddress;
+        private readonly Address _avatarAddress;
+        private readonly Currency _currency;
+        private readonly AvatarState _avatarState;
+        private readonly TableSheets _tableSheets;
+        private readonly GoldCurrencyState _goldCurrencyState;
+        private IAccountStateDelta _initialState;
+
+        public ReregisterTest(ITestOutputHelper outputHelper)
+        {
+            Log.Logger = new LoggerConfiguration()
+                .MinimumLevel.Verbose()
+                .WriteTo.TestOutput(outputHelper)
+                .CreateLogger();
+
+            _initialState = new State();
+            var sheets = TableSheetsImporter.ImportSheets();
+            foreach (var (key, value) in sheets)
+            {
+                _initialState = _initialState
+                    .SetState(Addresses.TableSheet.Derive(key), value.Serialize());
+            }
+
+            _tableSheets = new TableSheets(sheets);
+
+            _currency = new Currency("NCG", 2, minters: null);
+            _goldCurrencyState = new GoldCurrencyState(_currency);
+
+            var shopState = new ShopState();
+
+            _agentAddress = new PrivateKey().ToAddress();
+            var agentState = new AgentState(_agentAddress);
+            _avatarAddress = new PrivateKey().ToAddress();
+            var rankingMapAddress = new PrivateKey().ToAddress();
+            _avatarState = new AvatarState(
+                _avatarAddress,
+                _agentAddress,
+                0,
+                _tableSheets.GetAvatarSheets(),
+                new GameConfigState(),
+                rankingMapAddress)
+            {
+                worldInformation = new WorldInformation(
+                    0,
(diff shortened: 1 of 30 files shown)


.Lib9c.Tests/Action/ReregisterTest.cs
```diff
@@ -0,0 +1,359 @@
...
+                PreviousStates = state,
+                Signer = _agentAddress,
+                BlockIndex = 0,
+                Rehearsal = true,
+            });
+
+            Assert.Equal(updatedAddresses.ToImmutableHashSet(), nextState.UpdatedAddresses);
+        }
+    }
+}
```
ipdae, line 359:
```suggestion
}

```

.Lib9c.Tests/Action/ReregisterTest.cs
```diff
@@ -0,0 +1,359 @@
...
+                Random = new TestRandom(),
+                Rehearsal = false,
+                Signer = _agentAddress,
+            });
+
+            var reregisterShopAddress = ShardedShopStateV2.DeriveAddress(itemSubType, reregisterOrderId);
+            var nextShopState = new ShardedShopStateV2((Dictionary)nextState.GetState(reregisterShopAddress));
+            Assert.Equal(1, nextShopState.OrderDigestList.Count);
+            Assert.NotEqual(orderId, nextShopState.OrderDigestList.First().OrderId);
+            Assert.Equal(reregisterOrderId, nextShopState.OrderDigestList.First().OrderId);
```
ipdae, line 234: It would be good to check the re-registered order information & avatar inventory item information(like expiredBlockIndex & requiredBlockIndex)

ipdae (changes requested):
It seems that we are missing delete scheduled expirationMail.

Lib9c/Action/Reregister.cs
```diff
@@ -0,0 +1,197 @@
...
+using Serilog;
+using static Lib9c.SerializeKeys;
+using BxDictionary = Bencodex.Types.Dictionary;
+using BxList = Bencodex.Types.List;
+
+namespace Nekoyume.Action
+{
+    [Serializable]
+    [ActionType("reregister")]
+    public class Reregister : GameAction
```
boscohyun, line 24: This class name need to be more specific and meaningful.🙏

Lib9c/Action/Reregister.cs
```diff
@@ -0,0 +1,197 @@
...
+            sw.Restart();
+            states = states
+                .SetState(itemAddress, tradableItem.Serialize())
+                .SetState(reregisterOrderAddress, newOrder.Serialize())
+                .SetState(reregisterShopAddress, reregisterShopState.Serialize());
+            sw.Stop();
+
+            var ended = DateTimeOffset.UtcNow;
+            Log.Verbose("{AddressesHex} Reregister Set ShopState: {Elapsed}", addressesHex, sw.Elapsed);
+            Log.Verbose("{AddressesHex} Reregister Total Executed Time: {Elapsed}", addressesHex, ended - started);
```
boscohyun, line 192: How about use another `Stopwatch` to calculate whole time.

hskim881028: Caching DateTimeOffset.UtcNow in `started` `ended`. `ended - started' seems to be the total time

Lib9c/Action/SellCancellation.cs
```diff
@@ -134,9 +134,13 @@ public override IAccountStateDelta Execute(IActionContext context)
             Order order = OrderFactory.Deserialize(orderDict);
             order.ValidateCancelOrder(avatarState, tradableId);
             ITradableItem sellItem = order.Cancel(avatarState, context.BlockIndex);
-            var shardedShopState = new ShardedShopStateV2(shopStateDict);
-            shardedShopState.Remove(order, context.BlockIndex);
-            states = states.SetState(shardedShopAddress, shardedShopState.Serialize());
+            if (sellItem.RequiredBlockIndex < context.BlockIndex)
```
ipdae, line 137: it would be better use contains instead of blockIndex

ipdae (changes requested):
we need to add test case for expired items(ItemUsable, TradableMaterial) on SellCancellationTest

ipdae:
test failed.

ipdae (approved):

Lib9c/Action/Reregister.cs
hskim881028: To maintain consistency with `sell`or `buy`, I made the action name only as verbs without a subject. If you have a good name, please suggest it.🙏

boscohyun: I think 'Sell' and 'Buy' fall into the category of 'deal', whatever their object. And I think 'Reregister' has a broader meaning than that.
Considering we already have 'SellCancellation', how about 'ReSell'?
Of course.. I think it should have been 'RegisterToShop' rather than 'Sell' in the first place, and I wish the resale was 'ReregisterToShop'.😅

Lib9c/Action/SellCancellation.cs
hskim881028: `ITradableItem sellItem = order.Cancel(avatarState, context.BlockIndex);`
When canceling the order, the `RequiredBlockIndex` of sellItem is changed to `context.BlockIndex`.

Lib9c/Action/Reregister.cs
hskim881028: Let's change it to `resell`. Let's change other places that use reregister as well.

hskim881028: Change completed !

boscohyun (approved):

PR #510, opened 2021-07-21
Add some fields to `Equipment`

boscohyun:
- Add the `optionCountFromCombination` field which is determined at combine action.
- Add the `RequiredCharacterLevel` field which is determined at several actions.

### DIFF
--- a/Lib9c/Model/Item/Equipment.cs
+++ b/Lib9c/Model/Item/Equipment.cs
@@ -6,6 +6,7 @@ using Bencodex.Types;
 using Nekoyume.Model.Stat;
 using Nekoyume.Model.State;
 using Nekoyume.TableData;
+using static Lib9c.SerializeKeys;
 
 namespace Nekoyume.Model.Item
 {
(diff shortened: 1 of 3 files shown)


Lib9c/Model/Item/Equipment.cs
```diff
@@ -6,15 +6,26 @@
...
-        public bool equipped = false;
+        public bool equipped;
         public int level;
+        public int optionCountFromCombination;
+        public readonly int RequiredCharacterLevel;
+
+        /// <summary>
+        /// Do not use in Lib9c. This property only for UI.
+        /// </summary>
+        public int OptionCountFromCombinationForUI => optionCountFromCombination > 0
```
ipdae, line 25: If this property used only in unity, why not implemented in unity?

ipdae (approved):

Lib9c/Model/Item/Equipment.cs
boscohyun: Oh, that's make sense. I'll implement this in other place as extension.

PR #548, opened 2021-08-10
Bump libplanet 0.13.2

riemannulus:


moreal (changes requested):
I modified the title into *~0.13.2* from *~0.13.0*. If you intended to bump libplanet into *0.13.0*, this changes are incorrect 👀 

Also the build seems failed in `Lib9c/DebugPolicy.cs`😭 .

moreal (approved):

PR #563, opened 2021-08-17
Remove .Result properties from `HackAndSlash` (and its likes)

longfin:
This PR removes `.Result` property to lighten the action, reducing the memory load.

### DIFF
--- a/Lib9c/Renderer/ActionRenderer.cs
+++ b/Lib9c/Renderer/ActionRenderer.cs
@@ -42,6 +42,7 @@ namespace Lib9c.Renderer
                 BlockIndex = context.BlockIndex,
                 OutputStates = nextStates,
                 PreviousStates = context.PreviousStates,
+                RandomSeed = context.Random.Seed
             });
         }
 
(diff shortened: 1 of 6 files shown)


longfin:
I postponed removing `RankingBattle.Result` because it had been coupled with other states too tightly. 😓

ipdae (commented):
it would be better to check simulation result using random seed.

longfin:
> it would be better to check simulation result using random seed.

IMHO, if we want to check `StageSimulator.Result` on unit tests, `StageSimulatorTest` is a more proper place than `HackAndSlashTest` because `HackAndSlash` uses only `StageSimulator` and doesn't expose its result anymore.

Lib9c/Renderer/ActionRenderer.cs
```diff
@@ -42,6 +42,7 @@ IAccountStateDelta nextStates
                 BlockIndex = context.BlockIndex,
                 OutputStates = nextStates,
                 PreviousStates = context.PreviousStates,
+                RandomSeed = context.Random.Seed
```
hskim881028, line 45: I get an error in 9c.
`'IRandom' does not contain a definition for 'Seed' and no accessible extension method 'Seed' accepting a first argument of type 'IRandom' could be found (are you missing a using directive or an assembly reference?)`

longfin: We need to update and merge [9c side](https://github.com/planetarium/NineChronicles/pull/627/) too after merging this PR.

hskim881028 (approved):

PR #592, opened 2021-08-27
Remove useless columns of `EquipmentItemSubRecipeSheetV2`

boscohyun:
- We can modify this because the `EquipmentItemSubRecipeSheetV2` is not released yet.


area363 (approved):
Branch needs update :pray:

PR #615, opened 2021-09-08
Introduce `daily_reward6` as `DailyReward`

boscohyun:
SSIA

### DIFF
--- a/.Lib9c.Tests/Action/DailyRewardTest.cs
+++ b/.Lib9c.Tests/Action/DailyRewardTest.cs
@@ -106,17 +106,11 @@ namespace Lib9c.Tests.Action
 
             var nextState = ExecuteInternal(previousStates, 1800);
             var nextGameConfigState = nextState.GetGameConfigState();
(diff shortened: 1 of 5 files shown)


.Lib9c.Tests/Action/DailyRewardTest.cs
```diff
@@ -79,6 +79,9 @@ public void Rehearsal()
             var updatedAddresses = new List<Address>
             {
                 _avatarAddress,
+                _avatarAddress.Derive(LegacyInventoryKey),
+                _avatarAddress.Derive(LegacyWorldInformationKey),
+                _avatarAddress.Derive(LegacyQuestListKey),
```
longfin, line 84: IMO, it would be better if we have the assertion about re-creating from output states in `Execute()` test case as regression.

longfin (approved):

hskim881028 (approved):

boscohyun:
@hskim881028 @longfin I fixed unit test for `DailyReward` at 5df5f63. Please review this.

longfin (approved):

PR #634, opened 2021-09-14
Added booster to HAS action

hskim881028:
- Added booster to `HackAndSlash` action

Related Links 
- https://app.asana.com/0/958521740385861/list
- https://www.notion.so/planetarium/d6d58782022347d1b1a3a7fdff1d4af5#0650562bc96148e893923ef0e9ec7754


Lib9c/Action/HackAndSlash.cs
```diff
@@ -156,15 +159,22 @@ public override IAccountStateDelta Execute(IActionContext context)
             Log.Verbose("{AddressesHex}HAS get CostumeStatSheet: {Elapsed}", addressesHex, sw.Elapsed);
             sw.Restart();
 
-            if (avatarState.actionPoint < stageRow.CostAP)
+            if (playCount <= 0)
```
ipdae, line 162: we need check stage already cleared when playCount > 1

sonohoshi:
Can add `playCount` field in BattleLog class?
I want to use it in 9c - Stage.cs. like this.
```cs
var countableItem = new CountableItem(item, _battleLog.playCount); // In development branch, Stage.cs, 930 line
```
In addition to this, there are many situations to refer to `playCount`.

Lib9c/Action/HackAndSlash.cs
hskim881028: Added. Please review again 🙏🏻 @ipdae

ipdae:
test failed

ipdae (commented):
i think it would be good to add 2 test case
- execution hackAndSlash use booster(count > 1) 
- unit test for new get reward

hskim881028:
> i think it would be good to add 2 test case
> 
> * execution hackAndSlash use booster(count > 1)
> * unit test for new get reward

Added test code. considering the quest reward, the test code got longer. 😓 @ipdae

ipdae (approved):

PR #509, opened 2021-07-20
Reduce CheckSigner execution time.

ipdae:
- resolve #506

PR #636, opened 2021-09-15
TrasnferAsset2

### LINKED ISSUE #623: Prevent asset transfer to inactivated account
### Background
There are reports that instead of swapping NCG to WNCG, users sent the NCG to their Ethereum address in the Nine Chronicles network

### Problem
The recipient cannot generate any valid transaction without activation code. So the received NCG is locked.

### Preventive measure
Asset transfer such as NCG transfer should check the recipient is activate or not. If the recipient is inactivated, the transaction should be failed

### Solution for the Problem
It should be debated first within dev/ops in Planetarium team

riemannulus:
This PR resolve #623

### DIFF
--- a/.Lib9c.Tests/Action/TransferAssetTest.cs
+++ b/.Lib9c.Tests/Action/TransferAssetTest.cs
@@ -10,8 +10,10 @@ namespace Lib9c.Tests.Action
     using Libplanet;
     using Libplanet.Action;
     using Libplanet.Assets;
+    using Libplanet.Crypto;
     using Nekoyume;
     using Nekoyume.Action;
+    using Nekoyume.Model;
     using Nekoyume.Model.State;
     using Xunit;
 
@@ -44,12 +46,11 @@ namespace Lib9c.Tests.Action
         [Fact]
         public void Execute()
         {
-            var activatedAddress = new ActivatedAccountsState().AddAccount(_recipient);
             var balance = ImmutableDictionary<(Address, Currency), FungibleAssetValue>.Empty
                 .Add((_sender, _currency), _currency * 1000)
                 .Add((_recipient, _currency), _currency * 10);
             var state = ImmutableDictionary<Address, IValue>.Empty
-                .Add(Addresses.ActivatedAccount, activatedAddress.Serialize());
+                .Add(_recipient.Derive(ActivationKey.DeriveKey), true.Serialize());
             var prevState = new State(
                 state: state,
                 balance: balance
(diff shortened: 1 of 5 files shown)


kfangw:
btw, it is not easy for me to check the diff between old one and the patch. 
Could you explain with the pin point?

riemannulus:
> 
> 
> btw, it is not easy for me to check the diff between old one and the patch.
> Could you explain with the pin point?

https://github.com/planetarium/lib9c/pull/636/files#diff-af422b0b2b534088a52121cf2e8094a42c2ef5e73db0a97fc1dd924e46ecc052R87-R96

Get all activated account list from chain, check is list contained recipient address. if not, throw error. if yes, send NCG.

Lib9c/Action/TransferAsset.cs
```diff
@@ -84,6 +84,17 @@ public override IAccountStateDelta Execute(IActionContext context)
                 throw new InvalidTransferRecipientException(Sender, Recipient);
             }
 
+            if (state.GetState(Addresses.ActivatedAccount) is Dictionary asDict)
```
kfangw, line 87: 👍

Lib9c/Action/TransferAsset.cs
```diff
@@ -84,6 +84,17 @@ public override IAccountStateDelta Execute(IActionContext context)
...
+            {
+                var activatedAccountsState = new ActivatedAccountsState(asDict);
+                var activatedAccounts = activatedAccountsState.Accounts;
+                if (activatedAccounts.Count != 0
+                    && !activatedAccounts.Contains(Recipient))
+                {
+                    throw new InvalidTransferRecipientException(Sender, Recipient);
```
kfangw, line 94: I am not sure. But it would be better to define the exception for this case.

kfangw:
I don't see any relevant test case for this action. Do we need the test case,  which being executed in the CI, for newly introduced action type?

Lib9c/Action/TransferAsset.cs
riemannulus: It was vague to me. `InvalidTransferRecipientExecption` is perfect fit here too. but this error also mean sender and recipient are same. so, what is your opinion? If create new exception type, what name would be better?

kfangw: Someone who catch the exception does not know the reason why the exception thrown; due to the recipient is inactivated. 

`I am not sure` means I have no idea how the exception handled by the someone. If the exception is suitable for this case in your thought, it is good to me too.

riemannulus: I clearly make new exception for this case. c72621a

.Lib9c.Tests/Action/TransferAssetTest.cs
```diff
@@ -227,6 +231,38 @@ public void ExecuteWithMinterAsRecipient()
             Assert.Equal(_recipient, ex.Recipient);
         }
 
+        [Fact]
+        public void ExecuteWithUnactivatedRecipient()
```
kfangw, line 235: ❤️

Lib9c/Action/TransferAsset.cs
longfin: AFIK, we split monolithic `ActivatedAccountsState` to many derived states [here](https://github.com/planetarium/lib9c/pull/509). so I think we should check derived address (like [this](https://github.com/planetarium/lib9c/pull/509/files#diff-466f48999a3b02d0666d42edcec40b3f2712742275766d488c60875746134d6bR131-R144)) instead of global activate states. cc @ipdae

kfangw (approved):
LGTM.
But, I strongly recommend you to get another approvals by others too before merge it.

Lib9c/Action/TransferAsset.cs
ipdae: from `ActivateAccount` https://github.com/planetarium/lib9c/blob/rc-v100077/Lib9c/Action/ActivateAccount.cs#L42

riemannulus:
PTAL @ipdae @longfin

ipdae (approved):

PR #640, opened 2021-09-16
Policy packaging refactor.

### LINKED ISSUE #641: Policy specifications
This is to be used as a guideline in implementing pluggable policies. This tries to conform to existing rules without violating any, at least for the blocks with the old policies applied anyway, while also try to make existing policies expandable.

# Use Cases

There are two separate use cases both for authorized mining and permissioned mining policies.

- Control the behvaior at an upper level, namely headless.
  - Although strictly not necessary, this is to prevent unnecessary mining attempts.
  - Something akin to `isAllowedToMine()` is called to check to see if next block can be mined.
- Control the behavior at a lower level, namely libplanet.
  - In particular, `getNextBlockDifficulty()` is used to get proper difficulty for the next block.
  - As libplanet itself is agnostic about custom authorized mining and/or permissioned mining rules, the logic for validation of miners must be bound and injected into an implementation of `IBlockPolicy<T>.ValidateNextBlock()` method.

# Rule Precedences

Although not explicitly stated, genesis mining, for a lack of a better term, is considered a special rule in its own right. Several rules may overlap for a particular index. However, the rules are *not required* to be compatible with each other, and for a good reason. But, as there is a need to resolve as to which *single* rule should apply, at least the precedence between the type of rules must be specified beforehand. With this, we can formulate the following.

- Genesis mining rule takes precedence over any authorized mining rule and permissioned mining rule.
- Authorized mining rule takes precedence over any permissioned mining rule.

For example, for a block of index `0`, genesis mining rule overrules over authorized mining and permissioned mining rules. In particular, this affects its target difficulty to be `0` (although whether a hardcoded difficulty of `0` for genesis blocks is appropriate is a whole different matter) *and* allows it to be mined by *any* miner<sup>[1](#f1)</sup>.

To prevent the chain grinding to a halt due to no miners being allowed to mine, if none of the rules apply, the default policy behavior must be to allow *any* miner to mine. This requirement is pretty obvious in itself, but its implementation is not. Both authorized mining and permissioned mining rules are tied to certain indices in which they should apply, but in all cases, for both rules, lists of miners allowed to be mined should *never* be empty <sup>[2](#f2)</sup>.

To summarize, given block index `i`, we have the following.

1. If `i` is `0`, apply genesis mining rule. Otherwise move to 2.
2. If `i` is a target index for authorized mining rule, apply authorized mining rule. Otherwise move to 3.
3. If `i` is a target index for permissioned mining rule, apply permissioned mining rule. Otherwise move to 4.
4. Apply default mining rule where any miner can mine.

----

Footnotes:

<a name="#f1">1</a>: This is a design choice, but not a necessary condition. There are still several factors at play here. As with authorized mining rule, which depends on the record of the chain, as there is no record to speak of as yet, authorized mining rule cannot apply to the genesis block, at least not in a natural way. As for permissioned mining rule, this *can* be applied to a genesis block, as content for permissioned mining rule is prescribed in the code itself. However, simply overriding with permissioned mining rule upsets the precedences mentioned above. Additional *tricks* are required if we are to keep the precedences.

<a name="#f2">2</a>: Again, this is also a design choice to keep things simple.

greymistcube:
Closes #641. `BlockPolicy` is redesigned to conform to the design outlined in #641.

This is a continuation of #632.

This attempts to decouple tangled dependencies as much as possible and moved most implementations of policy components from `BlockPolicy` to `BlockPolicySource` as `static` methods.

Notes:
- As far as I know, there should be no behavioral change so far.
- Each hard coded index are incremented by `1`, since it seemed more natural to match the starting index not the last "obsolete" index.
- Might also move `GetNextBlockDifficulty()` to `BlockPolicySource` in the near future after modularizing the said component in [libplanet](https://github.com/planetarium/libplanet) first.
- Might be somewhat slower as `AuthorizedMinersState` is actually read on-the-fly from a `BlockState<T>` instance whenever needed.

----

Addendum:
- `IsAllowedToMine()` method is added to `BlockPolicy` to give control to headless whether to mine or not.

----

- <https://github.com/planetarium/lib9c/pull/640>
- <https://github.com/planetarium/NineChronicles.Headless/pull/740>


greymistcube:
Apparently removed `PermissionedMiningPolicy` and `AuthorizedMinersState` properties are needed to control [headless](https://github.com/planetarium/NineChronicles.Headless) behavior. ☹️

Lib9c/Policy/BlockPolicySource.Utils.cs
```diff
@@ -0,0 +1,380 @@
...
+
+        /// <summary>
+        /// Checks if permissioned mining policy applies to given block index.
+        /// </summary>
+        /// <remarks>
+        /// An implementation should be agnostic about other policies affecting the same index.
+        /// Policy overruling between different policies should be handled elsewhere.
+        /// </remarks>
+        internal static bool IsPermissionedMiningBlockIndexRaw(
+            BlockChain<NCAction> blockChain,
```
longfin, line 318: `blockChain` doesn't seem to be used. Is it a placeholder for matching types with other interfaces?

Lib9c/Policy/BlockPolicySource.Utils.cs
```diff
@@ -0,0 +1,380 @@
...
+                else
+                {
+                    throw new ArgumentException(
+                        $"Result of {nameof(GetAuthorizedMinersState)} cannot be null.");
+                }
+            }
+            else
+            {
+                throw new ArgumentException(
+                    $"Result of {nameof(IsAuthorizedMiningBlockIndex)} must be true.");
```
longfin, line 306: It seems that checking the block index in this function is too much work.

Or, I think it's okay to return `false` here instead of raising `ArgumentException`. (then, we can remove branch [here](https://github.com/planetarium/lib9c/pull/640/files#diff-0e19782ab5996884e5ecd2b582cce259a2691938e566b1e7f09557abc8bea8a0R238-R241))

Lib9c/Policy/BlockPolicySource.Utils.cs
```diff
@@ -0,0 +1,380 @@
...
+                else
+                {
+                    throw new ArgumentException(
+                        $"Argument {nameof(permissionedMiningPolicy)} cannot be null.");
+                }
+            }
+            else
+            {
+                throw new ArgumentException(
+                    $"Result of {nameof(IsPermissionedMiningBlockIndexRaw)} must be true.");
```
longfin, line 369: Same as `IsAuthorizedToMine()`...

Lib9c/Policy/BlockPolicySource.Utils.cs
```diff
@@ -0,0 +1,380 @@
...
+using Libplanet.Blockchain;
+using Libplanet.Blocks;
+using Libplanet.Tx;
+using Nekoyume.Action;
+using Nekoyume.Model.State;
+using NCAction = Libplanet.Action.PolymorphicAction<Nekoyume.Action.ActionBase>;
+
+namespace Nekoyume.BlockChain.Policy
+{
+    // Collection of helper methods not directly used as a pluggable component.
```
longfin, line 15: IMO some of these helper methods can be `private`...

Lib9c/Policy/BlockPolicySource.Utils.cs
greymistcube: It's purely a placeholder for consistency, a leftover artifact before deciding whether to either have both `AuthorizedMinersState` and `PermissionedMiningPolicy` in the ledger or as a hardcoded policy. Should be removed once `AuthorizedMinersState` is changed to use hardcoded data.

Lib9c/Policy/BlockPolicySource.Utils.cs
greymistcube: The method is under the assumption that `IsAuthorizedMiningBlockIndex()` is `true` in the first place, so index check is not strictly necessary. This is purely for clarity and/or defensive reasons. I agree the code is pretty icky at the moment, but I failed to come up with a better solution.

- Asking whether a miner is authorized to mine on a non-target block index is non-sensical.
  - Can be answered either way as vacuously `true` or vacuously `false`, but this seemed dangerous.
- I wanted to have `IsAuthorizedToMine()` as a separate method, but perhaps not having it as a separate method may be better.
- Using ternary logic may be most natural with `bool?` as its return type. Wasn't sure it's a good practice to do so.
- This is most likely to get overhauled in a future PR when variable sub-policies feature finally gets implemented, so I didn't want to spend *too* much time on this issue. 🙃

Lib9c/Policy/BlockPolicySource.Utils.cs
greymistcube: Either most should be `private` or `public`, but I guess `internal` might not be right. Initially I thought each method should be tested separately, but didn't want to expose them as `public`. Then again, the implementation detail is not quite finalized, so didn't really want to write tests which would be mostly overwritten in near future. ☹️

Lib9c/Policy/BlockPolicySource.Utils.cs
greymistcube: [Yup](https://github.com/planetarium/lib9c/pull/640/files/a9b022cf74fffbe44c2affe00452df9fb6ecef8a#r715472026).

Lib9c/Policy/BlockPolicySource.Utils.cs
greymistcube: Eh... changing things to `private` with unused parameters gives me build errors... 😮

greymistcube:
@longfin 
Although I agree with most of your comments, I'll try to update with some changes, but only with minor ones, as there are already two other branches being worked on top of this one.

As for my excuse, this is more of an intermediary checkpoint, with large chunks of design not yet finalized, with mostly an *explicit* goal of moving the responsibility of constructing `BlockPolicy` object to `BlockPolicySource`. I promise more things will get cleaned up with further update. 😶

Lib9c/Policy/BlockPolicySource.Utils.cs
longfin: > Can be answered either way as vacuously true or vacuously false, but this seemed dangerous.

- Basically, I think it's good to ask for a simple case if possible and to use a combination of these when necessary. 
- This predicate appears to be a subfunction that is only called from `IsAllowedToMineRaw()`. If so, it just looks up if there is a miner (miner) corresponding to the state obtained with `GetAuthorizedMinersState` and returns the result. (If the name of this predicate was `BelongsAuthorizedMinerState()`, it wouldn't look too dangerous without index checking.)
  - Of course, we may have questions like "is such a predicate really necessary?", and I agree with that. If so, we can define its return as just `true` or `false` (no exception). and abstract `IsAllowedToMineRaw()` as like...
  ```csharp
          internal static bool IsAllowedToMineRaw(
            BlockChain<NCAction> blockChain,
            Address miner,
            long index,
            Func<BlockChain<NCAction>, long, bool> isPermissionedMiningBlockIndex,
            Func<BlockChain<NCAction>, Address, long, bool> isPermissionedToMine)
        {
            // For genesis blocks, any miner is allowed to mine.
            if (index == 0)
            {
                return true;
            }

            return IsAuthorizedToMine(blockChain, miner, index) || isPermissionedToMine(blockChain, miner, index); // ???
        }
  ```
  - Oh, it's a matter. we can't express vanilla state (no auth target and/or no permissioned target block) well. 
  - Maybe instead of a bool-based predicate, it makes more sense to use a checker-based approach that raises an exception... but it would be a lot of work to fix it all. understand.

longfin (approved):

greymistcube:
Eh, aside from removing blatantly redundant piece of code, had to rolled back changes. While trying to streamline `IsAuthorizedMiningBlockIndex()`, `IsAuthorizedToMine()`, `IsPermissionedMiningBlockIndex()`, and `IsPermissionedToMine()`, due to inconsistent binding order, three more methods were getting added with non-trivial redirections just to get rid of the branching paths. This will be addressed in #646 if possible.

PR #652, opened 2021-09-28
add offset argument to `Lib9c.Benchmarks`

area363:
This PR adds an `offset` argument to `Lib9c.Benchmarks` so that it has the option to execute blocks from the offset value. 

## Usage
```
dotnet run --project .\.Lib9c.Benchmarks\Lib9c.Benchmarks.csproj [store-path] [limit] [offset]
```
- Example:
```
dotnet run --project .\.Lib9c.Benchmarks\Lib9c.Benchmarks.csproj C:\Users\user\9c-main 10 20
- This will execute blocks from #20 to #29.

dotnet run --project .\.Lib9c.Benchmarks\Lib9c.Benchmarks.csproj C:\Users\user\9c-main 10
- This will execute blocks from #0 to #9 like the original benchmark because an offset value is not provided.
```

**Note)** 
 When using the `offset value`, the blockchain store **MUST CONTAIN STATES** for those blocks or else, a `KeyNotFoundException`(https://github.com/planetarium/libplanet/blob/main/Libplanet.RocksDBStore/RocksDBKeyValueStore.cs#L30-L31) will occur.

### DIFF
--- a/.Lib9c.Benchmarks/Program.cs
+++ b/.Lib9c.Benchmarks/Program.cs
@@ -35,7 +35,13 @@ namespace Lib9c.Benchmarks
 
             string storePath = args[0];
             int limit = int.Parse(args[1]);
-            Log.Logger = new LoggerConfiguration().MinimumLevel.Debug().WriteTo.Console().CreateLogger();
+            int offset = 0;
+
+            if (args.Length == 3){
+                offset = int.Parse(args[2]);
+            }
+
+            Log.Logger = new LoggerConfiguration().MinimumLevel.Verbose().WriteTo.Console().CreateLogger();
             Libplanet.Crypto.CryptoConfig.CryptoBackend = new Secp256K1CryptoBackend<SHA256>();
             var policySource = new BlockPolicySource(Log.Logger, LogEventLevel.Verbose);
             IBlockPolicy<NCAction> policy =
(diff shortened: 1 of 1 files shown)


.Lib9c.Benchmarks/Program.cs
```diff
@@ -63,9 +69,15 @@ static void Main(string[] args)
             var stateStore = new TrieStateStore(stateKeyValueStore, stateRootKeyValueStore);
             var chain = new BlockChain<NCAction>(policy, stagePolicy, store, stateStore, genesis);
             long height = chain.Tip.Index;
-            BlockHash[] blockHashes = limit < 0
-                ? chain.BlockHashes.SkipWhile((_, i) => i < height + limit).ToArray()
-                : chain.BlockHashes.Take(limit).ToArray();
+            if (offset + limit > (int)height)
```
dahlia, line 72: The case where `limit` is less than zero should be also considered.

.Lib9c.Benchmarks/Program.cs
```diff
@@ -35,7 +35,13 @@ static void Main(string[] args)
 
             string storePath = args[0];
             int limit = int.Parse(args[1]);
-            Log.Logger = new LoggerConfiguration().MinimumLevel.Debug().WriteTo.Console().CreateLogger();
+            int offset = 0;
+
+            if (args.Length == 3){
```
dahlia, line 40:
```suggestion
            if (args.Length >= 3)
            {
```

ipdae (approved):

PR #646, opened 2021-09-23
Bump libplanet

greymistcube:
Closes #635.

Accommodates API changes in https://github.com/planetarium/libplanet/pull/1495.

This is a continuation of #640.

This completes binding of arguments from `BlockPolicySource` for building custom `BlockPolicy`.

----

- <https://github.com/planetarium/libplanet/pull/1495>
- <https://github.com/planetarium/lib9c/pull/646>

PR #653, opened 2021-09-28
Hardcoded authorized mining

### LINKED ISSUE #645: Remove `ignoreHardcodedPolicies` parameter
This is only suitable for a single instance of policy switch and is not suitable/compatible with planned pluggable policy overhaul. Alongside policy refactoring, this should be removed with substantial test code rewrites.

### LINKED ISSUE #650: Use hardcoded `AuthorizedMinersState` instead of reading from the ledger.
From a security and/or a decentralization perspective, reading a consensus policy from the ledger may be better. However, due to `PermissionedMiningPolicy` being implemented through hard-coding, I'd argue whatever upside with having `AuthorizedMinersState` read from the ledger is offset more than enough by having to deal with discrepancy in argument binding order, general interface, and lack of well thought out initial design.

Ultimately, moving the record of `PermissionedMiningPolicy` to the ledger may be desirable, but the current codebase lacks the appropriate support structure for this to be easily implemented, that is, without further contaminating the codebase.

Weighing the costs and benefits of having `AuthorizedMinersState` on the ledger, I suggest, at least for now, we remove it from the ledger.

P.S. Although, considering the backward compatibility issue, if we are to ever decide to roll back and use the record on the ledger, this would cause much headache in the future, but moving `PermissionedMiningPolicy` on to the ledger would cause similar problems anyway. 🙃

greymistcube:
Closes #645 and closes #650.

This is a continuation of #646.

As addressed in the referenced issues:
- `AuthorizedMinersState` is no longer used as a policy, but backward compatible `AuthorizedMiningPolicy` is newly implemented.
- Parameter `ignoreHardcodedPolicies` is removed. As old behavior of `ignoreHardcodedPolicies` was not clear enough.
  - This applied to only certain sub-policies but not all without clear reasoning.
  - Even when applied, there was no consistency on how it is applied to a specific policy.
  - Also there was general disregard as to how `BlockPolicy` as a whole would behave as `ignoreHardcodedPolicies` could affect multiple interdependent sub-policies.

### DIFF
--- a/.Lib9c.Tests/BlockPolicyTest.cs
+++ b/.Lib9c.Tests/BlockPolicyTest.cs
@@ -137,17 +137,22 @@ namespace Lib9c.Tests
             );
 
             var blockPolicySource = new BlockPolicySource(Logger.None);
-            IBlockPolicy<PolymorphicAction<ActionBase>> policy = blockPolicySource.GetPolicy(10000, 100);
+            AuthorizedMiningPolicy authorizedMiningPolicy = new AuthorizedMiningPolicy(
+                startIndex: 0,
+                endIndex: 10,
+                interval: 5,
+                miners: new[] { authorizedMinerPrivateKey.ToAddress() }.ToHashSet());
+            IBlockPolicy<PolymorphicAction<ActionBase>> policy = blockPolicySource.GetPolicy(
+                minimumDifficulty: 10000,
+                maxTransactionsPerBlock: 100,
+                ignoreHardcodedPolicies: false,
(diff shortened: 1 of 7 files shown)


greymistcube:
~Pretty lousy design at the moment, but I promise things will get cleaned up in a couple of days.~ 🙃 

As per old issues:
- Make more methods `private`.
- Remove redundant logic and streamline validation.

Only the latter is somewhat resolved. ☹️

greymistcube:
Diverged from #659 by mistake. Probably not worth the time to fix git branching issues.

PR #663, opened 2021-10-01
Catch up Libplanet upstream's API changes

dahlia:
*Note that this is a revival of <https://github.com/planetarium/lib9c/pull/657>, which was merged with nobody's approval by mistake and immediately reverted (<https://github.com/planetarium/lib9c/pull/662>) in the end.*

----

This PR tries to catch up Libplanet's recent API changes made by the PR <https://github.com/planetarium/libplanet/pull/1492>.

### DIFF
--- a/.Lib9c.Tests/MemoryKeyValueStore.cs
+++ b/.Lib9c.Tests/MemoryKeyValueStore.cs
@@ -1,0 +1,81 @@
+namespace Lib9c.Tests
+{
+    using System.Collections.Generic;
+    using Libplanet.Store.Trie;
+
+    public class MemoryKeyValueStore : IKeyValueStore
+    {
+        public MemoryKeyValueStore()
+            : this(new Dictionary<byte[], byte[]>())
+        {
+        }
+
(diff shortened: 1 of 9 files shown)


longfin (approved):

.Lib9c.Tests/MemoryKeyValueStore.cs
```diff
@@ -0,0 +1,81 @@
...
+                        return false;
+                    }
+                }
+
+                return true;
+            }
+
+            public int GetHashCode(byte[] obj)
+            {
+                int hash = 17;
```
greymistcube, line 68: This is just a random constant right?

dahlia: 17 and 31 are two different prime numbers I (more exactly, Rider) arbitrary chose.  It's okay even if they are any other prime numbers.  Frankly, I didn't think much about it, because it's what Rider generated for me.  See also: <https://stackoverflow.com/a/3613423>.

PR #665, opened 2021-10-05
Merge `rc-v100080` into `development`

boscohyun:
SSIA


Lib9c/Policy/BlockPolicySource.cs
```diff
@@ -51,6 +51,10 @@ public partial class BlockPolicySource
 
         // FIXME: Should be finalized before release.
         public const int MaxTransactionsPerSignerPerBlock = 4;
+        
+        public const long V100080ObsoleteIndex = 2_448_000;
+        
+        public const long V100081ObsoleteIndex = 2_500_000;
```
ipdae, line 57: It would be better to leave a comment for update required at actual release.

boscohyun: Applied.

area363 (approved):

PR #703, opened 2021-10-31
Max transactions removed from `ReorgMiner.MineBlockAsync()`

greymistcube:
Unless there is a specific reason, max transactions per block should be controlled by policy.


riemannulus (approved):
👍

PR #713, opened 2021-11-04
Clean up renderers

boscohyun:
- Clean up `BlockRenderer` and `ActionRenderer`

### DIFF
--- a/Lib9c/Renderer/BlockRenderer.cs
+++ b/Lib9c/Renderer/BlockRenderer.cs
@@ -34,14 +34,5 @@ namespace Lib9c.Renderer
 
         public void RenderReorgEnd(NCBlock oldTip, NCBlock newTip, NCBlock branchpoint) =>
             ReorgEndSubject.OnNext((oldTip, newTip, branchpoint));
-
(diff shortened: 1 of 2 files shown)


Lib9c/Renderer/BlockRenderer.cs
```diff
@@ -26,39 +26,13 @@ public class BlockRenderer : IRenderer<NCAction>
...
-        public IObservable<(NCBlock OldTip, NCBlock NewTip)> EveryBlock() =>
-            BlockSubject.AsObservable();
-
-        public IObservable<(NCBlock OldTip, NCBlock NewTip, NCBlock Branchpoint)>
-            EveryReorg() =>
-            ReorgSubject.AsObservable();
-
-        public IObservable<(NCBlock OldTip, NCBlock NewTip, NCBlock Branchpoint)>
-            EveryReorgEnd() =>
-            ReorgEndSubject.AsObservable();
```
area363, line 62: Are these no longer used?

boscohyun: These methods does `Subject.AsObservable()` but is unnecessary.

area363 (approved):

PR #742, opened 2021-11-23
Introduce MessagePack-Csharp

ipdae:
- resolve https://github.com/planetarium/NineChronicles.Headless/issues/860
- Use MessagePack-Chsarp for grpc serialize formatter.
- Implement NCActionEvaluation for grpc communicate.
  - Use PolyMorphicAction
  - Extra field for action result state(use case in RankingBattle, Buy)

### DIFF
--- a/Lib9c/Formatters/PublicKeyFormatter.cs
+++ b/Lib9c/Formatters/PublicKeyFormatter.cs
@@ -1,0 +1,23 @@
+using System;
+using System.Buffers;
+using Libplanet.Crypto;
+using MessagePack;
+using MessagePack.Formatters;
+
+namespace Lib9c.Formatters
+{
+    public class PublicKeyFormatter : IMessagePackFormatter<PublicKey>
+    {
+        public void Serialize(ref MessagePackWriter writer, PublicKey value, MessagePackSerializerOptions options)
+        {
+            writer.Write(value.Format(true));
+        }
+
+        public PublicKey Deserialize(ref MessagePackReader reader, MessagePackSerializerOptions options)
+        {
+            options.Security.DepthStep(ref reader);
+
+            return new PublicKey(reader.ReadBytes()?.ToArray() ?? throw new InvalidOperationException());
+        }
+    }
+}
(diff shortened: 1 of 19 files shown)


Lib9c/Lib9c.csproj
```diff
@@ -16,6 +16,9 @@
   <ItemGroup>
     <PackageReference Include="DecimalMath.DecimalEx" Version="1.0.2" />
     <PackageReference Include="JetBrains.Annotations" Version="2021.1.0" />
+    <PackageReference Include="MessagePack" Version="2.3.85" />
+    <PackageReference Include="MessagePack.Annotations" Version="2.3.85" />
+    <PackageReference Include="MessagePackAnalyzer" Version="2.3.85" />
```
longfin, line 21: How about excluding this dependency from output assembly like `Libplanet.Analyzer`? (see also. https://github.com/planetarium/lib9c/pull/742/files#diff-934feb0de25b514ecae56d97457a94b5fb36f0c7f997165bc8f34d1f9b840b4eR36)

Lib9c/Formatters/PublicKeyFormatter.cs
```diff
@@ -0,0 +1,23 @@
...
+        public void Serialize(ref MessagePackWriter writer, PublicKey value, MessagePackSerializerOptions options)
+        {
+            writer.Write(value.Format(true));
+        }
+
+        public PublicKey Deserialize(ref MessagePackReader reader, MessagePackSerializerOptions options)
+        {
+            options.Security.DepthStep(ref reader);
+
+            return new PublicKey(reader.ReadBytes()?.ToArray() ?? throw new InvalidOperationException());
```
longfin, line 20: How about adding information about bytes from `reader.ReadBytes()` to the exception?

longfin (commented):
I believe it would be better if we move `Lib9c.Foramtter` to `NineChronicles.RPC.Shared` because it relates with only RPC. (even if future)

area363 (approved):

longfin (approved):

PR #746, opened 2021-11-24
Introduce CreatePendingActivations action

longfin:
This PR adds a new `CreatePendingActivations` action for bulk inserting many `PendingActivationState`s at once.

### DIFF
--- a/.Lib9c.Tests/Action/ActionEvaluationTest.cs
+++ b/.Lib9c.Tests/Action/ActionEvaluationTest.cs
@@ -96,6 +96,7 @@ namespace Lib9c.Tests.Action
         [InlineData(typeof(Sell))]
         [InlineData(typeof(SellCancellation))]
         [InlineData(typeof(UpdateSell))]
+        [InlineData(typeof(CreatePendingActivations))]
         public void Serialize_With_MessagePack(Type actionType)
         {
             var action = GetAction(actionType);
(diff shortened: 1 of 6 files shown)


ipdae (approved):
it would be better add test case in messagepack serialize. https://github.com/planetarium/lib9c/blob/development/.Lib9c.Tests/Action/ActionEvaluationTest.cs#L99

area363 (approved):

longfin:
> it would be better add test case in messagepack serialize. https://github.com/planetarium/lib9c/blob/development/.Lib9c.Tests/Action/ActionEvaluationTest.cs#L99

I've added an `InlineData` for new action and found some bug related that. but there is still same bug on Libplanet side. 

So I'll turned this issue to draft and fix on Libplanet side too. cc @planetarium/libplanet

ipdae (approved):

ipdae (changes requested):
``` X Lib9c.Tests.Action.CreatePendingActivationsTest.PlainValue [13ms]
  Error Message:
   System.NullReferenceException : Object reference not set to an instance of an object.
  Stack Trace:
     at Lib9c.Tests.Action.CreatePendingActivationsTest.PlainValue() in /home/runner/work/lib9c/lib9c/.Lib9c.Tests/Action/CreatePendingActivationsTest.cs:line 58```

failed test

ipdae (approved):

PR #773, opened 2021-12-07
Bug fix/deterministic of ranking state

boscohyun:
SSIA

### DIFF
--- a/Lib9c/Model/State/RankingState.cs
+++ b/Lib9c/Model/State/RankingState.cs
@@ -66,20 +66,5 @@ namespace Nekoyume.Model.State
             return ((Bencodex.Types.Dictionary)base.Serialize())
                 .SetItem("ranking_map", rankingMapValue);
         }
-
-        [Obsolete("Since it contains a deterministic problem, replace it with the `Serialize()` method that operates deterministically like this logic.")]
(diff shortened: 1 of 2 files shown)


Lib9c/Model/State/RankingState.cs
```diff
@@ -53,7 +53,22 @@ public Address UpdateRankingMap(Address avatarAddress)
...
+                new KeyValuePair<IKey, IValue>(
+                    (Bencodex.Types.Binary)pair.Key.Serialize(),
+                    new Bencodex.Types.List(pair.Value
+                        .OrderBy(e => e.GetHashCode())
+                        .Select(e => e.Serialize())))));
+            return ((Bencodex.Types.Dictionary)base.Serialize())
+                .SetItem("ranking_map", rankingMapValue);
+        }
+
+        [Obsolete("Since it contains a deterministic problem, replace it with the `Serialize()` method that operates deterministically like this logic.")]
```
longfin, line 70: I guess if there isn't any usage for `SerializeV1()` and new `Serialize()` acts as previous, we don't need to preserve the previous code.

boscohyun: @longfin I applied this.

dahlia (approved):

PR #789, opened 2021-12-15
Revert lazy state

longfin:
This PR partially reverts #775 due to performance degradation.


hskim881028 (approved):
It was mainly checked whether it was properly restored based on [PR](https://github.com/planetarium/lib9c/pull/775/files).
That [part](https://github.com/longfin/lib9c/blob/feature/revert-lazy-state/Lib9c/Model/State/ArenaInfo.cs#L93) wasn't reversed, but I thought there would be no problem, so I pressed Approve .👍🏻

PR #798, opened 2021-12-29
Increase capacity of ranking map state

longfin:
This PR bumps `RankingState` model and `CreateAvatar` action to increase `RankingState.RankingMapCapacity` from 100 to 150.

### DIFF
--- a/Lib9c/Policy/BlockPolicySource.cs
+++ b/Lib9c/Policy/BlockPolicySource.cs
@@ -62,6 +62,8 @@ namespace Nekoyume.BlockChain.Policy
         /// </summary>
         public const long V100093ObsoleteIndex = 3_050_000;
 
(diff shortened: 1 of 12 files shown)


moreal (approved):

moreal (approved):

Lib9c/Policy/BlockPolicySource.cs
```diff
@@ -62,6 +62,8 @@ public partial class BlockPolicySource
         /// </summary>
         public const long V100093ObsoleteIndex = 3_050_000;
 
+        public const long V100095ObsoleteIndex = 3_317_632;
```
moreal, line 65: As a memo, 3077632 block was mined at Thu Dec 30 2021 01:45:41 GMT+0900.

PR #805, opened 2022-01-05
feat: introduce `RenewAdminState` action

moreal:
## Background

The `AdminState` will be soon expired after the 3153600 block number. When the `AdminState` was expired, some actions like `PatchTableSheet` and `CreatePendingAccount` will not work. But the Proof-of-Stake or other system is not built yet.

## Description

So this pull request introduces `RenewAdminState` action which renew `AdminState.ValidUntil` field. It requires Admin permission so it should be executed before the block 3153600.

### DIFF
--- a/.Lib9c.Tools/SubCommand/Tx.cs
+++ b/.Lib9c.Tools/SubCommand/Tx.cs
@@ -88,6 +88,7 @@ namespace Lib9c.Tools.SubCommand
                         nameof(Nekoyume.Action.MigrationActivatedAccountsState) => new MigrationActivatedAccountsState(),
                         nameof(Nekoyume.Action.MigrationAvatarState) => new MigrationAvatarState(),
                         nameof(Nekoyume.Action.CreatePendingActivations) => new CreatePendingActivations(),
+                        nameof(Nekoyume.Action.RenewAdminState) => new RenewAdminState(),
                         _ => throw new CommandExitedException($"Can't determine given action type: {type}", 128),
                     };
                     action.LoadPlainValue(plainValue);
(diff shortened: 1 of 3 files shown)


longfin:
It would be better we have CLI commands to create and sign this action via 9c-tools.

moreal:
@longfin Thanks for your suggestion.

I changed more after you looked:

- Introduced `9c-tools tx renew-admin-state <NEW-VALID-UNTIL>` command. (ref. https://github.com/planetarium/lib9c/pull/805#issuecomment-1005707992)
- `RenewAdminState` became to renew `AdminState`'s `ValidUntil` field even the `AdminState` is already expired. (ref. https://github.com/planetarium/lib9c/pull/805/commits/4b7cc0aa8dfd58ff3ba12b7ecddfd0acc051be39)

area363:
@moreal Could you redirect this PR to https://github.com/planetarium/NineChronicles.Headless/tree/rc-v100098 so that it can be included in the new rc? 🙏

moreal:
> @moreal Could you redirect this PR to https://github.com/planetarium/NineChronicles.Headless/tree/rc-v100098 so that it can be included in the new rc? 🙏

Thank you for letting me know it. I redirect this PR to the rc-v100098 branch and added this PR to the v100098 project board. 🙏🏻 ✅

longfin (changes requested):
We need to add [a type dispatcher in `Tx.Sign()` method](https://github.com/planetarium/lib9c/blob/1700da1dd178d0079d8d33b2199fc23e8139f7ad/.Lib9c.Tools/SubCommand/Tx.cs#L81-L91) too.

.Lib9c.Tools/SubCommand/Tx.cs
```diff
@@ -250,5 +251,16 @@ public void MigrationActivatedAccountsState()
...
+
+        [Command(Description = "Create RenewAdminState action and dump it.")]
+        public void RenewAdminState(
+            [Argument("NEW-VALID-UNTIL")]
+            long newValidUntil
+        )
+        {
+            PolymorphicAction<ActionBase> action = new RenewAdminState(newValidUntil);
+            byte[] raw = _codec.Encode(action.PlainValue);
+            Console.WriteLine(ByteUtil.Hex(raw));
```
longfin, line 263: We need to specify `nameof(RedeemAdminState)` as action type.

moreal: I had misunderstood their logic 😞 I'll patch it now.

area363 (approved):

.Lib9c.Tools/SubCommand/Tx.cs
moreal: I amended it in f3ecdde. Please review again! @longfin

longfin (approved):

PR #806, opened 2022-01-05
bump libplanet to 0.25.0

area363:
SSIA


area363:
@dahlia Could you take a closer look at https://github.com/planetarium/lib9c/pull/806/files#diff-96a9296132ace389c794e77943b53b5c9e2c04ca83bf1a0f6ac448e8fedccce6 and https://github.com/planetarium/lib9c/pull/806/files#diff-0c8e7c8ba6898e5efe2f083c4252a949adcecc17522ed64c12f7a576453183b1 on this PR? 

It's due to the new implementation added to https://github.com/planetarium/libplanet/blob/main/Libplanet/Store/Trie/MemoryKeyValueStore.cs. 🙇

.Lib9c.Tests/MemoryKeyValueStore.cs
```diff
@@ -1,46 +1,48 @@
 namespace Lib9c.Tests
 {
+    using System.Collections.Concurrent;
     using System.Collections.Generic;
+    using System.Collections.Immutable;
     using Libplanet.Store.Trie;
 
     public class MemoryKeyValueStore : IKeyValueStore
```
dahlia, line 8: It seems equivalent to `Libplanet.Store.Trie.MemoryKeyValueStore`.  Can't we get rid of it and use `Libplanet.Store.Trie.MemoryKeyValueStore` instead?

.Lib9c.Tools/SubCommand/State.cs
```diff
@@ -238,6 +238,50 @@ public void Set(IDictionary<byte[], byte[]> values)
...
+                {
+                    _dictionary.TryRemove(key.ToByteArray(), out _);
+                }
+            }
+
+            public bool Exists(in KeyBytes key) =>
+                _dictionary.ContainsKey(key.ToByteArray());
+
+            IEnumerable<KeyBytes> IKeyValueStore.ListKeys() =>
+                (IEnumerable<KeyBytes>)_dictionary.Keys;
```
dahlia, line 283: Existed methods with the same name still remains above.  These methods implemented the previous version of `IKeyValueStore` interface, but now they are unnecessary as they no more satisfy the current interface.

The changes made on `IKeyValueStore` from Libplanet 0.24 to 0.25 are mainly about key types.  Previously, the interface represented keys as `byte[]`, but now they are represented in `KeyBytes`.

In .NET, arrays do not implement `GetHashCode()` or any methods to determine if two arrays consist of the same elements with the same order.  There's a common mistake to use arrays as dictionary keys, because it causes a dictionary to allow to have duplicate keys.  In order to correctly use arrays as dictionary keys, we should implement an `IEqualityComparer<T[]>` and let dictionaries use it for comparing keys.  That's why `ArrayEqualityComparer<T>` class exists inside this `MemoryKeyValueStore` class.

`KeyBytes` was introduced to prevent from such mistake.  It implements `GetHashCode()` and `IEquatable<KeyBytes>` so that it can be dictionary keys without any burden.  It means we can safely delete `ArrayEqualityComparer<T>` as well if we change the type of `_dictionary` from `ConcurrentDictionary<byte[], byte[]>` to `ConcurrentDictionary<KeyBytes, byte[]>`.

To sum up:

- `_dictionary` now can be `ConcurrentDictionary<KeyBytes, byte[]>` and IMHO it's much better.
- Existed methods are now unnecessary.  Let's get rid of them.
- `ArrayEqualityComparer<T>` class is also unnecessary.  Let's remove it!

.Lib9c.Tools/SubCommand/State.cs
```diff
@@ -206,132 +205,66 @@ public void Dispose()
...
                 long i = 0;
-                foreach (KeyValuePair<byte[], byte[]> pair in _dictionary)
+                foreach (KeyValuePair<KeyBytes, byte[]> pair in _dictionary)
                 {
                     string subDir = Path.Join(
                         _dumpPath,
-                        pair.Key.LongLength.ToString(),
-                        pair.Key.Length > 0 ? $"{pair.Key[0]:x2}" : "_"
+                        pair.Key.Length.ToString(),
+                        pair.Key.Length > 0 ? $"{pair.Key.ToByteArray()[0]:x2}" : "_"
```
dahlia, line 213: `KeyBytes.ToByteArray()` method internally makes a copy of its internal buffer whereas `KeyBytes.ByteArray` property simply returns its internal buffer (and that's why one is a method and other one is a property).

```suggestion
                        pair.Key.Length > 0 ? $"{pair.Key.ByteArray[0]:x2}" : "_"
```

.Lib9c.Tools/SubCommand/State.cs
```diff
@@ -206,132 +205,66 @@ public void Dispose()
...
                     );
                     Directory.CreateDirectory(subDir);
-                    File.WriteAllBytes(Path.Join(subDir, $"{ByteUtil.Hex(pair.Key)}"), pair.Value);
+                    File.WriteAllBytes(
+                        Path.Join(subDir,$"{ByteUtil.Hex(pair.Key.ToByteArray())}"), pair.Value);
```
dahlia, line 217:
```suggestion
                        Path.Join(subDir, pair.key.Hex), pair.Value);
```

area363:
Thanks for the suggestions! Is this ok to merge? @dahlia 🙇

dahlia:
> Thanks for the suggestions! Is this ok to merge? @dahlia 🙇

*.Lib9c.Tests/MemoryKeyValueStore.cs* seems a duplicate of `Libplanet.Store.Trie.MemoryKeyValueStore`.  Couldn't you remove *.Lib9c.Tests/MemoryKeyValueStore.cs* file and use `Libplanet.Store.Trie.MemoryKeyValueStore` instead?

area363:
Ah 

> > Thanks for the suggestions! Is this ok to merge? @dahlia 🙇
> 
> _.Lib9c.Tests/MemoryKeyValueStore.cs_ seems a duplicate of `Libplanet.Store.Trie.MemoryKeyValueStore`. Couldn't you remove _.Lib9c.Tests/MemoryKeyValueStore.cs_ file and use `Libplanet.Store.Trie.MemoryKeyValueStore` instead?

Ah, I see you want me to remove it completely. I'll try that.

area363:
I've removed the file and it works as expected! @dahlia

greymistcube (approved):

moreal (approved):

dahlia (approved):
👍🏼

PR #815, opened 2022-01-14
Fix `RenewAdminState`'s deserialization issue

moreal:
SSIA.


.Lib9c.Tools/SubCommand/Tx.cs
```diff
@@ -88,7 +88,7 @@ public class Tx
                         nameof(Nekoyume.Action.MigrationActivatedAccountsState) => new MigrationActivatedAccountsState(),
                         nameof(Nekoyume.Action.MigrationAvatarState) => new MigrationAvatarState(),
                         nameof(Nekoyume.Action.CreatePendingActivations) => new CreatePendingActivations(),
-                        nameof(Nekoyume.Action.RenewAdminState) => new RenewAdminState(),
+                        nameof(Nekoyume.Action.RenewAdminState0) => new RenewAdminState0(),
```
longfin, line 91: I guess we need to keep a latest action name

Lib9c/Policy/BlockPolicySource.cs
```diff
@@ -58,14 +58,16 @@ public partial class BlockPolicySource
...
+        /// Temporary block index:
         /// </summary>
         public const long V100093ObsoleteIndex = 3_050_000;
 
         public const long V100095ObsoleteIndex = 3_317_632;
-        
+
         public const long V100096ObsoleteIndex = 3_317_632;
 
+        public const long V100099ObsoleteIndex = 3_490_961;
```
longfin, line 69: it's already over.

longfin (changes requested):

Lib9c/Policy/BlockPolicySource.cs
longfin: My fault. :sweat:

.Lib9c.Tools/SubCommand/Tx.cs
moreal: Thank you for letting me know it.

```
$ dotnet run -p .Lib9c.Tools -- tx renew-admin-state 100
6c7531353a52656e657741646d696e53746174656475323a696431363a424a1d1fc238984797fc51e8b0b2c6817531353a6e65775f76616c69645f756e74696c69313030656565
```

moreal: I amended it in ad6b757 commit.

longfin (approved):

moreal:
Since it was a problem that occurred while deserializing, we believe it is not able to be included in the chain. So it doesn't need action version upgrade it will be reverted.

moreal:
> Since it was a problem that occurred while deserializing, we believe it is not able to be included in the chain. So it doesn't need action version upgrade it will be reverted.

About the above, I reverted all changes and just fixed the deserialization issue with the test case. Could you review it again?

longfin (approved):

moreal:
Since it was approved, I'll merge this pull request when the check passed.

PR #818, opened 2022-01-14
Prevent double activation attempt

greymistcube:
Attempts to prevent invalid `Transaction<T>`s clogging the network making the game unplayable for those who try to activate an account more than once.

@longfin should take a look. 🙄

### DIFF
--- a/Lib9c/Policy/BlockPolicySource.cs
+++ b/Lib9c/Policy/BlockPolicySource.cs
@@ -58,12 +58,12 @@ namespace Nekoyume.BlockChain.Policy
         /// Current date: 2021/12/14 16:21
         /// Current block index: 2_965_521
         /// Temporary target date : 2021/12/20 12:00
-        /// Temporary block index: 
+        /// Temporary block index:
         /// </summary>
         public const long V100093ObsoleteIndex = 3_050_000;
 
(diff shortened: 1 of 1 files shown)


Lib9c/Policy/BlockPolicySource.cs
```diff
@@ -294,7 +294,8 @@ public partial class BlockPolicySource
                     transaction.Actions.First().InnerAction is IActivateAction aa)
                 {
                     return blockChain.GetState(aa.GetPendingAddress()) is Dictionary rawPending &&
-                        new PendingActivationState(rawPending).Verify(aa.GetSignature())
+                        new PendingActivationState(rawPending).Verify(aa.GetSignature()) &&
+                        transaction.Nonce == 0
```
longfin, line 298: Tiny optimization but we can skip unnecessary `GetState()` & `PendingActivationState.Verify()` calls if we move nonce checking to the first.

longfin (approved):

Lib9c/Policy/BlockPolicySource.cs
greymistcube: Fixed.

longfin (approved):

PR #832, opened 2022-01-27
Optimize miner to append block

moreal:
Originally, `Miner` mined block and append it in two steps. But `BlockChain<T>.MineBlock()` supports to append block in once to avoid evaluating actions again.


longfin (approved):

moreal:
This pull request should target `rc-v100102` branch after the branch created.

dahlia (approved):

PR #866, opened 2022-03-02
Change RankingSimulator Constructor Parameters

hskim881028:
- In the `RankingSimulator` constructor, inject `Player` instead of `AvatarState`

- Version up list
   - ranking_battle10 -> ranking_battle11
   - RankingSimulator ->RankingSimulatorV1
   - New RankingSimulator
   
- Related Link
   - https://github.com/planetarium/NineChronicles/pull/1220
   -  [RankingBattle 등에서 AvatarState 를 풀로 쓰는 대신 전투에 필요한 중간 상태만 사용하게끔](https://app.asana.com/0/958521740385861/1201869821990665/f)

### DIFF
--- a/Lib9c/Battle/SimulationEnemyPlayer.cs
+++ b/Lib9c/Battle/SimulationEnemyPlayer.cs
@@ -1,0 +1,33 @@
+﻿using System.Collections.Generic;
+using Nekoyume.Model;
+using Nekoyume.Model.Item;
+
+namespace Nekoyume.Battle
+{
+    public readonly struct SimulationEnemyPlayer
+    {
+        public readonly string NameWithHash;
+        public readonly int CharacterId;
+        public readonly int Level;
+        public readonly int HairIndex;
+        public readonly int LensIndex;
+        public readonly int EarIndex;
+        public readonly int TailIndex;
+
+        public readonly IReadOnlyList<Costume> Costumes;
+        public readonly IReadOnlyList<Equipment> Equipments;
+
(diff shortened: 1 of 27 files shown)


ipdae (changes requested):
Since the  Player also contains all the information  from AvatarState, it seems like a change away from the original goal of reducing the weight of the model. I think it would be better to create a new model.

boscohyun (approved):

boscohyun:
CI was broken.😥

```
Error: Action/RankingBattle10.cs(78,25): error CS7036: There is no argument given that corresponds to the required formal parameter 'migrationRequired' of 'AccountStateDeltaExtensions.TryGetAvatarStateV2(IAccountStateDelta, Address, Address, out AvatarState, out bool)' [/home/runner/work/lib9c/lib9c/Lib9c/Lib9c.csproj]
```

Lib9c/Battle/SimulationEnemyPlayer.cs
```diff
@@ -0,0 +1,33 @@
+﻿using System.Collections.Generic;
+using Nekoyume.Model;
+using Nekoyume.Model.Item;
+
+namespace Nekoyume.Battle
+{
+    public readonly struct SimulationEnemyPlayer
```
ipdae, line 7: i think we need add `MessagePackObject` for grpc serialize.

boscohyun (approved):

sonohoshi (approved):

longfin:
Do we need to bump RankingBattle action? I guess there is no changes about serialization format and execution result.

PR #871, opened 2022-03-03
Improve `9c-tools state` subcommands

dahlia:
- `9c-tools state rebuild` became possible to start from other than the genesis (unless its previous block has no states in the store).  It can be specified through `-f`/`-B`/`--bottommost` option.
- `9c-tools state` subcommands became to immediately stop if it receives `SIGINT`, i.e., <kbd>^C</kbd>.
- Added `9c-tools state check` subcommand which determines if the specified block has states in the store, and find the nearest ancestor and descendant blocks if it does not have states.
- Fixed *Lib9c.DevExtensions.Tests* which had not actually worked.


riemannulus (approved):
🚀

PR #873, opened 2022-03-04
Feature/use get states in hack and slash

boscohyun:
- This pull request is not hard fork
- Introduce "FailedLoadSheetException"
- Introduce "GetStatesAsDict()" and "GetSheets" in "AccountStateDeltaExtensions"
- Introduce "SheetsExtensions"

### DIFF
--- a/Lib9c/Action/AccountStateDeltaExtensions.cs
+++ b/Lib9c/Action/AccountStateDeltaExtensions.cs
@@ -65,6 +65,20 @@ namespace Nekoyume.Action
             return false;
         }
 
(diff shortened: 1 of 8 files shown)


Unengine (approved):

Lib9c/Action/AccountStateDeltaExtensions.cs
```diff
@@ -485,6 +502,130 @@ public static T GetSheet<T>(this IAccountStateDelta states) where T : ISheet, ne
             }
         }
 
+        public static Dictionary<Type, (Address address, ISheet sheet)> GetSheets(
```
hskim881028, line 505: I think it could be changed for the better... I approved 😎

hskim881028 (approved):

PR #875, opened 2022-03-08
Introduce EnemyPlayerDigest

ipdae:
- based on #866
- revert version up action

### DIFF
--- a/Lib9c/Action/RankingBattle.cs
+++ b/Lib9c/Action/RankingBattle.cs
@@ -29,7 +29,7 @@ namespace Nekoyume.Action
         public Address weeklyArenaAddress;
         public List<Guid> costumeIds;
         public List<Guid> equipmentIds;
-        public SimulationEnemyPlayer EnemyPlayer;
+        public EnemyPlayerDigest EnemyPlayerDigest;
         public ArenaInfo ArenaInfo;
         public ArenaInfo EnemyArenaInfo;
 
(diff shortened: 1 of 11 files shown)


Lib9c/Model/Character/Player.cs
```diff
@@ -100,6 +100,34 @@ public Player(AvatarState avatarState, Simulator simulator)
             PostConstruction(simulator.CharacterLevelSheet, simulator.EquipmentItemSetEffectSheet);
         }
 
+        protected Player(SimulationEnemyPlayer enemyPlayer,
```
longfin, line 103: It would be better moving this code to EnemyPlayer to clarify responsibility.

Lib9c/Battle/SimulationEnemyPlayer.cs
```diff
@@ -0,0 +1,65 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using Bencodex.Types;
+using Nekoyume.Model;
+using Nekoyume.Model.Item;
+using Nekoyume.Model.State;
+
+namespace Nekoyume.Battle
+{
+    public readonly struct SimulationEnemyPlayer: IState
```
longfin, line 10: I guess `EnemyPlayerDigest` is better name for it.

Lib9c/Action/RankingBattle.cs
```diff
@@ -192,12 +193,15 @@ public override IAccountStateDelta Execute(IActionContext context)
...
             EnemyArenaInfo = new ArenaInfo((Dictionary)weeklyArenaMap[enemyKey]);
+            var rankingSheets = states.GetRankingSimulatorSheets();
+            var player = new Player(avatarState, rankingSheets);
+            var enemyPlayer = new EnemyPlayer(enemyAvatarState, rankingSheets);
             var simulator = new RankingSimulator(
                 ctx.Random,
-                avatarState,
-                enemyAvatarState,
+                player,
+                enemyPlayer,
```
longfin, line 202: I guess it would be better it takes `SimualationEnemyPlayer` than `EnemyPlayer` because `SimualationEnemyPlayer` is smaller than `EnemyPlayer`.

longfin (approved):

PR #897, opened 2022-03-23
Divide each ArenaInfo for reduce RankingBattle execution time.

ipdae:
- Divide `WeeklyArenaState.Map` to `List<Address>, ArenaInfo` like `Order, OrderDigest`.
- Using new model greater than `RankigBattle.UpdateTargetIndex`
- 
---

This is local running test result with same avatar, opponent.
 
AS-IS
```
[15:08:43 DBG] 1 actions in transaction 5d2cfd7111096192fb2852ae4542a760bef0b849866a941ef4e337c0ac3e0bb1 by 0x8fF5e1c64860aF7d88b019837a378fBbec75c7D9 with timestamp 2022-03-23T06:07:28.828310Z evaluated in 1555ms.
```

TO-Be
```
[15:08:44 DBG] 1 actions in transaction 1531397e8885c533533db0822344c5ba62f9098065ff6b3c53dabb900323de4f by 0x8fF5e1c64860aF7d88b019837a378fBbec75c7D9 with timestamp 2022-03-23T06:07:28.948427Z evaluated in 145ms.
```


area363 (approved):

boscohyun (approved):

Lib9c/Action/RankingBattle.cs
```diff
@@ -146,17 +149,144 @@ public override IAccountStateDelta Execute(IActionContext context)
...
 
             bool arenaEnded = rawWeeklyArenaState["ended"].ToBoolean();
-            Dictionary weeklyArenaMap = (Dictionary) rawWeeklyArenaState["map"];
             if (arenaEnded)
             {
                 throw new WeeklyArenaStateAlreadyEndedException();
             }
 
-            var costumeStatSheet = sheets.GetSheet<CostumeStatSheet>();
+            if (context.BlockIndex >= UpdateTargetBlockIndex)
```
longfin, line 157: Extracting this section to private method is more natural for me.

Lib9c/Action/RankingBattle.cs
```diff
@@ -18,11 +18,13 @@
 namespace Nekoyume.Action
 {
     [Serializable]
-    [ActionType("ranking_battle10")]
+    [ActionType("ranking_battle11")]
     public class RankingBattle : GameAction
     {
         public const int StageId = 999999;
         public static readonly BigInteger EntranceFee = 100;
+        public const long UpdateTargetBlockIndex = 3_752_000L;
```
longfin, line 26: I think `Update` is too general term. How about using the name suggesting that each arena is segmented? (e.g. `SegmentedArenaTargetBlockIndex`)

Or, In cases where proposals are separated by a name like EIP, [it is sometimes used as it is in the variable name](https://github.com/ethereum/go-ethereum/blob/8bbf83e7a4e680c3b3b8e7a1cb09b76479dd26fe/core/tx_pool.go#L591-L592).

Lib9c/Action/RankingBattle.cs
```diff
@@ -18,11 +18,13 @@
...
+        public const int UpdateTargetIndex = 67;
```
longfin, line 27: `Index` postfix seems ambigious too.

Lib9c/Action/RankingBattle.cs
ipdae: Is it better to use the lib9c issue number(e.g. `issue-883`)?

longfin: GH issue number can be better but I think it isn't enough for naming... 

If we don't have a suitable name, it might be ok to keep the issue but leave a comment.

ipdae: ok, i'll leave a comment this.

Lib9c/Action/RankingBattle.cs
```diff
@@ -146,17 +152,152 @@ public override IAccountStateDelta Execute(IActionContext context)
...
+                var ended = DateTimeOffset.UtcNow;
+                Log.Verbose("{AddressesHex}RankingBattle Total Executed Time: {Elapsed}", addressesHex, ended - started);
+                EnemyPlayerDigest = enemyPlayerDigest;
+                return states;
+            }
+            // Run Backward compatible
+            return BackwardCompatibleExecute(rawWeeklyArenaState, sheets, avatarState, costumeStatSheet, sw, addressesHex, enemyAvatarState, ctx, states, inventoryAddress, questListAddress, migrationRequired, worldInformationAddress, started);
+        }
+
+        private IAccountStateDelta BackwardCompatibleExecute(Dictionary rawWeeklyArenaState, Dictionary<Type, (Address address, ISheet sheet)> sheets,
```
longfin, line 295:
```suggestion
        private IAccountStateDelta ExecuteAsRankingBattle10(Dictionary rawWeeklyArenaState, Dictionary<Type, (Address address, ISheet sheet)> sheets,
```

Lib9c/Action/RankingBattle.cs
```diff
@@ -146,17 +152,152 @@ public override IAccountStateDelta Execute(IActionContext context)
...
+                Log.Verbose("{AddressesHex}RankingBattle Serialize AvatarState: {Elapsed}", addressesHex, sw.Elapsed);
+                sw.Restart();
+
+                var ended = DateTimeOffset.UtcNow;
+                Log.Verbose("{AddressesHex}RankingBattle Total Executed Time: {Elapsed}", addressesHex, ended - started);
+                EnemyPlayerDigest = enemyPlayerDigest;
+                return states;
+            }
+            // Run Backward compatible
+            return BackwardCompatibleExecute(rawWeeklyArenaState, sheets, avatarState, costumeStatSheet, sw, addressesHex, enemyAvatarState, ctx, states, inventoryAddress, questListAddress, migrationRequired, worldInformationAddress, started);
```
longfin, line 292: How about inverting [if condition](https://github.com/planetarium/lib9c/pull/897/files#diff-12fde98e1392700396baaa420aa9fa7f535f5d36ddc14fa61d8305ef5486c1b4R160) and returning early?

longfin (approved):

PR #940, opened 2022-04-19
Fix arena score formula

boscohyun:
- Archive `RankingBattle11`
- Upgrade `RankingBattle`
- Upgrade `RankingSimulator.Simulate()`
- Upgrade `ArenaInfo.Update()`
- Upgrade `ArenaScoreHelper.GetScore()` and `ArenaScoreHelper.CachedScore`
- [[season3] 스코어 처리 방식 변경](https://app.asana.com/0/958521740385861/1202139879327858/f)


ipdae (commented):
in this pr, only the score calculation method is changed without other simulate logic, but it seems that too many models are being updated for this. IMO, it will be better calculate score outside of Simulator.Simulate can be reduce code changes.

boscohyun:
Close this pull request and instead with https://github.com/planetarium/lib9c/pull/941

PR #947, opened 2022-04-22
Check previous recipe id by ItemSubType

ipdae:
- Check recipe by ItemSubType(Weapon, Armor, etc.)

### DIFF
--- a/Lib9c/Action/UnlockEquipmentRecipe.cs
+++ b/Lib9c/Action/UnlockEquipmentRecipe.cs
@@ -6,6 +6,7 @@ using Bencodex.Types;
 using Libplanet;
 using Libplanet.Action;
 using Libplanet.Assets;
+using Nekoyume.Extensions;
 using Nekoyume.Helper;
 using Nekoyume.Model;
 using Nekoyume.Model.State;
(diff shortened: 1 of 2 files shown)


Lib9c/Action/UnlockEquipmentRecipe.cs
```diff
@@ -83,20 +90,34 @@ public override IAccountStateDelta Execute(IActionContext context)
...
+                if (equipmentRow.Id != firstId)
                 {
-                    throw new NotEnoughClearedStageLevelException($"clear {row.UnlockStage} first.");
+                    var prevId = recipeId - 1;
+                    if (!unlockedIds.Contains(prevId))
+                    {
+                        // Can't skip previous recipe unlock.
+                        throw new InvalidRecipeIdException($"unlock {prevId} first.");
+                    }
                 }
```
Unengine, line 116: I think previous recipe is checked via `Row.UnlockStage`, not with `Row.Id`.

ipdae: It is correct to check recipeId because it contains from list of unlocked recipe Ids. we check the stageId at 95line.

Unengine (approved):

PR #948, opened 2022-04-22
Change grinding formula and push meta files of TableSheets.

sonohoshi:
per ten thousand to per cent.


ipdae (changes requested):
we just change the data and calculation formulas are scaled differently, but the execution results shouldn't be different from before.

Lib9c/Action/Grinding.cs
```diff
@@ -166,7 +166,7 @@ CrystalMonsterCollectionMultiplierSheet crystalMonsterCollectionMultiplierSheet
             }
             CrystalMonsterCollectionMultiplierSheet.Row multiplierRow =
                 crystalMonsterCollectionMultiplierSheet[monsterCollectionLevel];
-            var extra = crystal.DivRem(100, out _) * (multiplierRow.Multiplier / 100);
+            var extra = crystal.DivRem(10, out _) * (multiplierRow.Multiplier / 10);
```
ipdae, line 169:
```suggestion
            var extra = crystal.DivRem(100, out _) * multiplierRow.Multiplier;
```

ipdae (approved):

PR #952, opened 2022-04-25
Introduce `Stake` action

moreal:
Originally, this pull request introduced both `Stake` and `ClaimStakeReward` actions for Monster Collection Improvement TF. But I'll open one more pull request for the `ClaimStakeReward` action by other branch forked from this branch because this pull requests changes' size is too big now.


Lib9c/Action/Stake.cs
```diff
@@ -0,0 +1,112 @@
...
+                    context.Signer.ToHex(),
+                    Amount,
+                    currentBalance.RawValue);
+            }
+
+            // Stake if it doesn't exist yet.
+            if (!states.TryGetStakeState(context.Signer, out StakeState stakeState))
+            {
+                var sheet = states.GetSheet<StakeAchievementRewardSheet>();
+                var stakeSheet = new StakeState(stakeStateAddress, context.BlockIndex);
```
ipdae, line 65:
```suggestion
                var stakeState = new StakeState(stakeStateAddress, context.BlockIndex);
```

Lib9c/Model/State/StakeState.cs
```diff
@@ -0,0 +1,98 @@
...
+
+            public StakeAchievements(Dictionary serialized)
+            {
+                _achievements = serialized.ToDictionary(
+                    pair => int.Parse(((Text) pair.Key).Value, CultureInfo.InvariantCulture),
+                    pair => (int) (Integer) pair.Value);
+            }
+
+            public IValue Serialize() =>
+                new Dictionary(_achievements.OrderBy(pair => pair.Key).Select(pair =>
```
ipdae, line 32: IMO, it would be better to serialize it as a List rather than a Dictionary.

Lib9c/Action/Stake.cs
```diff
@@ -0,0 +1,112 @@
...
+        private BigInteger Amount { get; set; }
+
+        public Stake(BigInteger amount)
+        {
+            Amount = amount >= 0
+                ? amount
+                : throw new ArgumentOutOfRangeException(nameof(amount));
+        }
+
+        public override IValue PlainValue { get; }
```
ipdae, line 23: I think it will be raise exception when serialize `Stake`.

Lib9c/Model/State/StakeState.cs
moreal: Currently, It seems like `Dictionary[Level => Step]` so you think like `List[Tuple[Level => Step]]`?
As a comment, the reason to call `OrderBy` explicitly there though Bencodex `Dictionary` guarantees its order, it is because Libplanet Analyzer warned it though it isn't misusage.

image

moreal: I rewrite them like the below lines. If you think like that because it has order, how about them?

```csharp
            public IValue Serialize() =>
                new Dictionary(_achievements.ToDictionary(
                    pair => (IKey)(Text) pair.Key.ToString(CultureInfo.InvariantCulture),
                    pair => (IValue)(Integer) pair.Value));
```

moreal: At this time, I'll merge this pull request 😞 Thanks for your opinions and reviews!

PR #954, opened 2022-04-25
Filter inactive ArenaInfo when prepare next arena

### LINKED ISSUE #946: Filter deactivated ArenaInfo
I found this problem while working on https://github.com/planetarium/NineChronicles/pull/1379.
```
[ArenaInfoList] Total:9664, Activated: 3325, Deactivated: 6339
```
2/3 of the current ArenaInfo is unused but copy every week.
we can filtering the list in [RewardGold](https://github.com/planetarium/lib9c/blob/6c2d89d9c240ea836e8f7c4a5165d829ff5c9e70/Lib9c/Action/RewardGold.cs#L172-L185) and it might be reduces the size of states to be fetched from clients and RankingBattle execution.

ipdae:
- resolve #946 
- Set blockIndex at `3_976_000L`(after 2weeks)


Lib9c/Action/RewardGold.cs
```diff
@@ -169,18 +173,28 @@ public IAccountStateDelta PrepareNextArena(IActionContext ctx, IAccountStateDelt
...
                                         out Dictionary rawInfo))
                                 {
                                     var prevInfo = new ArenaInfo(rawInfo);
-                                    var info = new ArenaInfo(prevInfo);
+                                    var record = prevInfo.ArenaRecord;
+                                    // Filter ArenaInfo
+                                    if (filterInactive && record.Win == 0 && record.Draw == 0 &&
+                                        record.Lose == 0)
+                                    {
+                                        addressList.Remove(address);
```
longfin, line 192: Instead of maintaining `addressList` itself, how about using new aggregated value for `listAddress` state?

Lib9c/Action/RewardGold.cs
```diff
@@ -135,6 +137,7 @@ public IAccountStateDelta PrepareNextArena(IActionContext ctx, IAccountStateDelt
                     var addressList = states.TryGetState(listAddress, out List rawList)
                         ? rawList.ToList(StateExtensions.ToAddress)
                         : new List<Address>();
+                    var serializedList = rawList ?? List.Empty;
```
longfin, line 140: `serializedList` is too general and ambigious naming. how about `nextAddresses` ?

longfin (approved):

longfin (approved):

PR #1018, opened 2022-05-17
Introduce `CrystalCalculator.CalculateMaterialCost`

Unengine:
- `CalculateMaterialCost` can calculate cost crystal of single/multiple materials.

### DIFF
--- a/Lib9c/Helper/CrystalCalculator.cs
+++ b/Lib9c/Helper/CrystalCalculator.cs
@@ -58,5 +58,32 @@ namespace Nekoyume.Helper
             var extra = crystal.DivRem(100, out _) * multiplierRow.Multiplier;
             return crystal + extra;
         }
+
+        /// <param name="materials"> Key : id of material, Value : count of material </param>
+        public static FungibleAssetValue CalculateMaterialCost(
+            Dictionary<int, int> materials,
+            CrystalMaterialCostSheet crystalMaterialCostsheet)
+        {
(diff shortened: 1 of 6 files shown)


ipdae:
build failed.

Unengine:
> build failed.
Fixed it with applying `OrderBy`.
```csharp
foreach (var material in materials.OrderBy(pair => pair.Key))
```

Lib9c/Helper/CrystalCalculator.cs
```diff
@@ -58,5 +58,32 @@ bool enhancementFailed
             var extra = crystal.DivRem(100, out _) * multiplierRow.Multiplier;
             return crystal + extra;
         }
+
+        /// <param name="materials"> Key : id of material, Value : count of material </param>
+        public static FungibleAssetValue CalculateMaterialCost(
```
ipdae, line 63: Is this used method?

ipdae (commented):
It would be good to add unit tests.

Lib9c/Helper/CrystalCalculator.cs
Unengine: It is used in 9c client. Is it not okay to keep unused method in lib9c?

ipdae: If it's only used on client, it might be better to move it to the client side.

ipdae (approved):

PR #1025, opened 2022-05-20
Fix first stage sweep bug

Atralupus:
Related to https://github.com/planetarium/lib9c/pull/1017

Add first stage sweep test case & Fix stage check rule

----
And... I'm add below code
```csharp
...
# check first world, first stage
var firstStage = prevStageId == 1;
...
if (!prevStageWorld.IsStageCleared && !firstStage)
...
```

I don't think it's best to do hardcoding like this
Is there any way to get information about the first stage of the first world?


boscohyun (approved):

tyrosine1153 (approved):
image
It works well on 9c!

PR #1028, opened 2022-05-20
Create preview net admin account action

Atralupus:
Create admin account Action for faucet
It's send 100,000,000ncg to admin account for support 10,000 users

Plan to create an admin account.
1. Create and register `BlockPolicySource.PreviewNetAdmin` private key in preview net.
2. Run this action within executable index. <- action executable index is `823,000` (now preview net's index is `793,888`)
3. After the executable index, the action becomes disable
   and the mainnet has a higher index, so there should be no problem it on the main net.
4. use admin account!

### DIFF
--- a/Lib9c/Action/CreatePreviewNetAdminAccount.cs
+++ b/Lib9c/Action/CreatePreviewNetAdminAccount.cs
@@ -1,0 +1,54 @@
+using System;
+using System.Collections.Generic;
+using Bencodex.Types;
+using Libplanet;
+using Libplanet.Action;
+using Libplanet.Assets;
+using Nekoyume.Model.State;
+
+namespace Nekoyume.Action
+{
+    /// <summary>
+    /// </summary>
+    [Serializable]
+    [ActionObsolete(BlockChain.Policy.BlockPolicySource.PreviewNetAllowIndex)]
+    [ActionType("create_preview_net_admin_account")]
+    public class CreatePreviewNetAdminAccount : ActionBase
+    {
+        public override IValue PlainValue =>
+            new Bencodex.Types.Dictionary(new Dictionary<IKey, IValue>
+            {
+            });
+
+        public override void LoadPlainValue(IValue plainValue)
+        {
+        }
+
+        public CreatePreviewNetAdminAccount()
+        {
+        }
+
+        public override IAccountStateDelta Execute(IActionContext context)
+        {
+            var states = context.PreviousStates;
+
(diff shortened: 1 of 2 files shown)


Atralupus:
test... ill fix it

Lib9c/Policy/BlockPolicySource.cs
```diff
@@ -44,6 +44,8 @@ public partial class BlockPolicySource
 
         public const int MaxTransactionsPerBlock = 100;
 
+        public const long PreviewNetAdminAccountAllowIndex = 823_000;
```
area363, line 47: Since the `previewnet` chain index is greater than this value, does this mean that the `CreatePreviewNetAdminAccount` action will be obsolete because the admin account is already created?

boscohyun (approved):

Lib9c/Action/CreatePreviewNetAdminAccount.cs
```diff
@@ -0,0 +1,53 @@
...
+
+        public CreatePreviewNetAdminAccount()
+        {
+        }
+
+        public override IAccountStateDelta Execute(IActionContext context)
+        {
+            var states = context.PreviousStates;
+
+            CheckObsolete(BlockChain.Policy.BlockPolicySource.PreviewNetAdminAccountAllowIndex, context);
```
ipdae, line 36: I think we must check signer is admin like [this](https://github.com/planetarium/lib9c/blob/v100200/Lib9c/Action/PatchTableSheet.cs)

Lib9c/Policy/BlockPolicySource.cs
```diff
@@ -81,6 +83,10 @@ public partial class BlockPolicySource
             new Address("636d187B4d434244A92B65B06B5e7da14b3810A9"),
         }.ToImmutableHashSet();
 
+        public static readonly Address PreviewNetAdmin = new Address(
```
ipdae, line 86: Where is this field used? If it is to check the address only in the action, AdminState already exists.

ipdae (changes requested):
I think we must check Tx Signer is admin.

Lib9c/Policy/BlockPolicySource.cs
Atralupus: yes... It's already been a while, need to modify the index

Atralupus:
https://github.com/planetarium/lib9c/pull/1028#discussion_r879075531
Already exists admin account... close this pr

PR #1006, opened 2022-05-12
(WIP) Introduce new arena system

hskim881028:
- [아레나 개선 기획 문서](https://www.notion.so/planetarium/0509-39d32d8767b246418f638367f19b9cd9)
---
- Create `JoinArena` action 
- Create `BattleArena` action (work in progress)
   - This is the action that replaces `RankingBattle`
- Create `ArenaAvatar` state
   - This state has the avatar information you need in the arena. (equipment, costumes, avatar level)
- Create `ArenaParticipants` 
   - ArenaParticipants has a list of participants for the arena
- Create `ArenaScore`
    -  ArenaScore only has a score.
- Create `ArenaInformation` 
   - ArenaInformation has win, lose, ticket
 - Create  `Arena` Sheet
   - This sheet contains arena related information
  ---
- Testbed 9c branch : [feature/new-arena-testbed](https://github.com/planetarium/NineChronicles/tree/feature/new-arena-testbed)
   - you can test `JoinArena` Action 
   - you can see `ArenaState` and `ArenaAvatarState`
   
image

PR #1029, opened 2022-05-20
Introduce BattleArena action

hskim881028:
Only parts related to the Join Arena action were taken from https://github.com/planetarium/lib9c/pull/1006

---

- Create `BattleArena` action
   - This is the action that replaces `RankingBattle`

### DIFF
--- a/Lib9c/Action/BattleArena.cs
+++ b/Lib9c/Action/BattleArena.cs
@@ -45,27 +45,27 @@ namespace Nekoyume.Action
         protected override IImmutableDictionary<string, IValue> PlainValueInternal =>
             new Dictionary<string, IValue>()
             {
-                ["myAvatarAddress"] = myAvatarAddress.Serialize(),
-                ["enemyAvatarAddress"] = enemyAvatarAddress.Serialize(),
-                ["championshipId"] = championshipId.Serialize(),
-                ["round"] = round.Serialize(),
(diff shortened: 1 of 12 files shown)


Lib9c/Action/BattleArena.cs
```diff
@@ -0,0 +1,333 @@
...
+            var arenaParticipantsAdr = ArenaParticipants.DeriveAddress(roundData.Id, roundData.Round);
+            if (!states.TryGetArenaParticipants(arenaParticipantsAdr, out var arenaParticipants))
+            {
+                throw new ArenaParticipantsNotFoundException(
+                    $"[{nameof(BattleArena)}] ChampionshipId({roundData.Id}) - round({roundData.Round})");
+            }
+
+            if (!arenaParticipants.AvatarAddresses.Contains(myAvatarAddress))
+            {
+                throw new AddressNotFoundInArenaParticipantsException(
```
ipdae, line 158: When my avatar information is not contain after the start of the round, if we can join process inside `BattleArena`, client does not have to wait `JoinArena`  render, and network can reduce the Tx to 1 from 2.

hskim881028: We can only execute `BatteArena` action by those who participate in The Arena. But we'll discuss it for performance. 🏃🏻‍♂️

ipdae (approved):

PR #1041, opened 2022-05-24
Fix FaucetAsset LoadPlainValue

ipdae:
- Avoid `MissingMethodException`


Lib9c/Action/FaucetAsset.cs
```diff
@@ -17,6 +17,9 @@ public class FaucetAsset : ActionBase
         public Address Recipient;
         public FungibleAssetValue Amount;
 
+        public FaucetAsset()
+        {
+        }
```
moreal, line 22:
```suggestion
        }

```

moreal (approved):

PR #1042, opened 2022-05-24
Update RedeemRewardSheet for Arena season3

longfin:
This PR adds rewards for [Arena Season 3](https://ninechronicles.medium.com/%EF%B8%8F-arena-season-3-252k-ncg-reward-pool-exclusive-costumes-staking-rewards-and-more-%EF%B8%8F-67d5ab75a547) to `RedeemRewardSheet`.

Note: we are already on the mainnet chain, so this sheet is just reference. (we can issue redeem codes before release with it).


longfin:
I amended previous sheet due to wrong ID. cc @Namyujeong @jaeho0103

PR #1046, opened 2022-05-25
MonsterCollection RegularRewardSheet updated

jaeho0103:
MonsterCollectionRegularRewardSheet
for previewnet 0525 data added


area363:
Tests are failing https://github.com/planetarium/lib9c/runs/6585510204?check_suite_focus=true#step:7:1300 🤔

moreal:
The test failure occurred because the tests are hard-coded according to TC.

So, I think it should use mocked table sheets or update test cases.. I'll look more.

moreal:
I pushed a commit to mock `StakeRegularRewardSheet` to fix the assumption of the test cases. [9a297ea](https://github.com/planetarium/lib9c/pull/1046/commits/9a297ea95ddb2f76a346717fadc58ee0cce70510). If it's a bad way, it should be reverted.

area363 (approved):

PR #1047, opened 2022-05-25
Crystal Data Added

crkim88:
Crystal data added


jaeho0103 (changes requested):
All Mimir material data in the Crystal Material Cost Sheet should be deleted.
Also, the data on the id 4 side of the World Unlock Sheet.csv is strange. Please modify it according to the rules.

jaeho0103 (approved):

PR #1066, opened 2022-05-26
Improve ArenaSheet

boscohyun:
### changes

- Rename ArenaSheet.Row.Id to ChampionshipId
- Add ArenaSheet.GetRowByBlockIndex()
- Add ArenaSheet.GetRoundByBlockIndex()
- Add ArenaSheet.Row.ChampionshipNumber
- Add ArenaSheet.RoundData.ChampionshipNumber

### DIFF
--- a/Lib9c/TableData/ArenaSheet.cs
+++ b/Lib9c/TableData/ArenaSheet.cs
@@ -109,6 +109,15 @@ namespace Nekoyume.TableData
             row.Round.Add(value.Round[0]);
         }
 
+        public bool TryGetRowByBlockIndex(long blockIndex, out Row row)
+        {
+            row = OrderedList
+                .FirstOrDefault(e => e.Round.Any(roundData =>
+                    roundData.StartBlockIndex <= blockIndex &&
+                    roundData.EndBlockIndex >= blockIndex));
+            return row != null;
+        }
+
         public bool TryGetRoundByBlockIndex(long blockIndex, out RoundData roundData)
         {
             roundData = OrderedList
--- a/Lib9c/TableData/ArenaSheet.cs
+++ b/Lib9c/TableData/ArenaSheet.cs
@@ -87,6 +87,25 @@ namespace Nekoyume.TableData
                 roundData = Round.FirstOrDefault(x => x.ArenaType.Equals(ArenaType.Championship));
                 return !(roundData is null);
             }
+
+            public bool TryGetSeasonNumber(int round, out int seasonNumber)
+            {
+                seasonNumber = 0;
+                foreach (var roundData in Round)
+                {
+                    if (roundData.ArenaType == ArenaType.Season)
+                    {
+                        seasonNumber++;
+                    }
+
+                    if (roundData.Round == round)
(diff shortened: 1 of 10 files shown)


area363 (approved):

ipdae (commented):
I don't know the exact situation because the methods where they are used, but they seem to be essential information. In this case, i think it would be better to just throw an exception.

Lib9c/TableData/ArenaSheet.cs
```diff
@@ -86,6 +106,25 @@ public bool TryGetChampionshipRound(out RoundData roundData)
                 roundData = Round.FirstOrDefault(x => x.ArenaType.Equals(ArenaType.Championship));
                 return !(roundData is null);
             }
+
+            public bool TryGetSeasonNumber(int round, out int seasonNumber)
```
hskim881028, line 110: `TryGetSeasonNumber()` is only used for 9c. I think it is better to move to 9c.

Lib9c/TableData/ArenaSheet.cs
```diff
@@ -52,9 +62,19 @@ public bool IsTheRoundOpened(long blockIndex)
...
+            
+            /// <summary>
+            /// The [`Round`] contains 8 elements always
+            /// </summary>
             public List<RoundData> Round { get; private set; }
 
+            /// <summary>
+            /// Championship number equals with Id always
+            /// </summary>
+            public int ChampionshipNumber => Id;
```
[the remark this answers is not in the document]
boscohyun: Because they share the same value, this may be redundant information.

I think using ArenaSheet.Row.Id as return value for ChampionshipNumber has the same function as we use ItemUsable.ItemId as return value for INonFungibleItem.NonFungibleId or ITradableItem.TradableId.

Even if the summary of ArenaSheet.Row.Id contains enough information, I don't think the user will want to find the Championship number by the name of Id. Just like when we want to get the Item's NonFungibleId, we don't reference it as ItemUsable.ItemId.

Lib9c/TableData/ArenaSheet.cs
boscohyun: It's still correct to actually use it only on the frontend.

However, I believe that these kinds of implicit promises should be protocolized, and validating them with unit tests helps the code to be sustainable.

I have added unit tests. What do you think of this part?

boscohyun:
> I don't know the exact situation because the methods where they are used, but they seem to be essential information. In this case, i think it would be better to just throw an exception.

@ipdae Can you explain a bit more about where you want the exception to be raised?

boscohyun:
@ipdae @hskim881028 I applied some suggestions.

hskim881028 (approved):

Lib9c/TableData/ArenaSheet.cs
ipdae: i agree with @hskim881028

boscohyun: @hskim881028 @ipdae I removed this function.

PR #1075, opened 2022-05-30
Update `HackAndSlashBuffState` in HAS action

sonohoshi:
SSIA

### DIFF
--- a/Lib9c/Action/HackAndSlash.cs
+++ b/Lib9c/Action/HackAndSlash.cs
@@ -10,6 +10,7 @@ using Nekoyume.Battle;
 using Nekoyume.Extensions;
 using Nekoyume.Model.State;
 using Nekoyume.TableData;
+using Nekoyume.TableData.Crystal;
 using Serilog;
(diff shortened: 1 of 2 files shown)


Lib9c/Action/HackAndSlash.cs
```diff
@@ -28,6 +29,7 @@ public class HackAndSlash : GameAction
         public List<Guid> foods;
         public int worldId;
         public int stageId;
+        public int randomBuffId;
```
ipdae, line 32: This field can to be nullable.

ipdae (commented):
we can create or update `HackAndSlashBuffState` only when failed 3 wave clear.

Lib9c/Action/HackAndSlash.cs
sonohoshi: That's right, so if the field is empty or incorrect, it's going to operate to select the best buff on our own.
It also does not consider the wave clear count when creating or updating a state.

PR #1088, opened 2022-06-02
Add zero item for chain consistency, in the condition

moreal:
This pull request is continued from #1073. The #1073 PR will break the chain consistency on the `previewnet` network because it was changed to update states differently. So this pull request makes the action execute like the original action if it's `previewnet` and the block index is under 1,200,000.


ipdae (approved):
it would be better add test case for avoid case.

PR #1089, opened 2022-06-02
Feature/fix policy

limebell:
Contains 2 changes which are:
- Remove authorized miner checking in `BlockPolicy`.
- Do not append block when mining.


Lib9c/Policy/BlockPolicySource.cs
```diff
@@ -234,11 +234,7 @@ public partial class BlockPolicySource
                     authorizedMinersPolicy,
                     defaultAlgorithm: chain => DifficultyAdjustment<NCAction>.BaseAlgorithm(
                         chain, BlockInterval, DifficultyStability, minimumDifficulty));
-            Func<Address, long, bool> isAllowedToMine = (address, index) => IsAllowedToMineRaw(
-                address,
-                index,
-                authorizedMinersPolicy,
-                permissionedMinersPolicy);
+            Func<Address, long, bool> isAllowedToMine = (address, index) => true;
```
riemannulus, line 237: I guess we are unnecessary this lambda. right?

limebell: Yes. This is an another temporary fix to run PBFT nodes ASAP///

riemannulus: Why can't we clear the lambda?

limebell: `BlockPolicy()` constructor takes `isAllowedToMine` parameter and it's default value is `null` not `true`.
https://github.com/planetarium/lib9c/blob/pbft/Lib9c/Policy/BlockPolicySource.cs#L260
https://github.com/planetarium/lib9c/blob/pbft/Lib9c/Policy/BlockPolicy.cs#L34

riemannulus: LGTM

PR #1102, opened 2022-06-03
Introduce `StakeRegularRewardSheet.Row.RewardInfo.Type` field

### LINKED ISSUE #1093: Introduce `StakeRegularRewardSheet.defaultReward` field

moreal:
This pull request introduces a new field `type` in the `StakeRegularRewardSheet`. There are two types, `arithmetic` and  `fixed`.
- `Arithmetic` type is the default type which uses the second field as `rate`. It gives rewards via the calculation formula, `stakedAmount / rate`. 
- But the `Fixed` type uses the second field as `count`. It gives rewards as the count.

For example, if there is `1,10,400000,1,fixed` row, it means:

> It's level 1 and requires 10 NCG. It'll give the 400000 id item as 1 count for every reward interval.

----

- @jaeho0103 Please review the table sheet changes.

----

This pull request closes #1093.


Lib9c/TableData/StakeRegularRewardSheet.cs
```diff
@@ -32,35 +38,45 @@ public override int GetHashCode()
...
-            public RewardInfo(int itemId, int rate)
+            public RewardInfo(int itemId, int rateOrCount, RewardType type)
             {
                 ItemId = itemId;
-                Rate = rate;
+                RateOrCount = rateOrCount;
+                Type = type;
             }
 
             public RewardInfo(Dictionary dictionary)
```
ipdae, line 65: There is no case RewardInfo serialize/deserialize to `IValue`.

moreal: I removed it in [dfc9470](https://github.com/planetarium/lib9c/pull/1102/commits/dfc9470600da5d781191357f8599aaa05b43da7a)

moreal:
As a record, there was chatting about the suggestion to split table sheets by their purpose instead of using the `type` column in the internal chat.

moreal:
> As a record, there was chatting about the suggestion to split table sheets by their purpose instead of using the `type` column in the internal chat.

- I'll introduce a new sheet called `StakeRegularFixedRewardSheet`, split from current `StakeRegularRewardSheet` implementation. On the sheet, every row will describe fixed rewards without confusion.
- At the `ClaimStakeReward` action, it will try to load `StakeRegularFixedRewardSheet` and gives fixed rewards if there are. If not, it will skip the step.

moreal:
> > As a record, there was chatting about the suggestion to split table sheets by their purpose instead of using the `type` column in the internal chat.
> 
>     * I'll introduce a new sheet called `StakeRegularFixedRewardSheet`, split from current `StakeRegularRewardSheet` implementation. On the sheet, every row will describe fixed rewards without confusion.
> 
>     * At the `ClaimStakeReward` action, it will try to load `StakeRegularFixedRewardSheet` and gives fixed rewards if there are. If not, it will skip the step.

I finished applying your suggestion. In your working time, please take a review the commits.

PR #1120, opened 2022-06-09
Restrict update staking when it is claimable

moreal:


area363:
https://github.com/planetarium/lib9c/runs/6806578697?check_suite_focus=true#step:7:1271
Could you check these errors? 🙏

moreal:
> https://github.com/planetarium/lib9c/runs/6806578697?check_suite_focus=true#step:7:1271 Could you check these errors? 🙏

Thanks for letting me know it. This pull request would restrict updating the staking amount even if the lockup was unlocked. I pushed a new commit to produce the situations in testcases. Thanks!

PR #1124, opened 2022-06-10
Merge `2022q2` to `previewnet`

moreal:
SSIA.

PR #1126, opened 2022-06-10
Trust lockup interval processing

moreal:
This pull request assumes it can be trusted in all cases if the case before just one block up to lockup interval is verified.

Originally, the #1124 build failed because the previewnet lockup interval was different from the 2022q2 branch. Because originally it subtracts over 50000 though the previewnet's lockup interval is about 5000. So this pull request resize the domain into only one case.


area363 (approved):

moreal:
I'll merge when the build passed.

moreal:
I did a mistake to target 2022q2.. I'll open two pull requests to apply to `features/stake` and to merge `features/stake` to `2022q2`.

PR #1136, opened 2022-06-14
Block 'ranking battle action' using arena sheet

hskim881028:
SSIA


ipdae (changes requested):
i think obsolete all previous ranking_battle action.

ipdae (approved):

PR #1138, opened 2022-06-15
Limit minimum staking amount

moreal:
By the request from @jaeho0103, I open a new pull request to restrict staking less amount than the minimum required gold.


.Lib9c.Tests/Action/StakeTest.cs
```diff
@@ -132,6 +134,19 @@ public void Execute_Throws_WhenCancelOrUpdateWhileLockup()
             }));
         }
 
+        [Fact]
+        public void Execute_Throws_WhenStakeLessThanMinimumRequiredGold()
+        {
+            Assert.Equal(50, _tableSheets.StakeRegularRewardSheet.Min(x => x.Value.RequiredGold));
```
BasixKOR, line 140: I think checking the requiredGold is greater than 10 would be enough in this case.

BasixKOR (approved):

PR #1180, opened 2022-06-22
Prevent renew staking with same amount

moreal:
Originally, it renewed staking with the same amount and set `startedBlockIndex` to 0 again. This pull request restricts a behavior like it since the specific block index, `4611070`.


moreal:
Could you review this pull request? @ipdae @planetarium/9c-dev ?

In this case, I thought it is better to branch since the `4611070` block index than upgrade the version of the `Stake` action because it has only one change, `>` -> `>=`. But if you have a different opinion about this, please feel free to let me know it. 🙏🏻

ipdae (approved):

PR #1187, opened 2022-06-27
feat(#1087): introduce useful `IActionContext` extensions

### LINKED ISSUE #1087: Suggestion: Make extension API to detect `previewnet`, `mainnet` easily
While working with Lib9c actions, we should execute conditionally sometimes for chain consistency. In my case, I checked it with `NCG`'s minter address and I thought it'll be helpful if there is `IActionContext.IsPreviewnet` or something else.

moreal:
This pull request resolves #1087.

This pull request introduces new extension methods of the `IActionContext` interface to help branch bugfixes by chain and block index. The methods are the same as the below list:

- `bool Since(long blockIndex)`
- `bool IsMainNet()`

I expected a use-case like the below code:

```csharp
const long hardforkBlockIndex = 1_000_000;
if (context.IsMainNet() && context.Since(hardforkBlockIndex))
```

I wrote code assuming that the `GoldCurrencyState.Currency` will not be changed. I think the gold currency should be static because everyone will lose their NCG if the gold currency is altered and it becomes a new standard. If you have an opinion about this, feel free to start a discussion about it.

And I wonder your opinions about what is better between using a new extension method `Since()` and comparing it by using `IActionContext.BlockIndex` directly, in your experience side.

Of course, we can close this pull request if the discussion's result is these changes are useless or not helpful.

### DIFF
--- a/Lib9c/Action/ActionContextExtensions.cs
+++ b/Lib9c/Action/ActionContextExtensions.cs
@@ -1,0 +1,22 @@
+using Libplanet;
+using Libplanet.Action;
+
+namespace Nekoyume.Action
+{
+    public static class ActionContextExtensions
+    {
+        public static bool IsMainNet(this IActionContext context)
+        {
+            var goldCurrency = context.PreviousStates.GetGoldCurrency();
+            return goldCurrency.Minters
+                       .Contains(new Address("47d082a115c63e7b58b1532d20e631538eafadde"))
+                   && goldCurrency.Ticker == "NCG"
+                   && goldCurrency.DecimalPlaces == 2;
+        }
+
+        public static bool Since(this IActionContext context, long blockIndex)
+        {
+            return blockIndex <= context.BlockIndex;
+        }
+    }
+}
--- a/.Lib9c.Tests/Action/ActionContextExtensionsTest.cs
+++ b/.Lib9c.Tests/Action/ActionContextExtensionsTest.cs
@@ -1,0 +1,137 @@
+namespace Lib9c.Tests.Action
+{
+    using System.Collections.Generic;
+    using Libplanet;
+    using Libplanet.Action;
+    using Libplanet.Assets;
+    using Libplanet.Crypto;
+    using Nekoyume;
+    using Nekoyume.Action;
+    using Nekoyume.Model.State;
+    using Xunit;
+
+    public class ActionContextExtensionsTest
+    {
+        public static IEnumerable<object[]> IsMainNetTestcases()
+        {
+            yield return new object[]
+            {
+                new GoldCurrencyState(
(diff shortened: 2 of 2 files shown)


area363 (approved):
I think this is a good feature moving forward! 👍

boscohyun (approved):

moreal:
First, thanks to review this pull request. 🙏🏻 

Though both a teammate of @planetarium/9c-backend and a teammate of @planetarium/9c-dev reviewed and approved this pull request, I want to receive more opinions and I know that they are preparing the v100230 release. Because it isn't in a hard schedule, I'll make this pull request stay open until 2022.06.30 and I'll merge this pull request if there are no more reviews. I'll remind you on 2022.06.29 and 2022.06.30. 🔔 

Thank you. 🙏🏻

moreal:
As I know, the v100230 and v100231 were released yesterday. So I remind you about this review request about **this feature's usage and function naming or else** because you have touched lib9c codes frequently while the last month. cc @ipdae @hskim1028 @Unengine @sonohoshi 
image

Lib9c/Action/ActionContextExtensions.cs
```diff
@@ -0,0 +1,31 @@
...
+        }
+
+        public static bool IsMainNet(this IActionContext context)
+        {
+            var goldCurrency = context.PreviousStates.GetGoldCurrency();
+            return goldCurrency.Minters
+                       .Contains(new Address("47d082a115c63e7b58b1532d20e631538eafadde"))
+                   && goldCurrency.Ticker == "NCG"
+                   && goldCurrency.DecimalPlaces == 2;
+        }
```
ipdae, line 24: I suggest like this. and i think previewnet minter can be change because we can reset previewnet be required.

```
public enum NetworkType
{
    Mainnet,
    Previewnet,
    Internal
}

public NetworkType GetNetworkType(this IActionContext ctx)
{
    var gc = ctx.GetGoldCurrency();
    if (gc.Contains(mainnetAddress)
        return NetoworkType.Mainnet;
}
```

moreal: Then, if it is assumed that the gold currency's minter can be changed, I guess this feature cannot be implemented. Right?

ipdae: I think, it would be helpful we can check `IsMainnet` right now.

moreal: Then I'll remove `IsPreviewNet()` method. Thanks for your opinion.

moreal:
I removed `ActionContextExtensions.IsMainNet(...)` method so I requested reviews again. 🙏🏻

ipdae (approved):

PR #1200, opened 2022-07-05
Suggestion: Snapshot Testing

moreal:
## Why to use Snapshot Testing?

This pull request is not perfect. I wanted to suggest testing with Snapshot for checking deterministic actions. Every actions' serialized result and execution result must be deterministic. In other words, they must be always the same in the same input (e.g. block index, policy, actions' parameters).

Let's think about the situation where one action, named `SomeAction`, is.

```csharp
class SomeAction : IAction {
  public IAccountStateDelta Execute(IActionContext context) {
    return context.PreviousStates.SetState("a", "a");
  }
}
```

Then the result states will be `{ a: a }`. But someday someone modified the code and the result became `{ a: b }`. Then the `SomeAction`'s consistency was broken. Currently, tests don't store its execution result as snapshot so we cannot realize the accident easily.

## How do we snapshot testing

This pull request suggests to use [`Verify`](https://github.com/VerifyTests/Verify/) package. When you run `dotnet test` command, the tests written with `Verify` also run.

```csharp
// 1. Make a class with `UsesVerify`
using VerifyXunit;

[UsesVerify]
public class TransferAsset0SnapshotTest
{
}

// 2. Import ActionUtils static class.
using static ActionUtils;

// 3. Write tests to take a snapshot like below case. The below test will take a snapshot of `PlainValue`, in other words, the action's serialization format.
using Xunit;

[Fact]
public Task PlainValue()
{
    var action = new TransferAsset0(
        default(Address),
        default(Address),
        Currency.Legacy("NNN", 2, null) * 100);

    return Verifier.Verify(action.PlainValue)
        .UseTypeName(GetActionTypeId<TransferAsset0>());
}

// 4. Accpet received.txt file.
// 5. Rerun test
```

## Request reviews 🙏🏻 

- @planetarium/9c-backend @planetarium/9c-dev Lib9c's main commiters

## Problems

 - ~~When writing snapshot test for actions, how to resolve its name? The snapshot filename is derived from the test's class name. But nine-chronicles commiters are using the action name without version number as the latest action. So when upgrading the action, the snapshot test will be broken and we should rename the snapshot files. It's easy to do mistake so it is bad pattern.~~
   - `VerifyTests` provides `UseTypeName` to override type name with custom type name. In our case, it should be used with actions' type id (e.g. `transfer_asset`, `transfer_asset2`).


moreal:
This pull request can be merged after the `Tx` Lib9c.Tools commands are fixed.

.Lib9c.Tests/Action/Snapshot/ActionUtils.cs
```diff
@@ -0,0 +1,20 @@
...
+        {
+            Type attrType = typeof(ActionTypeAttribute);
+            Type actionType = typeof(T);
+            return actionType.IsDefined(attrType) &&
+                ActionTypeAttribute.ValueOf(actionType) is { } tid
+                ? tid
+                : throw new MissingActionTypeException("The action type", actionType);
+        }
+    }
+}
```
longfin, line 20:
```suggestion
}

```

longfin (approved):

PR #1204, opened 2022-07-11
Improve `HackAndSlashTest` about selecting crystal random buff

sonohoshi:
- Add `CrystalRandomSkillState.GetHighestRankSkill()`
- Add test case to `HackAndSlashTest` about using random skill without `stageBuffId` argument.


Lib9c/Model/State/CrystalRandomSkillState.cs
```diff
@@ -73,5 +73,12 @@ public IValue Serialize()
 
             return SkillFactory.Get(skillRow, default, 100);
         }
+
+        public int GetHighestRankSkill(CrystalRandomBuffSheet crystalRandomBuffSheet)
+        {
+            return SkillIds
+                .OrderBy(key => crystalRandomBuffSheet[key].Rank)
+                .FirstOrDefault();
+        }
```
boscohyun, line 82: Why removed `.ThenBy(id => id)` after `.OrderBy(key => ...)`?
If there are Ids that has same `Rank`, this function possible to has nondeterministic issue.

sonohoshi: Oh it's my mistake. Thx for your comment, I will fix it!

boscohyun (approved):

PR #1212, opened 2022-07-15
fix action evaluation results in `v100240`

area363:
When we updated the mainnet version from `v230` to `v240`, there were some deployment errors which ended up causing blocks from `#4374157` and `#4377430` to be evaluated differently compared to the canonical chain. I've narrowed down the specific actions (`Buy11`, `CombinationEquipment11`, `HackAndSlashSweep4`, `ItemEnhancement10`, `RankingBattle`, and `RapidCombination6`) causing these problems and added conditions that would allow the execution results to match the canonical chain.

### DIFF
--- a/Lib9c/Action/Buy11.cs
+++ b/Lib9c/Action/Buy11.cs
@@ -104,7 +104,14 @@ namespace Nekoyume.Action
             var arenaSheetState = states.GetState(arenaSheetAddress);
             if (arenaSheetState != null)
             {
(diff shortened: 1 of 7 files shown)


area363:
I'll fix the tests first.

area363:
@planetarium/9c-dev @planetarium/9c-backend PTAL, this fix has been tested by appending blocks `#4133000` to `#4380000`. 🙏

ipdae (approved):

PR #1213, opened 2022-07-15
Introduce `EventDungeonInfo`

boscohyun:
The `EventDungeonInfo` class is a model stored at a unique address derived from the address of a specific `AvatarState` and the `Id` of a specific `EventDungeonSheet.Row`.

The `EventDungeonInfo` class stores the number of remaining tickets for a specific event dungeon in a specific `AvatarState`, and the `Id` of the highest cleared `EventDungeonStageSheet.Row` is stored.

Unlike the `WorldInformation` class, the `EventDungeonInfo` class does not validate the game logic deeply. Validation delegates to actions using the `EventDungeonInfo` class, etc.

### DIFF
--- a/Lib9c/Model/Event/EventDungeonInfo.cs
+++ b/Lib9c/Model/Event/EventDungeonInfo.cs
@@ -40,14 +40,36 @@ namespace Nekoyume.Model.Event
             .Add(_remainingTickets.Serialize())
             .Add(_clearedStageId.Serialize());
 
-        public void ResetTickets(int tickets) =>
-            _remainingTickets = tickets;
-
-        public bool HasTickets(int tickets) =>
-            _remainingTickets >= tickets;
-
-        public bool TryUseTicket(int tickets)
+        public void ResetTickets(int tickets)
         {
+            if (tickets < 0)
+            {
+                throw new ArgumentException(
+                    $"{nameof(tickets)} must be greater than or equal to 0.");
+            }
+
+            _remainingTickets = tickets;
+        }
+
+        public bool HasTickets(int tickets)
+        {
+            if (tickets < 0)
+            {
+                throw new ArgumentException(
+                    $"{nameof(tickets)} must be greater than or equal to 0.");
+            }
+
+            return _remainingTickets >= tickets;
+        }
+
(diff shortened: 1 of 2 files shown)


Lib9c/Model/Event/EventDungeonInfo.cs
```diff
@@ -0,0 +1,92 @@
...
+        private int _remainingTickets;
+        private int _clearedStageId;
+
+        public EventDungeonInfo()
+        {
+            _remainingTickets = 0;
+            _clearedStageId = 0;
+        }
+
+        public EventDungeonInfo(Bencodex.Types.IValue serialized)
```
ipdae, line 25:
```suggestion
        public EventDungeonInfo(Bencodex.Types.List serialized)
```

Lib9c/Model/Event/EventDungeonInfo.cs
```diff
@@ -0,0 +1,92 @@
...
+        public bool TryUseTicket(int tickets)
+        {
+            if (_remainingTickets < tickets)
+            {
+                return false;
+            }
+
+            _remainingTickets -= tickets;
+            return true;
+        }
```
ipdae, line 58: it would be better check negative case.

Lib9c/Model/Event/EventDungeonInfo.cs
```diff
@@ -0,0 +1,92 @@
...
+{
+    public class EventDungeonInfo : IState
+    {
+        public static Address DeriveAddress(Address address, int dungeonId)
+        {
+            return address.Derive($"event_dungeon_info_{dungeonId}");
+        }
+
+        private int _remainingTickets;
+        private int _clearedStageId;
```
ipdae, line 17: It only have the last cleared stageId. is this order of stage id not changed? it would be better save list of stage id.

ipdae (commented):
test can be simplify with [Theory]

Lib9c/Model/Event/EventDungeonInfo.cs
boscohyun: I plan to delegate that part to using this class.

ipdae (approved):

PR #1218, opened 2022-07-18
Introduce `EventDungeonBattle` action

boscohyun:
- Introduce `InvalideActionFieldsException`
- Introduce `EventDungeonExtensions`

- Introduce `NotEnoughEventDungeonTicketsException`
- Introduce `EventDungeonBattleSimulatorSheets`
- Introduce `EventDungeonBattleSimulator`
- Introduce `EventDungeonBattle` action
- (WIP)Write tests for `EventDungeonBattle`

### DIFF
--- a/.Lib9c.Tests/Action/ExceptionTest.cs
+++ b/.Lib9c.Tests/Action/ExceptionTest.cs
@@ -59,6 +59,7 @@ namespace Lib9c.Tests.Action
         [InlineData(typeof(AlreadyRecipeUnlockedException))]
         [InlineData(typeof(InvalidRecipeIdException))]
         [InlineData(typeof(AlreadyWorldUnlockedException))]
+        [InlineData(typeof(InvalidActionFieldException))]
         public void Exception_Serializable(Type excType)
         {
             if (Activator.CreateInstance(excType, "for testing") is Exception exc)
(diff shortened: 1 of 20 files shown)


ipdae:
build failed. 😅

ipdae (commented):
only the sheet list is different between `EventDungeonBattleSimulator` and `StageSimulator` and they run same simulate. right? If so, if bind the list of sheets received as a constructor with an interface or inheritance, we don't need to create a new simulator.

boscohyun:
> only the sheet list is different between `EventDungeonBattleSimulator` and `StageSimulator` and they run same simulate. right? If so, if bind the list of sheets received as a constructor with an interface or inheritance, we don't need to create a new simulator.

@ipdae Yes, your prediction is almost correct. And `EventDungeonSimulator` has little different with `StageSimulator`.

1. `Exp` amount.
  - `StageSimulator`: [StageRewardExpHelper.GetExp(avatarState.level, stageId)][1]
  - `EventDungeonSimulator`: default(0) or other formula.
2. Check a cleared stage.
  - `StageSimulator`: [avatarState.worldInformation.IsStageCleared(StageId)][2]
  - `EventDungeonSimulator`: `eventDungeonInfo.IsCleared(eventDungeonStageId)`

There is not enough time to consider a better way to me.😇

[1]: https://github.com/planetarium/lib9c/blob/feature%2Fevent-dungeons/Lib9c/Battle/StageSimulator.cs#L69
[2]: https://github.com/planetarium/lib9c/blob/feature%2Fevent-dungeons/Lib9c/Battle/StageSimulator.cs#L58

ipdae:
@boscohyun https://www.diffchecker.com/KLeRNAQk The difference in simulate seems to be the stage ID check when calculate exp. If the data of the event dungeon is managed in a separate table, it would be sufficient to limit only the id.

sonohoshi (approved):

PR #1229, opened 2022-07-27
Add PlayCount:int field to HackAndSlash and Version up

sonohoshi:
SSIA

The `PlayCount` field controls loop count of `StageSimulator.Simulate(1)`.

### DIFF
--- a/Lib9c/Action/HackAndSlash.cs
+++ b/Lib9c/Action/HackAndSlash.cs
@@ -1,4 +1,4 @@
-using System;
+﻿using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Diagnostics;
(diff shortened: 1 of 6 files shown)


ipdae:
test failed

Lib9c/Action/HackAndSlash.cs
```diff
@@ -274,81 +230,94 @@ public override IAccountStateDelta Execute(IActionContext context)
...
-                    worldId,
-                    stageId,
-                    blockIndex,
-                    worldSheet,
-                    sheets.GetSheet<WorldUnlockSheet>()
-                );
+                // First simulating will use Foods and Random Skills.
+                // Remainder simulating will not use Foods.
+                var simulator = new StageSimulator(
+                    new HackAndSlashRandom(random.Next()),
```
ipdae, line 244: Whats the different between using has `HackAndSlashRandom` and `random`?

sonohoshi: I expected a problem with reusing random objects, but I was wrong.
I will remove `HackAndSlashRandom`. 😅

ipdae (approved):

PR #1235, opened 2022-07-29
Introduce ClaimWorldBossKillReward

ipdae:
- Update sheet data
- Use rune currency ticker instead of rune id.
- Claim boss kill reward.


hskim881028:
Conflicting files (RuneWeightSheet.csv)  👀

PR #1238, opened 2022-08-01
Introduce `ISimulator` and `IStageSimulator` interface

boscohyun:
This refactoring is a preparation work to make the `StageSimulator` hard forking work smoothly.

- Introduce `ISimulator`
- Change argument type of `Wave.Spawn()` method from `StageSimulator` to `ISimulator`
- Introduce `IStageSimulator`
- Use `IStageSimulator` instead of `StageSimulator` on `AvatarState.Update()` method
- Use `IStageSimulator` instead of `StageSimulator` on `Enemy` class


sonohoshi (approved):

Lib9c/Model/Character/Enemy.cs
```diff
@@ -49,18 +48,23 @@ private void PostConstruction()
...
         protected sealed override void SetSkill()
         {
             base.SetSkill();
 
-            var dmg = (int) (ATK * 0.3m);
-            var skillIds = StageSimulator.EnemySkillSheet.Values.Where(r => r.characterId == RowData.Id)
-                .Select(r => r.skillId).ToList();
-            var enemySkills = Simulator.SkillSheet.Values.Where(r => skillIds.Contains(r.Id))
+            var dmg = (int)(ATK * 0.3m);
+            var skillIds = ((IStageSimulator)Simulator)
```
moreal, line 60: I'm not an expert on this project but the `Enemy` class requires the `Simulator` to be `IStageSimulator`. In my opinion, it seems better to check the type in the constructor like before because it prevents to create `Enemy` object with `Simulator` not typed `IStageSimulator`.

boscohyun: @moreal It seems be possible to reduce member field when I was refactoring the `Enemy` class.
But I also agree to validate arguments in constructor.
Thanks!

cc @ipdae

ipdae (approved):

PR #1256, opened 2022-08-04
Introduce `HammerPointState` for new feature of combination

sonohoshi:
SSIA

```
public class HammerPointState : IState
{
    public Address Address { get; }
    public int RecipeId { get; }
    public int HammerPoint { get; private set; }
}
```

### DIFF
--- a/.Lib9c.Tests/Model/State/HammerPointStateTest.cs
+++ b/.Lib9c.Tests/Model/State/HammerPointStateTest.cs
@@ -1,0 +1,53 @@
+﻿namespace Lib9c.Tests.Model.State
+{
+    using System.Collections.Generic;
+    using Bencodex.Types;
+    using Libplanet;
+    using Libplanet.Crypto;
+    using Nekoyume.Model.State;
+    using Xunit;
+
+    public class HammerPointStateTest
+    {
+        private readonly TableSheets _tableSheets;
+
+        public HammerPointStateTest()
+        {
+            var sheets = TableSheetsImporter.ImportSheets();
+            _tableSheets = new TableSheets(sheets);
+        }
+
+        [Fact]
+        public void Serialize()
+        {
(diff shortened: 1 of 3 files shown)


Lib9c/Model/State/HammerPointState.cs
```diff
@@ -0,0 +1,43 @@
...
+        {
+            return List.Empty
+                .Add(ItemId.Serialize())
+                .Add(HammerPoint.Serialize());
+        }
+
+        public void UpdateHammerPoint(int point)
+        {
+            HammerPoint += point;
+        }
```
boscohyun, line 36: How about rename this like `AddHammerPoint` or `IncreaseHammerPoint`?

.Lib9c.Tests/Model/State/HammerPointStateTest.cs
```diff
@@ -0,0 +1,53 @@
+﻿namespace Lib9c.Tests.Model.State
+{
+    using System.Collections.Generic;
+    using Bencodex.Types;
+    using Libplanet;
+    using Libplanet.Crypto;
+    using Nekoyume.Model.State;
+    using Xunit;
+
+    public class HammerPointStateTest
```
boscohyun, line 10: Can you please add a test that use `UpdateHammerPoint()` method?

Lib9c/Model/State/HammerPointState.cs
sonohoshi: I like 'Update' word but understand 'Increase' is better than it. I will change it~

.Lib9c.Tests/Model/State/HammerPointStateTest.cs
sonohoshi: OK I'll add that. THX

boscohyun (approved):

PR #1257, opened 2022-08-04
Introduce CrystalHammerPointSheet

sonohoshi:
for new combination equipment feature

```cs
public class CrystalHammerPointSheet : Sheet<int, CrystalHammerPointSheet.Row>
{
    public class Row : SheetRow<int>
    {
        public int RecipeId { get; private set; }
        public int MaxPoint { get; private set; }
        public int CRYSTAL { get; private set; }
        public override int Key => RecipeId;
    }
}
```


hskim881028 (approved):

Lib9c/Model/State/HammerPointState.cs
```diff
@@ -30,9 +32,9 @@ public IValue Serialize()
                 .Add(HammerPoint.Serialize());
         }
 
-        public void AddHammerPoint(int point)
+        public void AddHammerPoint(int point, CrystalHammerPointSheet sheet)
         {
-            HammerPoint += point;
+            HammerPoint = Math.Min(HammerPoint + point, sheet[RecipeId].MaxPoint);
```
boscohyun, line 37: It will be better to do a `null` check for the `sheet` argument.

Lib9c/TableData/Crystal/CrystalHammerPointSheet.cs
```diff
@@ -0,0 +1,27 @@
...
+            public int RecipeId { get; private set; }
+            public int MaxPoint { get; private set; }
+            public int CRYSTAL { get; private set; }
+            public override int Key => RecipeId;
+
+            public override void Set(IReadOnlyList<string> fields)
+            {
+                RecipeId = ParseInt(fields[0]);
+                MaxPoint = ParseInt(fields[1]);
+                CRYSTAL = ParseInt(fields[2]);
```
boscohyun, line 19: Can you set a default value here?

```suggestion
                RecipeId = ParseInt(fields[0], 0);
                MaxPoint = ParseInt(fields[1], 0);
                CRYSTAL = ParseInt(fields[2], 0);
```

Lib9c/Model/State/HammerPointState.cs
sonohoshi: OK I will add that!

PR #1258, opened 2022-08-04
Introduce TransferAssets

ipdae:
- Introduce multiple transfer assets
- allowed recipient to avatar

### DIFF
--- a/.Lib9c.Tests/Action/ActionEvaluationTest.cs
+++ b/.Lib9c.Tests/Action/ActionEvaluationTest.cs
@@ -101,6 +101,7 @@ namespace Lib9c.Tests.Action
         [InlineData(typeof(UnlockEquipmentRecipe))]
         [InlineData(typeof(UnlockWorld))]
         [InlineData(typeof(Raid))]
+        [InlineData(typeof(TransferAssets))]
         public void Serialize_With_MessagePack(Type actionType)
         {
             var action = GetAction(actionType);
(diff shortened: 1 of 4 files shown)


longfin:
As far I known, `TransferAssets` has the purpose that sending asset to avatars. but if we have only that, we don't need a new action and enables that on `TrnasferAsset` already exists. 

To enable transferring assets to avatar, I think that we have to loosen constraints for #623. it may be accomplish by...

- Adding a flag to distinct `recipient`'s type (i.e., agent, avatar)
- Or, Adding another state lookup to determine [when failed on activation check](https://github.com/planetarium/lib9c/blob/v100260/Lib9c/Action/TransferAsset.cs#L103).

I guess second option is also possible if we don't care about very frequently cases. (if we care, may need to choose other derived address from avatar)

However, if there are another reasons about multiplexing recipients as this PR proposed, please let me know. 🙇 

@ipdae

ipdae:
@longfin 
- `RecipientAgentAddress` has check  an activation  and only the avatar included in the recipient can be received.
- listing `FungibleAssetValue` and `RecipientAddress` has for the following reasons.
  - `FungibleAssetValue` now managed into agents and avatars.   
  - 4 types of assets are paid as rewards in actual TransferAssets using case.
  - it would be better to sign tx only once than to sign multiple as many types of FungibleAssetValue.

PR #1276, opened 2022-08-09
Correct `Stake` action

moreal:
This pull request does:

 - Correct `stake` action.
 - Introduce a new action called `stake2` to continue `stake`.
   - `stake2` action became `GameAction` now.
 
You can review it through commits because they have a commit to copy files.


area363 (approved):
👍

PR #1292, opened 2022-08-11
Feature/optional admin state

### LINKED ISSUE #1287: Make `AdminState` in `InitializeStates` optional
In NineChronicles Headless, we have to provide admin address to create genesis block.
This `Admin` address is useful in several situations, but I think this required admin address raises hurdle to make local genesis block and start new network.

So I suggest to make `AdminState` optional in `InitializeStates` method. This will make admin-less blockchain network, but that is no big deal for local network.

I'm not sure how much works to achieve this, so please comment what you think about before I start.

U-lis:
resolves #1287 

---

`AdminState` is now optional and `InitializeStates` handles null case.
Add testcase with null `AdminState` and made it pass.

### DIFF
--- a/Lib9c/Action/InitializeStates.cs
+++ b/Lib9c/Action/InitializeStates.cs
@@ -29,27 +29,27 @@ namespace Nekoyume.Action
     [ActionType("initialize_states")]
     public class InitializeStates : GameAction
     {
-        public Bencodex.Types.Dictionary Ranking { get; set; } = Dictionary.Empty;
-        public Bencodex.Types.Dictionary Shop { get; set; } = Dictionary.Empty;
+        public Dictionary Ranking { get; set; } = Dictionary.Empty;
+        public Dictionary Shop { get; set; } = Dictionary.Empty;
         public Dictionary<string, string> TableSheets { get; set; }
-        public Bencodex.Types.Dictionary GameConfig { get; set; } = Dictionary.Empty;
-        public Bencodex.Types.Dictionary RedeemCode { get; set; } = Dictionary.Empty;
+        public Dictionary GameConfig { get; set; } = Dictionary.Empty;
+        public Dictionary RedeemCode { get; set; } = Dictionary.Empty;
 
-        public Bencodex.Types.Dictionary AdminAddress { get; set; } = Dictionary.Empty;
+        public Dictionary AdminAddressState { get; set; } = Dictionary.Empty;
 
-        public Bencodex.Types.Dictionary ActivatedAccounts { get; set; } = Dictionary.Empty;
+        public Dictionary ActivatedAccounts { get; set; } = Dictionary.Empty;
 
-        public Bencodex.Types.Dictionary GoldCurrency { get; set; } = Dictionary.Empty;
+        public Dictionary GoldCurrency { get; set; } = Dictionary.Empty;
 
-        public Bencodex.Types.List GoldDistributions { get; set; } = List.Empty;
+        public List GoldDistributions { get; set; } = List.Empty;
 
-        public Bencodex.Types.List PendingActivations { get; set; } = List.Empty;
-
-        // This property can contain null:
-        public Bencodex.Types.Dictionary AuthorizedMiners { get; set; } = null;
+        public List PendingActivations { get; set; } = List.Empty;
 
         // This property can contain null:
-        public Bencodex.Types.Dictionary Credits { get; set; } = null;
+        public Dictionary AuthorizedMiners { get; set; }
+
+        // This property can contain null:
+        public Dictionary Credits { get; set; }
 
         public InitializeStates()
         {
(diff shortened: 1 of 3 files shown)


moreal (commented):
I'll review this pull request again after my high-priority task but I left a comment.

moreal:
You can get the `initialize_states` by using the script.

https://gist.github.com/moreal/a1c90b029903251522a68566d8b23ce9

Lib9c/Action/InitializeStates.cs
```diff
@@ -149,11 +149,15 @@ public override IAccountStateDelta Execute(IActionContext context)
                 .SetState(ShopState.Address, Shop)
                 .SetState(GameConfigState.Address, GameConfig)
                 .SetState(RedeemCodeState.Address, RedeemCode)
-                .SetState(AdminState.Address, AdminAddress)
                 .SetState(ActivatedAccountsState.Address, ActivatedAccounts)
                 .SetState(GoldCurrencyState.Address, GoldCurrency)
                 .SetState(Addresses.GoldDistribution, GoldDistributions);
 
+            if (AdminAddressState != null)
```
moreal, line 156: I recommend using the `is` operator keyword to compare with the `null` value.

It's because of the below part. ([source](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/type-testing-and-cast#is-operator))

> The is operator doesn't consider user-defined conversions.

U-lis: Related to this, I originally wrote a code like `if (AdminAddressState is not null)` but Rider says I have to update C# language version.
So I changed it to use `!=` operator but I should change that to `if (!(AdminAddressState is null))`.

p.s. Should we update C# language version to use latest features?

moreal: > Should we update C# language version to use latest features?

Yes, we should update C# language version if you want. But the lib9c must guarantee netstandard2.0 so it should check if it doesn't have any side effects.

moreal: It seems okay when I looked [this document](https://docs.microsoft.com/en-us/dotnet/standard/net-standard?tabs=net-standard-1-0#when-to-target-net50-or-net60-vs-netstandard). We can check it by replacing the language version with 9 or 10, and building it.

Lib9c/Action/InitializeStates.cs
```diff
@@ -29,27 +29,27 @@ namespace Nekoyume.Action
...
+        public Dictionary Ranking { get; set; } = Dictionary.Empty;
+        public Dictionary Shop { get; set; } = Dictionary.Empty;
         public Dictionary<string, string> TableSheets { get; set; }
-        public Bencodex.Types.Dictionary GameConfig { get; set; } = Dictionary.Empty;
-        public Bencodex.Types.Dictionary RedeemCode { get; set; } = Dictionary.Empty;
+        public Dictionary GameConfig { get; set; } = Dictionary.Empty;
+        public Dictionary RedeemCode { get; set; } = Dictionary.Empty;
 
-        public Bencodex.Types.Dictionary AdminAddress { get; set; } = Dictionary.Empty;
+        public Dictionary AdminAddressState { get; set; } = Dictionary.Empty;
```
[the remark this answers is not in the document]
U-lis: I thought it's typo issue by seeing [here](https://github.com/planetarium/lib9c/blob/96a122732b6e788e342f47eeb5efeb5048beec0f/Lib9c/Action/InitializeStates.cs#L77); "Why serialize result of `adminAddressState` named `adminAddress`?" 
Am I missing something?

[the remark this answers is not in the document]
moreal: Ah, I didn't intend to remove this part. This comment was for splitting the commit 😅 Because the commit was for '[Remove redundant qualifier](https://github.com/planetarium/lib9c/pull/1292/commits/ee7b6e06a3e31a2cedf973f71f0aca31901fe19c)' but it also renamed the variable unexpectedly. That's what I wanted to say.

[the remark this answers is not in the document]
U-lis: @moreal  Oh, I see. As you say, it's my bad using `git add -p` command.  sry for mixed commit contents.

Lib9c/BlockHelper.cs
```diff
@@ -57,7 +57,7 @@ public static class BlockHelper
                 redeemCodeState: new RedeemCodeState(redeemCodeListSheet),
                 adminAddressState: adminState,
                 activatedAccountsState: new ActivatedAccountsState(
-                    isActivateAdminAddress
+                    isActivateAdminAddress && adminState != null
```
moreal, line 60:
```suggestion
                    isActivateAdminAddress && !(adminState is null)
```

Lib9c/Action/InitializeStates.cs
```diff
@@ -150,11 +149,15 @@ public override IAccountStateDelta Execute(IActionContext context)
                 .SetState(ShopState.Address, Shop)
                 .SetState(GameConfigState.Address, GameConfig)
                 .SetState(RedeemCodeState.Address, RedeemCode)
-                .SetState(AdminState.Address, AdminAddress)
                 .SetState(ActivatedAccountsState.Address, ActivatedAccounts)
                 .SetState(GoldCurrencyState.Address, GoldCurrency)
                 .SetState(Addresses.GoldDistribution, GoldDistributions);
 
+            if (AdminAddressState != null)
```
moreal, line 156:
```suggestion
            if (!(AdminAddressState is null))
```

moreal (commented):
Can they be applied also through 07b053665fbce6c4093e08a05650edf2b61b1355 commit?

moreal (approved):

Lib9c/Action/InitializeStates.cs
U-lis: About updating C# language version, continue at #1295

Issue #1251, opened 2022-08-03
Staking rewards' intervals should be static


moreal:
It doesn't allow to claim stake reward `StakeState.RewardInterval * 2` if you received at `StakeState.RewardInterval + n`.
But it should allow them claim stake reward even if they received stake rewards at `StakeState.RewardInterval + n`

PR #1303, opened 2022-08-17
Make staking able to claim stake reward at static reward interval

moreal:
This pull request tries to fix https://github.com/planetarium/lib9c/issues/1251.

### DIFF
--- a/.Lib9c.Tests/Model/State/StakeStateTest.cs
+++ b/.Lib9c.Tests/Model/State/StakeStateTest.cs
@@ -1,11 +1,31 @@
 namespace Lib9c.Tests.Model.State
 {
     using Bencodex.Types;
+    using Nekoyume.BlockChain.Policy;
     using Nekoyume.Model.State;
     using Xunit;
 
     public class StakeStateTest
     {
+        [Fact]
+        public void IsClaimable()
+        {
+            Assert.False(new StakeState(
+                    default,
+                    0,
+                    StakeState.RewardInterval + 1,
+                    StakeState.LockupInterval,
+                    new StakeState.StakeAchievements())
+                .IsClaimable(StakeState.RewardInterval * 2));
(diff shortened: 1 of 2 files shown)


moreal:
I wonder the prefered style to fix bugs in this project. I tried to fix this bug by branching the `StakeState.IsClaimable` on `BlockPolicySource.V100280ObsoletedIndex`. The model's schema wasn't changed and only one logic will be changed. 

Why I ask this question, is that there are some models versioned (e.g. MonsterCollectionState, AvatarState).

@ipdae

ipdae:
If the change may occur `InvalidBlockStateRootHashException`, the default policy hard fork them.

moreal:
> If the change may occur `InvalidBlockStateRootHashException`, the default policy hard fork them.

I think it will not occur `InvalidBlockStateRootHashException` if the v100280 version was released well before the block index.

longfin:
@moreal It has conflicts now. 😢 (I'll start to rewiew regardless of that)

Lib9c/Policy/BlockPolicySource.cs
```diff
@@ -80,6 +80,8 @@ public partial class BlockPolicySource
 
         public const long V100270ObsoleteIndex = 4_841_774;
 
+        public const long V100280ObsoleteIndex = 4_972_746;
```
sonohoshi, line 83: I think, need to change it to `V100290ObsoleteIndex`.

longfin (approved):

Lib9c/Policy/BlockPolicySource.cs
moreal: Because there was already `V100290ObsoleteIndex`, I removed it and replace with the `V100290ObsoleteIndex` while resolving the conflicts. 🙏🏻 Please take a look again 🙇🏻‍♂️

PR #1328, opened 2022-08-25
Introduce `BuffRemovalAttack`.

Unengine:
- Implemented skill that removes one of recently added buff.


Lib9c/Model/Character/ArenaCharacter.cs
```diff
@@ -393,6 +393,28 @@ public void AddBuffV1(Buff.Buff buff, bool updateImmediate = true)
...
+                    minDuration = elapsedTurn;
+                    removedBuff = buff;
+                }
+            }
+
+            if (removedBuff != null)
+            {
+                _stats.RemoveBuff(removedBuff);
+                Buffs.Remove(removedBuff.RowData.GroupId);
+                _stats.IncreaseHpForArena();
```
hskim881028, line 414: 👍🏻

hskim881028 (approved):

PR #1338, opened 2022-08-29
Change AP by Staking

sonohoshi:
Change CostAP in HAS & Sweep, by staking amount.


Lib9c/TableCSV/StakeActionPointCoefficientSheet.csv
```diff
@@ -0,0 +1,6 @@
+level,required_gold,coefficient
```
Unengine, line 1: How about containing calculated cost of each action rather than calculating cost in action?

Lib9c/TableData/StakeActionPointCoefficientSheet.cs
```diff
@@ -0,0 +1,40 @@
+﻿using System;
+using System.Collections.Generic;
+using static Nekoyume.TableData.TableExtensions;
+
+namespace Nekoyume.TableData
+{
+    [Serializable]
+    public class StakeActionPointCoefficientSheet : Sheet<int, StakeActionPointCoefficientSheet.Row>, IStakeRewardSheet
+    {
+        public IReadOnlyList<IStakeRewardRow> OrderedRows => OrderedList;
```
boscohyun, line 10: Is `OrderedRows` necessary?
If you think the `OrderedList` is not intuitive, you can rename it.

boscohyun (approved):

Lib9c/TableData/StakeActionPointCoefficientSheet.cs
sonohoshi: That is member of `IStakeRewardSheet`. Just a implementation 😉

boscohyun: Oh, I didn't know about `IStakeRewardSheet`.😅

PR #1345, opened 2022-08-31
Update Validator.ValidateForHackAndSlash to using costAp argument

sonohoshi:
SSIA


Lib9c/Action/HackAndSlash.cs
```diff
@@ -186,6 +168,25 @@ public override IAccountStateDelta Execute(IActionContext context)
             sw.Stop();
             Log.Verbose("{AddressesHex}HAS use ActionPoint: {Elapsed}", addressesHex, sw.Elapsed);
 
+            // Validate about avatar state.
```
boscohyun, line 171:
```suggestion
            sw.Restart();
            // Validate about avatar state.
```

sonohoshi: This part is being processed within the method. 😅

boscohyun (approved):

Issue #1357, opened 2022-09-04
`Libplanet.Blocks.InvalidBlockStateRootHashException: Block #4374126`


area363:
```
Libplanet.Blocks.InvalidBlockStateRootHashException: Block #4374126 
1c920bca6b0425a51f79a92e346bb949aaf26e2671431de7994fc086f605ce36's state root hash is 
7f3fb15818d3cac0b8ba3767765f96e9000f461515ee55975a960f56433dbc13, but the execution result is 
4a2cb9e86d15c001afc74065db78fc48c55325d4dffdba35bc79926d3b854ea4.
```
https://9cscan.com/blocks/4374126
Most likely due to the patch table applied in `#4374126` which is causing different execution results for `HAS13`.

PR #1362, opened 2022-09-07
fix exception handling in Block `#4374126`

area363:
Resolves https://github.com/planetarium/lib9c/issues/1357.


area363:
PTAL 🙏  @planetarium/9c-backend @planetarium/9c-dev

ipdae (approved):

PR #1372, opened 2022-09-13
Bug fixes in raid

Unengine:
- Apply used consumable on states.
- Calculate battle reward based on dealt damage instead of wave number.


ipdae (changes requested):
we must check migration avatar case like #1371

ipdae (approved):

PR #1373, opened 2022-09-14
⬆️ Bump libplanet to 0.41.2

tkiapril:
Most of the content are identical with https://github.com/planetarium/lib9c/pull/1331 before revert, except those added between the revert and head.


Lib9c/CurrencyExtensions.cs
```diff
@@ -26,8 +41,27 @@ public static Currency Deserialize(Bencodex.Types.Dictionary serialized)
             {
                 minters = mintersAsList.Select(b => new Address(((Binary) b).ByteArray)).ToImmutableHashSet();
             }
-            
-            return new Currency((Text)serialized["ticker"], ((Binary)serialized["decimalPlaces"]).First(), minters);
+
+            if (serialized.ContainsKey("totalSupplyTrackable"))
+            {
+                if (serialized.ContainsKey("maximumSupplyMajor"))
```
moreal, line 47: Is it okay to check `maximumSupplyMinor` also?

Lib9c/CurrencyExtensions.cs
```diff
@@ -13,10 +13,25 @@ public static Bencodex.Types.Dictionary Serialize(this Currency currency)
...
+            {
+                serialized = serialized.Add("totalSupplyTrackable", true);
+                if (currency.MaximumSupply.HasValue)
+                {
+                    serialized = serialized.Add(
+                        "maximumSupplyMajor",
+                        (IValue)new Integer(currency.MaximumSupply!.Value.MajorUnit)
+                        ).Add(
+                        "maximumSupplyMinor",
+                        (IValue)new Integer(currency.MaximumSupply!.Value.MinorUnit));
```
moreal, line 30: Instead of using `!` keyword (null-check ignore), how about using `is` keyword? like:

```suggestion
                if (currency.MaximumSupply is { } maximumSupply)
                {
                    serialized = serialized.Add(
                        "maximumSupplyMajor",
                        (IValue)new Integer(maximumSupply.MajorUnit)
                        ).Add(
                        "maximumSupplyMinor",
                        (IValue)new Integer(maximumSupply.MinorUnit));
```

Lib9c/CurrencyExtensions.cs
tkiapril: If there is `maximumSupplyMajor`, there also must be `maximumSupplyMinor` as well. If one of them are missing it's a malformed serialization. Maybe should I raise an exception if one of them are missing?

moreal: I'm sorry for confusing about the `Currency` struct. As you said, it should raise an exception because it's a malformed serialization. And I expect it will raise an exception like `KeyNotFoundException` in the below line:

```
(Integer)serialized["maximumSupplyMinor"]
```

I think this confusion is from serialization implementation was leaked from out `Currency`.

I asked a question related with this problem on Discord Server. https://discord.com/channels/928926944937013338/970554437733073027/1020254911495159808

tkiapril: Added logic to reject malformed serializations.

moreal (approved):

dahlia (approved):

PR #1378, opened 2022-09-14
Edit workshop identifying block

tyrosine1153:
Edit identifying block for crafting or upgrading equipment. (from 50 blocks to 10)


boscohyun (changes requested):
Do not modify a value in the `GameConfig`.🥲

boscohyun (approved):

PR #1392, opened 2022-09-23
Turn on nullable reference checker

moreal:
This pull request enables nullable reference feature at `Lib9c` and `Lib9c.Tests` project and disables the feature for every files of the projects. It's to fix build errors in step by step.

## 🤔 What's null reference checker

Since C# 8.0, it was introduced. It separates `nonnull` type and `nullable` type. For example, every type was nullable. When you wrote `string name` line, you can assign `null` to the `name` variable like `string name = null`. But it should be `string? name = null` with `?` operator with the feature. It means you can believe the `name` variable in `string name` line, cannot have `null`. And you don't have to check `name is null` every time with the feature.

- https://learn.microsoft.com/en-us/dotnet/csharp/nullable-references

## 🙀 There are too many files to check

Yes, so I suggest to use replacement feature of editors. Before this pull request, this repository may not have any `#nullable disable` lines. You can replace `#nullable disable` with empty string and check difference with the development branch.

### DIFF
--- a/.Lib9c.Tests/Action/AccountStateDeltaExtensionsTest.cs
+++ b/.Lib9c.Tests/Action/AccountStateDeltaExtensionsTest.cs
@@ -1,3 +1,4 @@
+#nullable disable
 namespace Lib9c.Tests.Action
 {
     using System;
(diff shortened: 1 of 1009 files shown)


moreal:
I'll close this pull request and open other pull request using `#nullable enable` pragrma.

Issue #1406, opened 2022-09-29
Make series actions implement common interface


moreal:
Make actions of the same series are implemented so that the series can be confirmed by implementing a common interface.

For instance:

- `ITransferAsset`
  -  `TransferAsset0: ITransferAsset`
  -  `TransferAsset2: ITransferAsset`
  -  `TransferAsset: ITransferAsset`

PR #1407, opened 2022-09-29
Make `transfer_asset*` actions implement common interface

moreal:
This pull request just do only `transfer_asset*` actions implement common interface, `ITransferAsset`. Fortunately. their variables organizations are all same so it was easy. It is related with #1406.

I requested a review for @ipdae because he made `transfer_asset3` and I want to request a check about it.

### DIFF
--- a/Lib9c/Action/ITransferAsset.cs
+++ b/Lib9c/Action/ITransferAsset.cs
@@ -1,0 +1,14 @@
+#nullable enable
+using Libplanet;
+using Libplanet.Assets;
+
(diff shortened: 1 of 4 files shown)


boscohyun (approved):

moreal:
Though it doesn't receive a review from ipdae[^1], I'll merge this pull request when the build passed because he's busy now and I checked the difference between `transfer_asset2` and `transfer_asset3` manually with `diff` command.

 [^1]: I wanted to avoid sending him mention notification.

PR #1420, opened 2022-10-06
Update debug log for `InvalidBlockStateRootHashException` debugging

area363:
In an effort to better analyze potential `InvalidBlockStateRootHashException` error in future releases, this PR changes the previous `Verbose` level logs of `Lib9c` actions to `Debug` level. This change only applies to the start and end of each action so that it will be easier to assess which action failed when an `InvalidBlockStateRootHashException` occurs.


area363:
I've rebased it to the current development branch. PTAL! 🙏 @ipdae @sonohoshi @hskim1028

ipdae (approved):

PR #1434, opened 2022-10-17
Apply rune stats

Unengine:
Ownership check will be implemented soon.

Version up actions
- `HackAndSlash`
- `MimisbrunnrBattle`
- `EventDungeonBattle`
- `Raid`

Not implemented for now
- `BattleArena`


ipdae (commented):
IMO, it would be better to check to Execute existing action test without RuneSheet.

Lib9c/Action/Raid.cs
```diff
@@ -18,17 +18,18 @@
...
-    [ActionType("raid2")]
+    [ActionType("raid3")]
     public class Raid : GameAction
     {
         public const long RequiredInterval = 5L;
         public Address AvatarAddress;
         public List<Guid> EquipmentIds;
         public List<Guid> CostumeIds;
         public List<Guid> FoodIds;
+        public List<int> Runes;
```
ipdae, line 32: missing this change in plain value.

Lib9c/Battle/RaidBoss.cs
```diff
@@ -11,9 +11,6 @@ namespace Nekoyume.Model
 {
     public class RaidBoss : Enemy
     {
-        [NonSerialized]
-        public RaidSimulator RaidSimulator;
-
```
ipdae, line 16: is this correct working without simulator?

Lib9c/Battle/RaidSimulator.cs
```diff
@@ -19,7 +19,6 @@ public class RaidSimulator : Simulator
         public int DamageDealt { get; private set; }
         public List<FungibleAssetValue> AssetReward { get; private set; } = new List<FungibleAssetValue>();
         public override IEnumerable<ItemBase> Reward => new List<ItemBase>();
-        private const int TurnLimit = 150;
```
ipdae, line 22: missing copy params in `RaidSimulatorV1`

Lib9c/Model/Character/ArenaCharacter.cs
```diff
@@ -71,7 +71,7 @@ public int CurrentHP
         public ArenaCharacter(
             ArenaSimulator simulator,
             ArenaPlayerDigest digest,
-            ArenaSimulatorSheets sheets,
```
ipdae, line 74: hard-forked action doesn't need sheets anymore?

Lib9c/Battle/RaidBoss.cs
Unengine: This field is initialized but not used.

Lib9c/Model/Character/ArenaCharacter.cs
Unengine: This sheet is renamed to `ArenaSimulatorSheetsV1`.

Lib9c/Battle/RaidSimulator.cs
Unengine: This field is not used so I removed it.

ipdae (approved):

PR #1438, opened 2022-10-18
Introduce RuneEnhancement

hskim881028:
Introduce RuneEnhancement


Lib9c/Action/RuneEnhancement.cs
```diff
@@ -0,0 +1,155 @@
...
+            FungibleAssetValue ncg,
+            FungibleAssetValue crystal,
+            FungibleAssetValue rune,
+            Currency ncgCurrency,
+            Currency crystalCurrency,
+            Currency runeCurrency,
+            RuneCostSheet.RuneCostData cost,
+            IRandom random,
+            bool once,
+            out int tryCount)
```
ipdae, line 124: it can be reusable in client side when action rendered. how about seperate external class like `RuneHelper`

Lib9c/Model/EnumType/Grade.cs
```diff
@@ -0,0 +1,11 @@
+namespace Nekoyume.Model.EnumType
+{
+    public enum Grade
+    {
+        Normal = 1,
+        Rare = 2,
+        Epic = 3,
+        Unique = 4,
+        Legendary = 4,
```
ipdae, line 9: it seems duplicate value. is this correct?

PR #1450, opened 2022-10-26
Introduce RuneEnhancement

hskim881028:
Introduce RuneEnhancement


Lib9c/Helper/RuneHelper.cs
```diff
@@ -92,5 +93,46 @@ IRandom random
...
+                if (ncg < ncgCost || crystal < crystalCost || rune < runeCost)
+                {
+                    tryCount--;
+                    if (tryCount == 0)
+                    {
+                        throw new NotEnoughFungibleAssetValueException($"{nameof(RuneEnhancement)}" +
+                            $"[ncg:{ncg} < {ncgCost}] [crystal:{crystal} < {crystalCost}] [rune:{rune} < {runeCost}]");
+                    }
+                    return false;
+                }
```
ipdae, line 131: it would be better check balance first before increase tryCount

ipdae (approved):

PR #1453, opened 2022-10-27
Introduce rune slot content

hskim881028:
- State
   - **RuneSlotState**
   - **RuneSlot**
   
- Action
   - **EquipRune**

- Data Transfer Object
   - **RuneSlotInfo**
   
- Enum
   - **RuneSlotType**
   - **BattleType**

### DIFF
--- a/Lib9c/Model/State/RuneSlotState.cs
+++ b/Lib9c/Model/State/RuneSlotState.cs
@@ -6,6 +6,7 @@ using Libplanet;
 using Nekoyume.Action;
 using Nekoyume.Model.EnumType;
 using Nekoyume.Model.Rune;
+using Nekoyume.TableData;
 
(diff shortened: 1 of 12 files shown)


Lib9c/Model/State/RuneSlotState.cs
```diff
@@ -0,0 +1,136 @@
...
+                if (runeInfo is null)
+                {
+                    slot.Unequip();
+                }
+                else
+                {
+                    if (IsUsableSlot(runeStates, runeListSheet, slot, runeInfo, out var runeState))
+                    {
+                        slot.Equip(runeState);
+                    }
```
Unengine, line 64: Is it okay to do nothing on opposite condition here?

ipdae (approved):

Unengine (approved):

PR #1458, opened 2022-10-28
Add rune reward in stake reward

ipdae:
- fix TC
- hard-fork new action `ClaimStakeReward3`
- update `ClaimStakeRewardSheet` data & add type for rune reward
- obsolete & available claim_stake_reward actions


Lib9c/Action/ActionUnAvailableException.cs
```diff
@@ -0,0 +1,21 @@
+using System;
+using System.Runtime.Serialization;
+
+namespace Nekoyume.Action
+{
+    [Serializable]
+    public class ActionUnAvailableException : InvalidOperationException
```
boscohyun, line 7: How about use "Un`A`valiable" as lower case like below?

```suggestion
    public class ActionUnavailableException : InvalidOperationException
```

Lib9c/Action/ClaimStakeReward.cs
```diff
@@ -88,7 +89,7 @@ public override IAccountStateDelta Execute(IActionContext context)
             foreach (var reward in rewards)
             {
                 var (quantity, _) = stakedAmount.DivRem(currency * reward.Rate);
-                if (quantity < 1)
+                if (quantity < 1 || reward.Type != StakeRegularRewardSheet.StakeRewardType.Item)
```
boscohyun, line 92: Does not this cause consistency issues?

Lib9c/Action/ClaimStakeReward3.cs
```diff
@@ -0,0 +1,169 @@
...
+
+namespace Nekoyume.Action
+{
+    /// <summary>
+    /// Hard forked at https://github.com/planetarium/lib9c/pull/1371
+    /// </summary>
+    [ActionType(ActionTypeText)]
+    public class ClaimStakeReward3 : GameAction
+    {
+        public const long HardForkIndex = 5_599_601L;
```
boscohyun, line 21: How do you think renaming this something like `ActionAvailableIndex`? This pull request contains `ActionUnavaliableException` so it can be more readable.

.Lib9c.Tests/Action/ClaimStakeRewardTest.cs
```diff
@@ -108,24 +108,57 @@ public void Serialization()
...
-        public void Execute_With_Old_AvatarState_Success()
+        public void Execute_Throw_ActionObsoletedException()
         {
-            Execute(_avatarAddressForBackwardCompatibility);
+            var action = new ClaimStakeReward(_avatarAddress);
+            Assert.Throws<ActionObsoletedException>(() => action.Execute(new ActionContext
+            {
+                PreviousStates = _initialState,
+                Signer = _signerAddress,
+                BlockIndex = ClaimStakeReward3.HardForkIndex + 1,
```
boscohyun, line 135: If the `ClaimStakeReward3` is available in `ClaimStakeReward3.HardForkIndex`, and the `ClaimStakeReward`("xxx2") should throw an exception in `ClaimStakeReward3.HardForkIndex`. How you think about this?

```suggestion
                BlockIndex = ClaimStakeReward3.HardForkIndex,
```

hskim881028 (changes requested):
https://github.com/planetarium/lib9c/blob/d54635059b478ed357b723551e68c48b41e670c5/Lib9c/Action/ClaimStakeReward.cs#L19
I think.. It should be version up
`private const string ActionTypeText = "claim_stake_reward4";`

Lib9c/Action/ClaimStakeReward.cs
```diff
@@ -37,6 +37,7 @@ public override IAccountStateDelta Execute(IActionContext context)
             }
 
             var states = context.PreviousStates;
+            CheckObsolete(ClaimStakeReward3.HardForkIndex, context);
```
boscohyun, line 40: If use `ClaimStakeReward3.HardForkIndex` as is, `ClaimStakeReward3` and `ClaimStakeReward`("xxx2") available in the block which index is `ClaimStakeReward3.HardForkIndex`.

```suggestion
            CheckObsolete(ClaimStakeReward3.HardForkIndex - 1, context);
```

Unengine (approved):

Lib9c/Action/ClaimStakeReward.cs
ipdae: yes, type default value is item.

U-lis (approved):

boscohyun (approved):

PR #1464, opened 2022-11-01
Hardfork BattleArena to "battle_arena6"

sonohoshi:
- Add state about purchased ticket count during interval
  - `ArenaInformation.Address.Derive(BattleArena.PurchasedCountKey): Integer`
- Change signature of `ArenaInformation.BuyTicket()`
  - change type of argument to `long maxCount` from `ArenaSheet.RoundData roundData`
- Add Exception `ExceedTicketPurchaseLimitDuringIntervalException`


Lib9c/Action/BattleArena.cs
```diff
@@ -248,9 +249,17 @@ public override IAccountStateDelta Execute(IActionContext context)
...
+            if (!states.TryGetState(purchasedCountAddr, out Integer purchasedCountDuringInterval))
+            {
+                purchasedCountDuringInterval = 0;
+            }
+
             if (arenaInformation.TicketResetCount < currentTicketResetCount)
             {
                 arenaInformation.ResetTicket(currentTicketResetCount);
+                purchasedCountDuringInterval = 0;
+                states = states.SetState(purchasedCountAddr, purchasedCountDuringInterval);
```
ipdae, line 262: Isn't interval should be keeping count throughout the season? In this case, it seems that the interval is always reset to 0 every cycle.

.Lib9c.Tests/Action/ArenahelperTest.cs
```diff
@@ -129,10 +129,10 @@ public void ExecuteGetTicketPrice()
                     }
 
                     var arenaInformation = new ArenaInformation(_avatar1Address, roundData.ChampionshipId, roundData.Round);
-                    var max = ArenaHelper.GetMaxPurchasedTicketCount(roundData);
+                    var max = roundData.MaxPurchaseCount;
                     for (var i = 0; i < max; i++)
                     {
-                        arenaInformation.BuyTicket(roundData);
+                        arenaInformation.BuyTicket(roundData.MaxPurchaseCount);
```
ipdae, line 135: is this correct change? why not use ArenaHelper?

sonohoshi: I changed the role of `ArenaHelper.GetMaxPurchasedTicketCount` to `roundData.MaxPurchaseCount`. The intention is correct.

Lib9c/Policy/BlockPolicySource.cs
```diff
@@ -84,15 +84,16 @@ public partial class BlockPolicySource
...
-        // Target time(KST): 2022. 10. 12 오전 3:42:19
         public const long V100300ObsoleteIndex = 5_150_000;
 
         public const long V100310ObsoleteIndex = 5_200_400;
 
+        // NOTE:
+        // Current time(KST): 2022. 10. 28 16:26:59, 5_276_632
+        // Target release time(KST): 2022. 11. 09, 5_363_032
+        // Target obsolete time(KST): 2022. 12. 08, 5_579_032
+        public const long V100320ObsoleteIndex = 5_579_032;
```
ipdae, line 95: this index is not match any season start index. is this correct?

sonohoshi: Is it better to unify with the season start index?

Lib9c/Action/BattleArena.cs
sonohoshi: The intention is correct!

.Lib9c.Tests/Action/ArenahelperTest.cs
ipdae: it feels strange to test the changed method in this test class. because there is no required `ArenaHelper` anymore, i think this test case freeze with deprecated model for avoid `InvalidBlockStateRootHashException` and create new test in `BattleArenaTest`

Lib9c/Policy/BlockPolicySource.cs
sonohoshi: I reflected it. thank you!

.Lib9c.Tests/Action/ArenahelperTest.cs
sonohoshi: ok I understand. I will make new unit test.

tyrosine1153 (approved):

ipdae (approved):

PR #1472, opened 2022-11-03
Merge `RuneSkillSheet` into `RuneStatSheet`.

Unengine:
SSIA


Lib9c/Model/Character/Player.cs
```diff
@@ -558,18 +557,18 @@ public void SetCostumeStat(CostumeStatSheet costumeStatSheet)
                 Stats.AddOption(statModifiers);
                 Stats.EqualizeCurrentHPWithHP();
 
-                if (!runeSkillSheet.TryGetValue(runeId, out var runeSkillRow) ||
-                    !runeSkillRow.LevelSkillMap.TryGetValue(level, out var skillInfo) ||
+                if (!statRow.LevelStatsMap.TryGetValue(level, out var skillInfo) ||
+                    skillInfo.SkillId == default ||
```
ipdae, line 561: this change can be occur `InvalidblockStateRootHashException`

Unengine: This codes are in `SetRunes()`. Previous actions don't call them so I think it's ok.

hskim881028 (approved):

PR #1475, opened 2022-11-07
Set UpdateSell.updateSellInfos capacity

ipdae:
- set updateSellInfos capacity

### DIFF
--- a/Lib9c/Action/UpdateSell.cs
+++ b/Lib9c/Action/UpdateSell.cs
@@ -26,13 +26,14 @@ namespace Nekoyume.Action
     [ActionType("update_sell4")]
     public class UpdateSell : GameAction
     {
+        private const int UpdateCapacity = 100;
         public Address sellerAvatarAddress;
         public IEnumerable<UpdateSellInfo> updateSellInfos
         {
             get => _updateSellInfos;
             set
             {
-                if (value.Count() > 100)
+                if (value.Count() > UpdateCapacity)
                 {
                     throw new ArgumentOutOfRangeException($"{nameof(value)} must be less than or equal 100.");
                 }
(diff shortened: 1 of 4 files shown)


Lib9c/Action/UpdateSell.cs
```diff
@@ -23,11 +23,25 @@ namespace Nekoyume.Action
...
-        public IEnumerable<UpdateSellInfo> updateSellInfos;
+        public IEnumerable<UpdateSellInfo> updateSellInfos
+        {
+            get => _updateSellInfos;
+            set
+            {
+                if (value.Count() > 100)
+                {
+                    throw new ArgumentOutOfRangeException($"{nameof(value)} must be less than or equal 100.");
+                }
```
longfin, line 38: I think that we need to check in the `.Execute()` too because users still can send `update_sell4` action with 100+ arguments through editing the payload by hand.

Lib9c/Action/UpdateSell.cs
```diff
@@ -23,11 +23,25 @@ namespace Nekoyume.Action
     /// Updated at https://github.com/planetarium/lib9c/pull/1022
     /// </summary>
     [Serializable]
-    [ActionType("update_sell3")]
+    [ActionType("update_sell4")]
```
longfin, line 26: Obsoletion may be needed. 🤔

ipdae: obsolete index set on update_sell3 action

Lib9c/Action/UpdateSell.cs
ipdae: it will throw exception in LoadPlainValue, but I'll fix it.

sonohoshi (approved):

PR #1498, opened 2022-11-11
Introduce `BattleGrandFinale` action

sonohoshi:
Add: `BattleGrandFinale` action
```cs
public class BattleGrandFinale : GameAction
{
    public Address myAvatarAddress;
    public Address enemyAvatarAddress;
    public int grandFinaleId;
    public List<Guid> costumes;
    public List<Guid> equipments;

    public ArenaPlayerDigest ExtraMyArenaPlayerDigest;
    public ArenaPlayerDigest ExtraEnemyArenaPlayerDigest;
}
```
Add: `GrandFinaleParticipantsSheet`, `GrandFinaleScheduleSheet` sheet
```cs
public class GrandFinaleParticipantsSheet : Sheet<int, GrandFinaleParticipantsSheet.Row>
{
    [Serializable]
    public class Row : SheetRow<int>
    {
        public override int Key => GrandFinaleId;

        public int GrandFinaleId { get; private set; }

        public List<Address> Participants { get; private set; }

        public override void Set(IReadOnlyList<string> fields)
        {
            GrandFinaleId = ParseInt(fields[0]);
            Participants = new List<Address> {new Address(fields[1])};
        }
    }
}

public class GrandFinaleScheduleSheet : Sheet<int, GrandFinaleScheduleSheet.Row>
{
    [Serializable]
    public class Row : SheetRow<int>
    {
        public override int Key => Id;

        public int Id { get; private set; }

        public long StartBlockIndex { get; private set; }

        public long EndBlockIndex { get; private set; }

        public bool IsOpened(long blockIndex);
    }
}
```
Add: `GrandFinaleInformation` model
```cs
public class GrandFinaleInformation : IState
{
    public static Address DeriveAddress(Address avatarAddress, int grandFinaleId) =>
        avatarAddress.Derive($"arena_information_{grandFinaleId}");

    public Address Address;
    public int Ticket { get; private set; }
    private Dictionary<Address, bool> BattleRecordDictionary { get; }
}
```
Add: `AlreadyFoughtAvatarException` Exception


ipdae (commented):
The cycle of the grand finale is one year, and it is expected that the rules will change slightly each time. it would be better to manage a class for grand finale season 1 by collecting rule-related data (address list, score, period, etc.).

Lib9c/Model/GrandFinale/GrandFinaleInformation.cs
```diff
@@ -0,0 +1,70 @@
...
+{
+    using System.Linq;
+    public class GrandFinaleInformation : IState
+    {
+        public static Address DeriveAddress(Address avatarAddress, int grandFinaleId) =>
+            avatarAddress.Derive($"grand_finale_information_{grandFinaleId}");
+
+        public Address Address;
+        public int Ticket { get; private set; }
+        private Dictionary<Address, bool> BattleRecordDictionary { get; }
```
ipdae, line 18: it can be simplify contains address list.

Lib9c/Model/GrandFinale/GrandFinaleInformation.cs
```diff
@@ -0,0 +1,57 @@
...
+
+        public IValue Serialize()
+        {
+            var battleRecordDict = new Dictionary(BattleRecordDictionary
+                .OrderBy(pair => pair.Key)
+                .Select(pair =>
+                    new KeyValuePair<IKey, IValue>(
+                        (IKey)pair.Key.Serialize(),
+                        pair.Value.Serialize()))
+            );
```
ipdae, line 40: it would be better serialize list.

Lib9c/Action/BattleGrandFinale.cs
```diff
@@ -0,0 +1,230 @@
...
+            var enemyArenaAvatarStateAdr = ArenaAvatarState.DeriveAddress(enemyAvatarAddress);
+            if (!states.TryGetArenaAvatarState(
+                    enemyArenaAvatarStateAdr,
+                    out var enemyArenaAvatarState))
+            {
+                throw new ArenaAvatarStateNotFoundException(
+                    $"[{nameof(BattleArena)}] enemy avatar address : {enemyAvatarAddress}");
+            }
+
+            var scoreAddress = myAvatarAddress.Derive(string.Format(CultureInfo.InvariantCulture, ScoreDeriveKey, grandFinaleId));
```
ipdae, line 172: Is there a case only the used score without the arena information?

sonohoshi: Come to think of it, it's right to include it in GrandFinaleInformation. I'll reflect it.

ipdae (approved):

PR #1522, opened 2022-11-17
Bump libplanet

moreal:
SSIA.

This pull request's main purpose is to make the `IActionTypeLoader` feature available.

See also https://github.com/planetarium/libplanet/pull/2539.


greymistcube:
Libplanet release cycle is not strongly tied to 9c release schedule. Has it been decided this should be included in the next 9c release? If so, then it must be communicated with @planetarium/libplanet about the libplanet 0.45 release schedule. 😶

moreal:
> Libplanet release cycle is not strongly tied to 9c release schedule. Has it been decided this should be included in the next 9c release? If so, then it must be communicated with @planetarium/libplanet about the libplanet 0.45 release schedule. 😶

No, it may not be included the next time. I don't think it is not required feature yet so it can be resolved in the next Libplanet release bump.

PR #1530, opened 2022-11-22
Introduce `IClaimStakeReward` interface and apply to actions

boscohyun:
SSIA


U-lis (approved):
I think that is good to include `AvatarAddress` to interface to enforce all implements to receive `AvatarAddress` :thinking:  but it's ok as-is to me.

moreal (approved):

PR #1533, opened 2022-11-22
Introduce `ClaimStakeRewardFactory`

boscohyun:
The action factories provides two methods to create an action.

1. `CreateByBlockIndex()`
2. `CreateByVersion()`

This pull request is based on #1530.

### DIFF
--- a/Lib9c/Action/Factory/ClaimStakeRewardFactory.cs
+++ b/Lib9c/Action/Factory/ClaimStakeRewardFactory.cs
@@ -1,0 +1,37 @@
+using System;
+using Libplanet;
+
+namespace Nekoyume.Action.Factory
+{
+    public static class ClaimStakeRewardFactory
+    {
+        // NOTE: This method does not return a type of `ClaimStakeReward1`.
+        //       Because it is not obsoleted yet.
+        public static IClaimStakeReward CreateByBlockIndex(
+            long blockIndex,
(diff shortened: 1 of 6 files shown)


Lib9c/Action/Factory/ClaimStakeRewardFactory.cs
```diff
@@ -0,0 +1,37 @@
+using System;
+using Libplanet;
+
+namespace Nekoyume.Action.Factory
+{
+    public static class ClaimStakeRewardFactory
+    {
+        // NOTE: This method does not return a type of `ClaimStakeReward1`.
+        //       Because it is not obsoleted yet.
+        public static IClaimStakeReward CreateByBlockIndex(
```
U-lis, line 10: I think the method name `ClaimStakeReward` is the best one for both functions because it's easy to use with the same name and can shorten code: `ClaimStakeRewardFactory.CreateByBlockIndex` > `ClaimStakeReward`

boscohyun: @U-lis I also prefer shorter code. However, you don't want to miss out on necessary information.
Can you explain the advantages of making method names shorter than lengthy code but explicitly using them?

U-lis (approved):
Despite of method naming issue, the functionality itself looks good to me.

PR #1534, opened 2022-11-22
Introduce TransferAssets

ipdae:
Introduce `TransferAssets` action for bulk transfer asset.

### DIFF
--- a/Lib9c/Action/TransferAssets.cs
+++ b/Lib9c/Action/TransferAssets.cs
@@ -27,10 +27,10 @@ namespace Nekoyume.Action
         {
         }
 
-        public TransferAssets(Address sender, Dictionary<Address, FungibleAssetValue> map, string memo = null)
+        public TransferAssets(Address sender, List<(Address, FungibleAssetValue)> recipients, string memo = null)
         {
             Sender = sender;
-            Map = map;
+            Recipients = recipients;
 
             CheckMemoLength(memo);
             Memo = memo;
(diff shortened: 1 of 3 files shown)


Lib9c/Action/TransferAssets.cs
```diff
@@ -0,0 +1,170 @@
...
+        protected TransferAssets(SerializationInfo info, StreamingContext context)
+        {
+            var rawBytes = (byte[])info.GetValue("serialized", typeof(byte[]));
+            Dictionary pv = (Dictionary) new Codec().Decode(rawBytes);
+
+            LoadPlainValue(pv);
+        }
+
+        public Address Sender { get; private set; }
+        public Dictionary<Address, FungibleAssetValue> Map { get; private set; }
```
longfin, line 48:
```suggestion
        public Dictionary<Address, FungibleAssetValue> Recipients { get; private set; }
```

longfin, line 48: If we want to keep this `Dicionary<>`'s order, how about using `List<Tuple<Address, FungibleAssetValue>>` instead?

longfin: or, preserving the ordered internal field seems helpful to avoid re-ordering at least.

Lib9c/Action/TransferAssets.cs
```diff
@@ -0,0 +1,170 @@
...
+                throw new InvalidTransferRecipientException(Sender, recipient);
+            }
+
+            Address recipientAddress = recipient.Derive(ActivationKey.DeriveKey);
+
+            // Check new type of activation first.
+            // If result of GetState is not null, it is assumed that it has been activated.
+            if (
+                state.GetState(recipientAddress) is null &&
+                state.GetState(Addresses.ActivatedAccount) is Dictionary asDict &&
```
longfin, line 141: It seems that this action will try this `.GetState()` lookup per `Map` count, but they are all also the same. can we move it from this local method to reduce it?

longfin (approved):

PR #1542, opened 2022-11-28
pbft: bump libplanet

colibrishin:


limebell (approved):
Little worried about duplicated calls of `GenerateBlockCommit` but I guess it's okay because it is just a unit test.

PR #1559, opened 2022-12-02
Bump Libplanet to 0.45.0

longfin:
It consists of these cherry-picks

- 623fa547cfcd5475c66af0049c8a89c6bf22c531
- e36166ba088db48993943b6c8bd6f487d6c13410

### DIFF
--- a/Lib9c/Policy/MinBlockProtocolVersionPolicy.cs
+++ b/Lib9c/Policy/MinBlockProtocolVersionPolicy.cs
@@ -1,0 +1,26 @@
+namespace Nekoyume.BlockChain.Policy
+{
+    public sealed class MinBlockProtocolVersionPolicy : VariableSubPolicy<int>
+    {
+        private MinBlockProtocolVersionPolicy(int defaultValue)
+            : base(defaultValue)
+        {
+        }
+
+        private MinBlockProtocolVersionPolicy(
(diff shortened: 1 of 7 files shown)


Lib9c/Policy/MinBlockProtocolVersionPolicy.cs
```diff
@@ -0,0 +1,26 @@
...
+        }
+
+        public static IVariableSubPolicy<int> Default =>
+            new MinBlockProtocolVersionPolicy(0);
+
+        public static IVariableSubPolicy<int> Mainnet =>
+            Default
+                .Add(new SpannedSubPolicy<int>(
+                    startIndex: 4_000L,
+                    value: 4));
```
greymistcube, line 24: We need to set this to `Default`, otherwise the chain would stop. 😱

greymistcube (changes requested):

dahlia (approved):

Lib9c/Policy/MinBlockProtocolVersionPolicy.cs
greymistcube: Added 500dc116b1041704551e6ff830b3faabc7ea6590.

greymistcube (approved):

dahlia (approved):

PR #1568, opened 2022-12-07
Add dynamic validators

OnedgeLee:
### Changes
Bumped libplanet, and apply proper changes.

ValidatorAdmin : Public key of validator-admin have to be assigned on the policy, to apply `SetValidator` system action. Non-validator-admin cannot modify validator set.

### Remarks
Uploaded genesis block on local github branch, and this have to be moved to stable bucket.


.Lib9c.Tests/Action/RewardGoldTest.cs
```diff
@@ -484,7 +484,7 @@ public async Task Genesis_StateRootHash(bool mainnet)
             Block<PolymorphicAction<ActionBase>> genesis;
             if (mainnet)
             {
-                const string genesisBlockPath = "https://9c-test.s3.ap-northeast-2.amazonaws.com/genesis-block-9c-main";
+                const string genesisBlockPath = "https://github.com/OnedgeLee/lib9c/raw/pbft-dynamic_validator/genesis-block";
```
OnedgeLee, line 487: @limebell This have to be moved to proper bucket

limebell (approved):

PR #1585, opened 2022-12-14
Introduce test-only action: Faucet

U-lis:
- DO NOT RELEASE THIS FUNCTION TO PRODUCTION
- This action can mint ncg and/or crystal to target address.
- This action is only for test.


boscohyun (commented):
Can we remove the csv files that looks like same with Lib9c's? We can refer Lib9c's csv files with `TableSheetsImporter` and I suggest the way in my other [pr](https://github.com/planetarium/lib9c/pull/1606/files).

Lib9c/Model/Faucet/FaucetRuneInfo.cs
```diff
@@ -0,0 +1,42 @@
+using System;
+using Bencodex.Types;
+using Nekoyume.Model.State;
+
+namespace Nekoyume.Action
```
boscohyun, line 5:
```suggestion
namespace Nekoyume.Model.Faucet
```

Lib9c.DevExtensions/Action/FaucetRune.cs
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    protected override IImmutableDictionary<string, IValue> PlainValueInternal =>
+        new Dictionary<string, IValue>
+        {
+            ["avatarAddress"] = AvatarAddress.Serialize(),
+            ["faucetRuneInfos"] = FaucetRuneInfos
+                .OrderBy(x => x.RuneId)
+                .Select(x => x.Serialize())
+                .Serialize()
```
boscohyun, line 55: This operator help to avoid the indeterminism.

```suggestion
            ["faucetRuneInfos"] = FaucetRuneInfos
                .OrderBy(x => x.RuneId)
                .ThenBy(x => x.Amount)
                .Select(x => x.Serialize())
                .Serialize()
```

U-lis: It was initially assumed that one RuneId appeared only once. To cover multiple cases, secondary sort might be helpful.
Thanks!

boscohyun (approved):

PR #1592, opened 2022-12-19
Use linux 8cores GitHub Actions larger runner

moreal:
Since the `main` workflow takes about 20~27 minutes, I suggest using high-core GitHub Actions Runners.

There are GitHub Actions larger runners feature and this pull request makes the workflow use them.

## Reviewers

Because GitHub Actions larger runner feature occurs cost issue, I request reviews to @planetarium/devops team.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -9,7 +9,8 @@ on:
 
 jobs:
   build-and-test:
-    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


Atralupus:
I guess you would better to use the strategy github action

```
dotnet test --filter "FullyQualifiedName=.Lib9c.Tests.Action"
```

PR #1611, opened 2022-12-27
Mimic Lib9c action structure

U-lis:
- To test Lib9c and DevExtensions actions seperately, we need to mimic lib9c structure into DevExtensions.
- Add `TestActionBase` and `TestGameAction` from `ActionBase` and `GameAction`.


U-lis:
We cannot dynamically load ActionTypes from now on.
So this change does not make any sense.

PR #1612, opened 2022-12-27
Fix `ToLong()` extension method

### LINKED ISSUE #1609: Parse serialized long value by parser of integer
from:
- https://github.com/planetarium/lib9c/blob/main/Lib9c/Model/State/StateExtensions.cs#L123-L124
- https://github.com/planetarium/lib9c/blob/main/Lib9c/Model/State/StateExtensions.cs#L126-L127

boscohyun:
Resolve #1609

### DIFF
--- a/.Lib9c.Tests/Model/State/StateExtensionsTest.cs
+++ b/.Lib9c.Tests/Model/State/StateExtensionsTest.cs
@@ -1,14 +1,37 @@
 namespace Lib9c.Tests.Model.State
 {
-    using System;
     using System.Numerics;
-    using Nekoyume.Battle;
-    using Nekoyume.Model.BattleStatus;
     using Nekoyume.Model.State;
     using Xunit;
 
     public class StateExtensionsTest
     {
+        [Theory]
+        [InlineData(long.MinValue)]
+        [InlineData(0L)]
+        [InlineData(long.MaxValue)]
+        public void Long(long value)
+        {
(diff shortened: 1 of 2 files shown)


moreal (approved):
I worried that failed transactions by the overflow already exist. When I looked for the usage of the `StateExtensions.ToLong` and `StateExtensions.GetLong`, they were used for de/serialization of `block index`, `exp`. I expect they will not over exceed `int`'s range so I'll approve. If we find a related issue, we should rollback.

boscohyun:
@moreal Thanks for your thoughtful review. I talked about that points with @longfin and came to the same conclusion as you. 

Here is some reasons for this pull request can be merged.

1. Can expect there is no overflows yet.
2. The lib9c cannot evaluate the mainnet blockchain from the genesis block already. 
3. We are preparing the pluggable lib9c.(free from backward compatibility)

moreal:
I also think this pull request can be merged. But I hope we shouldn't say `The lib9c cannot evaluate the mainnet blockchain from the genesis block already.` as the reason to merge this pull request. It is the fault and should be fixed sometime. If we will not care about it, I don't think we cannot say Nine Chronicles is a blockchain game.

boscohyun:
@moreal You said something very important. I agree with you and will keep it in mind.👍

boscohyun:
@moreal I made #1622. Thanks!

PR #1613, opened 2022-12-27
Fix exception message

boscohyun:
When you play the `HackAndSlash` with world(1) and stage(7) and if your `AvatarState` has cleared the world(1) and stage(5) yet. The `InvalidStageException` will be occurred with some message.
This pull request fix the message. And separates the conditions for readability.

Before
> conditions: !world.IsStageCleared && stageId != world.StageBegin || world.IsStageCleared && stageId - 1 > world.StageClearedId
> format: {addressesHex}Aborted as the stage ({worldId}/{stageId}) is not cleared; cleared stage: {world.StageClearedId}
> output: Aborted as the stage (1/7) is not cleared; cleared stage: 5

The "(1/7)" part is confusing.

After

> condition: !world.IsStageCleared && stageId != world.StageBegin
> format: {addressesHex}Aborted as the stage ({worldId}/{stageId - 1}) is not cleared; clear the stage first: ({world.Id}/{world.StageBegin})
> output: Aborted as the stage (1/6) is not cleared; clear the stage first: (1/1)

> condition: world.IsStageCleared && stageId - 1 > world.StageClearedId
> format: {addressesHex}Aborted as the stage ({worldId}/{stageId - 1}) is not cleared; cleared stage: ({world.Id}/{world.StageClearedId})
> output: Aborted as the stage (1/6) is not cleared; cleared stage: (1/5)

### DIFF
--- a/Lib9c/Helper/Validator.cs
+++ b/Lib9c/Helper/Validator.cs
@@ -74,7 +74,7 @@ namespace Nekoyume.Helper
             {
                 throw new InvalidStageException(
                     $"{addressesHex}Aborted as the stage ({worldId}/{stageId - 1}) is not cleared; " +
-                    $"clear the stage first: ({world.Id}/{world.StageBegin})"
+                    $"clear the stage ({world.Id}/{world.StageBegin}) first"
                 );
             }
 
@@ -82,7 +82,8 @@ namespace Nekoyume.Helper
             {
                 throw new InvalidStageException(
                     $"{addressesHex}Aborted as the stage ({worldId}/{stageId - 1}) is not cleared; " +
-                    $"cleared stage: ({world.Id}/{world.StageClearedId})"
+                    $"cleared stage is ({world.Id}/{world.StageClearedId}), so you can play stage " +
+                    $"({world.Id}/{world.StageClearedId + 1})"
                 );
             }
 
--- a/Lib9c/Helper/Validator.cs
+++ b/Lib9c/Helper/Validator.cs
@@ -70,12 +70,19 @@ namespace Nekoyume.Helper
                 worldInformation.UpdateWorld(worldRow);
             }
 
-            if (world.IsStageCleared && stageId - 1 > world.StageClearedId ||
-                !world.IsStageCleared && stageId != world.StageBegin)
+            if (!world.IsStageCleared && stageId != world.StageBegin)
             {
                 throw new InvalidStageException(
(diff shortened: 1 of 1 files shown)


Lib9c/Helper/Validator.cs
```diff
@@ -70,12 +70,19 @@ public static class Validator
...
             }
 
-            if (world.IsStageCleared && stageId > world.StageClearedId + 1 ||
-                !world.IsStageCleared && stageId != world.StageBegin)
+            if (!world.IsStageCleared && stageId != world.StageBegin)
             {
                 throw new InvalidStageException(
-                    $"{addressesHex}Aborted as the stage ({worldId}/{stageId}) is not cleared; " +
-                    $"cleared stage: {world.StageClearedId}"
+                    $"{addressesHex}Aborted as the stage ({worldId}/{stageId - 1}) is not cleared; " +
```
moreal, line 76: It is just a question. Maybe I guess it is because I have a gap in lib9c domain knowledge. Why does it subtract one from `stageId`?

When it tried to clear `3/2` (`world/stage`) but the beginning stage was `3/1`, it should show stage id `2` without any modification. Doesn't it?

boscohyun: In `HackAndSlash`, Player should cleared 1/6 before play 1/7. So the `stageId` subtract one in that error message.

> When it tried to clear 3/2 (world/stage) but the beginning stage was 3/1, it should show stage id 2 without any modification. Doesn't it?

- When your avatar has been cleared 1/10,
- And if try to clear 1/20
- InvalidStageException occurred with message: Aborted as the stage (1/19) is not cleared; cleared stage: (1/10)

The "cleared stage: (1/10)" part is little bit confusing. This means "your avatar has been cleared stage (1/10), so try to clear (1/11) first"..

moreal: > The "cleared stage: (1/10)" part is little bit confusing. This means "your avatar has been cleared stage (1/10), so try to clear (1/11) first"..

I think the adjusted version seems easy to understand than before.

But anyway it is good if @planetarium/9c-dev feels comfortable when debugging.

moreal (approved):

Lib9c/Helper/Validator.cs
boscohyun: I'll adjust. Thanks!

PR #1619, opened 2022-12-29
Implement critical damage (CDMG)

Unengine:
CDMG is calculated as permyriad.
e.g. If CDMG is 10000, critical damage is increased by 100%. So critical will deal 250% of original damage. (150% + 100%p)


ipdae (commented):
skill damage calculations are becoming more complex each new buff added. Add a unit test to check if the damage calculation works properly when different types of buffs and debuffs are combined.

Lib9c/Model/Skill/AttackSkill.cs
```diff
@@ -61,7 +61,9 @@ protected AttackSkill(SkillSheet.Row skillRow, int power, int chance) : base(ski
                             isCritical = caster.IsCritical(isNormalAttack);
                             if (isCritical)
                             {
-                                damage = (int) (damage * CharacterBase.CriticalMultiplier);
+                                var critMultiplier =
+                                    Math.Max(1m, CharacterBase.CriticalMultiplier + (caster.CDMG / 10000m));
```
hskim881028, line 65:
```suggestion
                                    Math.Max(1.5m, CharacterBase.CriticalMultiplier + (caster.CDMG / 10000m));
```

hskim881028 (changes requested):

Lib9c/Model/Skill/AttackSkill.cs
Unengine: This is intended because crit damage can be decreased with debuffs.

hskim881028 (approved):

ipdae (approved):

PR #1640, opened 2023-01-09
Hot fix `Sell` and `UpdateSell`

boscohyun:
SSIA

### DIFF
--- a/Lib9c/Action/Sell.cs
+++ b/Lib9c/Action/Sell.cs
@@ -123,12 +123,18 @@ namespace Nekoyume.Action
             Log.Verbose("{AddressesHex}Sell IsStageCleared: {Elapsed}", addressesHex, sw.Elapsed);
             sw.Restart();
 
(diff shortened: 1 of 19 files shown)


Lib9c/Action/Sell.cs
```diff
@@ -19,7 +19,7 @@ namespace Nekoyume.Action
     /// Hard forked at https://github.com/planetarium/lib9c/pull/1376
```
sonohoshi, line 19:
```suggestion
    /// Hard forked at https://github.com/planetarium/lib9c/pull/1640
```

sonohoshi (approved):

ipdae (commented):
It would be good to sanity check for totalCount when the loop is finished.

boscohyun:
> It would be good to sanity check for totalCount when the loop is finished.

I applied!

sonohoshi (approved):

hskim881028 (approved):

PR #1654, opened 2023-01-11
Prepare for mobile build

longfin:


Lib9c/Action/GoldDistribution.cs
```diff
@@ -37,10 +37,24 @@ public string AddressString
         public static GoldDistribution[] LoadInDescendingEndBlockOrder(string csvPath)
         {
             GoldDistribution[] records;
-            using (var reader = new StreamReader(csvPath))
-            using (var csv = new CsvReader(reader, CultureInfo.InvariantCulture))
+            if(UnityEngine.Application.platform == UnityEngine.RuntimePlatform.Android)
```
longfin, line 40: I think that we need to use preprocessor instead of `UnityEngine`... 😢

PhangCY: Agree, we should avoid this kind of runtime if-else condition for platform-specific code.
Will check all codes, change this in our side.

longfin:
Replaced by #1714

PR #1686, opened 2023-01-30
Introduce PetEnhancement action

sonohoshi:
[펫 소환 및 강화 백엔드](https://app.asana.com/0/958521740385861/1203841632223461/f)

- Action "pet_enhancement" `(AvatarAddress:Address, PetId:int, TargetLevel:int)`
- Add PetHelper
  - `PetHelper.CalculateEnhancementCost(PetCostSheet, PetId, NowLevel, TargetLevel)`: return `(Ncg, SoulStone)`
- Check `PetEnhancement0Test` and `PetHelperTest`
- Improve unit tests for several sheets related to the pet system.


.Lib9c.Tests/Model/State/PetStateTest.cs
```diff
@@ -9,37 +9,56 @@
...
+            var serialized2 = deserialized.Serialize();
+            Assert.Equal(serialized, serialized2);
         }
 
-        [Fact]
-        public void LevelUp()
+        [Theory]
+        [InlineData(int.MinValue, false)]
+        [InlineData(0, false)]
+        [InlineData(int.MaxValue, true)]
```
U-lis, line 32: I think we should test within pet's level range, not int min/max.

.Lib9c.Tests/TableData/Pet/PetCostSheetTest.cs
```diff
@@ -1,58 +1,57 @@
...
         [Fact]
-        public void SetToSheet()
+        public void Set()
         {
-            const string content =
-                @"ID,_PET NAME,PetLevel,SoulStoneQuantity,NcgQuantity
+            const string content = @"pet_id,_pet_name,pet_level,soul_stone_quantity,ncg_quantity
 1,D:CC 블랙캣,1,10,0
-        ";
+1,D:CC 블랙캣,2,0,10";
```
U-lis, line 13: Can we use actual CSV file, not hard coded sample data?
It would be more helpful if CSV contents are changed.

.Lib9c.Tests/Helper/PetHelperTest.cs
```diff
@@ -0,0 +1,107 @@
...
+        public void CalculateEnhancementCost_Throw_ArgumentException()
+        {
+            var row = _petCostSheet.First!;
+            var petId = row.PetId;
+            Assert.Throws<ArgumentException>(() =>
+                PetHelper.CalculateEnhancementCost(
+                    _petCostSheet,
+                    -1,
+                    0,
+                    1));
```
U-lis, line 57: How about moving these args to `inlineData` of test?
I cannot get what `-1, 0, 1` means until look inside of `CalculateEnhancement` function.

Lib9c/Helper/PetHelper.cs
```diff
@@ -7,13 +9,41 @@ public static class PetHelper
         public static (int ncgQuantity, int soulStoneQuantity) CalculateEnhancementCost(
             PetCostSheet costSheet,
             int petId,
-            int nowLevel,
+            int currentLevel,
```
U-lis, line 12: This change is nice :+1:

Lib9c/Helper/PetHelper.cs
```diff
@@ -7,13 +9,41 @@ public static class PetHelper
...
+            if (targetLevel < 0)
+            {
+                throw new ArgumentException(
+                    $"{nameof(targetLevel)} must be greater than or equal to 0.");
+            }
+
+            if (!costSheet.TryGetValue(petId, out var row))
+            {
+                throw new SheetRowNotFoundException(nameof(PetCostSheet), petId);
+            }
```
U-lis, line 41: Thanks for handling edge cases.
Can you add one more case: `if (currentLevel >= targetLevel)` ?

Lib9c/Helper/PetHelper.cs
```diff
@@ -7,13 +9,41 @@ public static class PetHelper
...
+                    $"{nameof(petId)} must be greater than or equal to 0.");
+            }
+
+            if (currentLevel < 0)
+            {
+                throw new ArgumentException(
+                    $"{nameof(currentLevel)} must be greater than or equal to 0.");
+            }
+
+            if (targetLevel < 0)
```
U-lis, line 32: I think lower bound of `targetLevel` is 1, not 0. Can you check this out?

.Lib9c.Tests/Model/State/PetStateTest.cs
boscohyun: There is only one case where `PetState.Level` to be negative. That is the constructor with serialized argument. And that case does not happen unless someone intend to.
But spec is spec.😅 So I wrote unit test like this.

.Lib9c.Tests/TableData/Pet/PetCostSheetTest.cs
boscohyun: I guess the actual CSV file you said is that contained in `Lib9c/TableCSV/`.
These CSV files already tested when the `Lib9c.Tests.TableSheets` instantiated.

Currently all `XxxSheet.Set()` test just check couple of success cases. And I also think if there are failure cases it would be better.

U-lis (approved):

PR #1694, opened 2023-02-01
Remove check in Miner.cs

limebell:
Remove check that exclude action which has no any action.

### DIFF
--- a/Lib9c.Miner/Miner.cs
+++ b/Lib9c.Miner/Miner.cs
@@ -68,7 +68,7 @@ namespace Nekoyume.BlockChain
                         var nextBlockIndex = _chain.Tip.Header.Index + 1;
                         var types = actionTypeLoader.Load(new ActionTypeLoaderContext(nextBlockIndex));
 
-                        return tx.CustomActions?.All(ca =>
+                        return !(tx.SystemAction is null) || tx.CustomActions?.All(ca =>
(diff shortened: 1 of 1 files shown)


Lib9c.Miner/Miner.cs
```diff
@@ -61,21 +61,6 @@ public class Miner
...
                 };
 
-                if (_actionTypeLoader is { } actionTypeLoader)
-                {
-                    txValidators.Add(tx =>
-                    {
-                        var nextBlockIndex = _chain.Tip.Header.Index + 1;
-                        var types = actionTypeLoader.Load(new ActionTypeLoaderContext(nextBlockIndex));
-
-                        return tx.CustomActions?.All(ca =>
```
moreal, line 71: If it causes an issue when the `ITransaction.CustomActions` is `null`, how about checking `ITransaction.SystemAction` is not `null`?

LIke:
```csharp
return !(tx.SystemAction is null) || tx.CustomActions?.All(ca =>
```

OnedgeLee (approved):

Lib9c.Miner/Miner.cs
limebell: Amended!

moreal (approved):
LGTM 👍🏻

PR #1712, opened 2023-02-09
Craft action scenario tests

U-lis:
Add tests for craft actions to prepare new feature: pet

Works form now makes sure all the actions are working properly before pet comes in.
After this PR is merged, craft actions with pet will be tested.

- [x] CombinationEquipment
- [x] CombinationConsumable
- [x] EventEquipmentItemCrafts
- [x] EventMaterialItemCrafts
- [x] RapidCombination


Lib9c/Model/EnumType/CraftType.cs
```diff
@@ -0,0 +1,9 @@
+namespace Nekoyume.Model.EnumType
+{
+    public enum CraftType
+    {
+        Normal = 0,
+        Premium = 1,
+        Mimisbrunner = 2
```
Unengine, line 7:
```suggestion
        Mimisbrunnr = 2
```

.Lib9c.Tests/Action/Scenario/ItemCraftTest.cs
```diff
@@ -0,0 +1,405 @@
...
+            Assert.Equal(recipeList.Count, inventoryState.Items.Count);
+            foreach (var itemId in targetItemIdList)
+            {
+                // TEST: Created equipment should match with targetItemList
+                Assert.NotNull(inventoryState.Items.Where(e => e.item.Id == itemId));
+            }
+        }
+
+        [Theory]
+        [InlineData(1, new[] { 200000 })] // 맛없는 요리
```
Unengine, line 150: I think the test result of this inline data row is not helpful. Can you consider adding craftable recipe?

.Lib9c.Tests/Action/Scenario/ItemCraftTest.cs
```diff
@@ -0,0 +1,405 @@
...
+                _avatarAddr
+            );
+
+            // Setup requirements
+            var random = new TestRandom(randomSeed);
+            var recipeList = _tableSheets.ConsumableItemRecipeSheet.OrderedList.Where(
+                recipe => targetItemIdList.Contains(recipe.ResultConsumableItemId)
+            ).ToList();
+            var allMaterialList = new List<EquipmentItemSubRecipeSheet.MaterialInfo>();
+            foreach (var recipe in recipeList)
```
Unengine, line 166: I think count of the recipes should be checked. (Assert if count is non-zero or positive number)

.Lib9c.Tests/Action/Scenario/ItemCraftTest.cs
U-lis: > I think the test result of this inline data row is not helpful. Can you consider adding craftable recipe?

Oh, I just rechecked `ConsumableItemRecipeSheet` and 20`1`000 was what I originally attended.
Thanks for pointing out!

Unengine (approved):

PR #1718, opened 2023-02-14
Introduce transfer crystal white listing

ipdae:
- allow transfer crystal with set block index(6,200,000) with white list.


Lib9c/Action/TransferAsset.cs
```diff
@@ -9,21 +9,32 @@
...
-    /// Updated at https://github.com/planetarium/lib9c/pull/957
+    /// Updated at https://github.com/planetarium/lib9c/pull/1718
     /// </summary>
     [Serializable]
     [ActionType("transfer_asset3")]
     public class TransferAsset : ActionBase, ISerializable, ITransferAsset, ITransferAssetV1
     {
         private const int MemoMaxLength = 80;
 
+        public const long WhiteListAvailableIndex = 6_200_000L;
```
moreal, line 28: I think this interval is too close 🤔 I think it should be bigger as 300000 than the current block index (6087423)

longfin, line 28:
```suggestion
        public const long CrystalTransferringRestrictionStartIndex = 6_200_000L;
```

if we feel `CrystalTransferringRestriction` is too verbose, leveraging an external NCIP number (if exists) would be good too. cc @jaeho0103

Lib9c/Action/TransferAsset.cs
```diff
@@ -9,21 +9,32 @@
...
+
+        public static readonly IReadOnlyList<Address> WhiteList = new List<Address>
```
longfin, line 30:
```suggestion
        public static readonly IReadOnlyList<Address> AllowedCrystalTransferers = new List<Address>
```

longfin: Is there any external resource(e.g., NCIP, DAO proposal) to justify their addresses? @ipdae @jaeho0103

Lib9c/Action/TransferAsset.cs
```diff
@@ -9,21 +9,34 @@
...
     [ActionType("transfer_asset3")]
     public class TransferAsset : ActionBase, ISerializable, ITransferAsset, ITransferAssetV1
     {
         private const int MemoMaxLength = 80;
 
+        // FIXME justify this policy.
+        public const long CrystalTransferringRestrictionStartIndex = 6_220_000L;
+
+        // FIXME justify this policy.
+        public static readonly IReadOnlyList<Address> AllowedCrystalTransferList = new List<Address>
```
longfin, line 32:
```suggestion
        public static readonly ISet<Address> AllowedCrystalTransfers = new List<Address>
```

Sorry for the late & trivial checking. I think we don't need to keep the order. 😅

longfin (approved):

Lib9c/Action/TransferAssets.cs
```diff
@@ -9,6 +9,7 @@
 using System.Linq;
 using System.Runtime.Serialization;
 using Lib9c.Abstractions;
+using Nekoyume.Helper;
```
moreal, line 12: This addition seems unnecessary.

```suggestion
```

moreal (approved):

PR #1720, opened 2023-02-14
Introduce SecureMiningReward action

longfin:
This PR introduces `SecureMiningReward` to retrieve NCG from authorized miners to arbitrary addresses, for [NCIP-10](https://docs.google.com/document/d/1ErZ5JQia03KqXRG6IRZ7SORfnxMLZfJg4patVKFGX5Y/edit).

### DIFF
--- a/Lib9c/Action/SecureMiningReward.cs
+++ b/Lib9c/Action/SecureMiningReward.cs
@@ -1,0 +1,83 @@
+using System.Collections.Immutable;
+using Bencodex.Types;
+using Lib9c.Abstractions;
+using Libplanet;
+using Libplanet.Assets;
+using Libplanet.Action;
+using Nekoyume.Model.State;
+using System.Linq;
+
+namespace Nekoyume.Action
+{
+    [ActionType("secure_mining_reward")]
+    public class SecureMiningReward : ActionBase, ISecureMiningReward
+    {
+        // Copied from https://github.com/planetarium/lib9c/blob/v100361/Lib9c/Policy/BlockPolicySource.cs#L112-L118
+        // FIXME consider better way to check.
+        private static readonly ImmutableList<Address> AuthorizedMiners = new[]
+        {
+            new Address("ab1dce17dCE1Db1424BB833Af6cC087cd4F5CB6d"),
+            new Address("3217f757064Cd91CAba40a8eF3851F4a9e5b4985"),
+            new Address("474CB59Dea21159CeFcC828b30a8D864e0b94a6B"),
+            new Address("636d187B4d434244A92B65B06B5e7da14b3810A9"),
+        }.ToImmutableList();
+
+        // You can check about the treasury address and following rates from NCIP-10
+        // https://docs.google.com/document/d/1ErZ5JQia03KqXRG6IRZ7SORfnxMLZfJg4patVKFGX5Y/edit#
+        private static readonly Address Treasury = new Address("0xB3bCa3b3c6069EF5Bdd6384bAD98F11378Dc360E");
+
+        private const int TreasuryRate = 40;
+
+        private const int EarnRate = 20;
+
+        internal static readonly Currency NCG =
+            Currency.Legacy(
+                "NCG",
+                2,
+                ImmutableHashSet.Create(new Address("0x47D082a115c63E7b58B1532d20E631538eaFADde"))
+            );
+
+
+        public SecureMiningReward(Address recipient)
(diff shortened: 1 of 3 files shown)


Lib9c/Action/SecureMiningReward.cs
```diff
@@ -0,0 +1,65 @@
...
+            {
+                return state.MarkBalanceChanged(
+                    Amount.Currency,
+                    new [] { Miner, Recipient}
+                );
+            }
+
+            CheckPermission(context);
+
+            // TODO consider a better solution to determine.
```
longfin, line 47:
```suggestion
            // FIXME considers better solutions to check.
```

Lib9c/Action/SecureMiningReward.cs
```diff
@@ -0,0 +1,65 @@
...
+            if (!BlockChain.Policy.BlockPolicySource.AuthorizedMiners.Contains(Miner))
+            {
+                throw new InvalidAddressException($"Given address[{Miner}] isn't an authorized miner");
+            }
+
+            // TODO Do we need the maximum cap per miner? (but... if then, how do we maintain?)
```
longfin, line 53:
```suggestion
```

Lib9c/Action/SecureMiningReward.cs
```diff
@@ -0,0 +1,73 @@
...
+        );
+
+        public override IAccountStateDelta Execute(IActionContext context)
+        {
+            IAccountStateDelta state = context.PreviousStates;
+            if (context.Rehearsal)
+            {
+                return state.MarkBalanceChanged(
+                    Amount.Currency,
+                    new [] { Miner, Recipient}
```
moreal, line 51: Just style comment, you can ignore.

```suggestion
                    new [] { Miner, Recipient }
```

moreal:
Well, as I understood, the owner according to `AdminState` will control the assets of authorized miners. And according to NCIP-10, I understood the treasury (maybe Planetarium) will get 40%, 40% will be burnt and 20% must be retained. But the code seems not to handle anything about the rule. Is it based on the assumption that *Planetarium* can do it well? Or just not considered?

moreal (commented):
Well, as I understood, the owner according to `AdminState` will control the assets of authorized miners. And according to NCIP-10, I understood the treasury (maybe Planetarium) will get 40%, 40% will be burnt and 20% must be retained. But the code seems not to handle anything about the rule. Is it based on the assumption that *Planetarium* can do it well? Or just not considered?

longfin:
> Well, as I understood, the owner according to `AdminState` will control the assets of authorized miners. And according to NCIP-10, I understood the treasury (maybe Planetarium) will get 40%, 40% will be burnt and 20% must be retained. But the code seems not to handle anything about the rule. Is it based on the assumption that _Planetarium_ can do it well? Or just not considered?

It is assumed that the user who is authorized (i.e., passes `CheckPermission()`) performs "well". 😅

longfin:
> > It is assumed that the user who is authorized (i.e., passes `CheckPermission()`) performs "well". 😅
> 
> I'm very careful to reply. But, in my honest opinion, I think it is better to automate by implementing the NCIP-10 as the action code, to block the possibilities of mistakes. It seems able to be implemented.
> 
> Of course, it is just my preferred way so if they prefer this way, I will not object.

@moreal First, thanks for your opinion and I also understand that this action seems quite strange.

- Despiting I've mentioned NCIP-10, it lacks some information to apply directly (e.g., it doesn't specify the actual starting block or miner address).
   - but on second thought, it had an address to withdraw (i.e., `0xB3bCa3b3c6069EF5Bdd6384bAD98F11378Dc360E`). so it might be better to force that address.
- Even if we fixed block indices arbitrarily, it isn't matched the actual amount for a specific miner because we've changed the addresses of the miner for reasons such as outages.

To sum up, if we secure all miner's rewards for all periods instead of NCIP-10's, it can be executed automatically and easily.

@jckdotim do you have any idea?

longfin:
Amended https://github.com/planetarium/lib9c/pull/1720/commits/2d2d018a7e0d057168b638c77f0ba3013038e1b3  to address https://github.com/planetarium/lib9c/pull/1720#issuecomment-1439915335. 

PTAL @moreal @jckdotim

jckdotim:
@longfin I agree with your suggestion to automate the process as the action code to prevent mistakes.

moreal (approved):

PR #1580, opened 2022-12-13
Check raw custom actions in `BlockPolicySource`

moreal:
This pull request makes `BlockPolicySource` check actions through `ITransaction.CustomActions`, not been instantiated yet. Also, it makes `Miner` discard (i.e., unstage) transactions not available in the next block when the `IActionTypeLoader` is given.

PR #1725, opened 2023-02-17
Check custom actions deserialization

moreal:
There is a bug to disturb mining blocks by staging transactions having non-deserializable actions. So I applied #1580 changes, resolving some parts of the issue. But it didn't also check that the `IAction.LoadPlainValue()` can be executed well. This pull request makes it check that too.


Lib9c.Miner.Tests/Usings.cs
```diff
@@ -0,0 +1 @@
+global using Xunit;
```
longfin, line 1:
```suggestion
global using Xunit;

```

longfin (approved):

PR #1732, opened 2023-02-21
feat: Introduce test utils

boscohyun:
## Features

- CraftUtil
- CsvUtil
- CurrencyUtil
- InitializeUtil
- QuestUtil

### DIFF
--- a/.Lib9c.Tests/Util/CraftUtil.cs
+++ b/.Lib9c.Tests/Util/CraftUtil.cs
@@ -1,0 +1,75 @@
+namespace Lib9c.Tests.Util
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Globalization;
+    using Libplanet;
+    using Libplanet.Action;
+    using Nekoyume;
+    using Nekoyume.Action;
+    using Nekoyume.Model;
+    using Nekoyume.Model.Item;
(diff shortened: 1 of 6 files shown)


.Lib9c.Tests/Util/CraftUtil.cs
```diff
@@ -0,0 +1,75 @@
...
+            var slotAddress = avatarAddress.Derive(string.Format(
+                CultureInfo.InvariantCulture,
+                CombinationSlotState.DeriveFormat,
+                slotIndex));
+            var slotState = new CombinationSlotState(
+                slotAddress,
+                // CombinationEquipment: 3
+                // CombinationConsumable: 6
+                // ItemEnhancement: 9
+                GameConfig.RequireClearedStageLevel.ItemEnhancementAction
```
U-lis, line 33: Originally this code is written by me. I want review for this code.
Is it better to give action type from function caller?

U-lis (approved):

.Lib9c.Tests/Util/CraftUtil.cs
boscohyun: Yes that's right! We can apply it next time.🙂

PR #1738, opened 2023-02-22
Sync class name and filename

moreal:
It just fixes a typo that doesn't sync class name and filename.


moreal:
Because I think it doesn't affect functions, I'll merge in force.

PR #1745, opened 2023-02-22
🔧 Fixed index used for policy

greymistcube:
Internally, there are two places using `index`, namely `IsObsolete()` (which also internally uses `IActionTypeLoader.Load()`) and `IActionTypeLoader.Load()`. At a glance, seems like this should be fine.

- ~~`IsObsolete()`: If an action is obsoleted at certain index and permanently onwards, passing in a higher index should be fine, as this results in a more lax condition.~~
  - ~~Suppose some action $X$ was supposed to be obsoleted at block $n$. Before, with **wrong implementation**, action $X$ would've been banned starting from block $n - 1$. Policy is enforced as long as $X$ is not in block $n$.~~
  - As @moreal has pointed out below, this line of reasoning is flawed.
- `IActionTypeLoader.Load()`: This is also used inside `Miner.MineBlockAsync()` with a **proper** block index. So blocks were being mined using the right action. As for policy consideration, seems like inside `ValidateNextBlockTxRaw()` it is only used for account activation check. From what I gather so far, some account activation actions might've failed around the index where the implementation of `IActivateAccount` changed from `activate_account` to `activate_account2`.

### DIFF
--- a/Lib9c.Policy/Policy/BlockPolicySource.cs
+++ b/Lib9c.Policy/Policy/BlockPolicySource.cs
@@ -276,9 +276,7 @@ namespace Nekoyume.BlockChain.Policy
             ImmutableHashSet<Address> allAuthorizedMiners)
         {
             // Avoid NRE when genesis block appended
-            // Here, index is the index of a prospective block that transaction
-            // will be included.
-            long index = blockChain.Count > 0 ? blockChain.Tip.Index : 0;
+            long index = blockChain.Count > 0 ? blockChain.Tip.Index + 1: 0;
 
             if (((ITransaction)transaction).CustomActions?.Count > 1)
             {
--- a/Lib9c.Policy/Policy/BlockPolicySource.Utils.cs
+++ b/Lib9c.Policy/Policy/BlockPolicySource.Utils.cs
@@ -18,6 +18,25 @@ namespace Nekoyume.BlockChain.Policy
     // Collection of helper methods not directly used as a pluggable component.
     public partial class BlockPolicySource
     {
+        /// <summary>
+        /// <para>
+        /// Checks if <paramref name="transaction"/> includes any <see cref="IAction"/> that is
+        /// obsolete according to <see cref="ActionObsoleteAttribute"/> attached.
+        /// </para>
+        /// <para>
+        /// Due to a bug, an <see cref="IAction"/> is considered obsolete starting from
+        /// <see cref="ActionObsoleteAttribute.ObsoleteIndex"/> + 2.
+        /// </para>
+        /// </summary>
+        /// <param name="transaction">The <see cref="Transaction{T}"/> to consider.</param>
+        /// <param name="actionTypeLoader">The loader to use <see cref="IAction"/>s included
+        /// in <paramref name="transaction"/>.</param>
+        /// <param name="blockIndex">Either the index of a prospective block to include
+        /// <paramref name="transaction"/> or the index of a <see cref="Block{T}"/> containing
+        /// <paramref name="transaction"/>.</param>
+        /// <returns><see langword="true"/> if <paramref name="transaction"/> includes any
+        /// <see cref="IAction"/> that is obsolete, <see langword="false"/> otherwise.</returns>
+        /// <seealso cref="ActionObsoleteAttribute"/>
         internal static bool IsObsolete(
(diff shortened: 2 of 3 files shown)


Lib9c.Policy/Policy/BlockPolicySource.cs
```diff
@@ -278,7 +278,7 @@ public partial class BlockPolicySource
             // Avoid NRE when genesis block appended
             // Here, index is the index of a prospective block that transaction
             // will be included.
-            long index = blockChain.Count > 0 ? blockChain.Tip.Index : 0;
+            long index = blockChain.Count > 0 ? blockChain.Tip.Index + 1 : 0;
```
moreal, line 281: I think it has a possibility to make some blocks having obsoleted actions, invalid. I described the situation I imagined. The Nth block (having obsoleted actions at N index) will be invalid when appending Nth block.

image

So I think it should be branched by block index to avoid making some blocks invalid.
```suggestion
            long index = blockChain.Count > 0 ? (blockChain.Tip.Index > SOME_HARDFORK_INDEX ? blockChain.Tip.Index + 1 : blockChain.Tip.Index) : 0;
```

If my assumption was wrong, please let me know it 🙏🏻 .

greymistcube: Hmm... I guess I had it wrong. Seems like there is no easy way around this issue then. 😥

greymistcube: Okay, the more I think about it, I think we need to do some damage control at this point. 😣

greymistcube:
@moreal BTW, what's the purpose of `IActionTypeLoaderContext.Index` anyway?

moreal:
> @moreal BTW, what's the purpose of `IActionTypeLoaderContext.Index` anyway?

@greymistcube The purpose is to make `IActionTypeLoader.Load` able to load action types by the block index. (e.g., `DynamicActionTypeLoader`, you can see it in NineChronicles.Headless project)

Lib9c.Policy/Policy/BlockPolicySource.cs
```diff
@@ -287,7 +290,7 @@ public partial class BlockPolicySource
                     $"{((ITransaction)transaction).CustomActions?.Count}",
                     transaction.Id);
             }
-            else if (IsObsolete(transaction, actionTypeLoader, index))
+            else if (IsObsolete(transaction, actionTypeLoader, improperIndex))
```
longfin, line 293: How about using `properIndex + 1` (or just `index + 1`) and leaving a comment here? (I think `improperIndex` term can be a little burden to understand)

longfin (approved):

greymistcube:
https://github.com/planetarium/lib9c/blob/715d86bcb4412dd124a359e8220edb1d0e4c074f/Lib9c.Policy/Policy/BlockPolicySource.Utils.cs#L21-L43

So... there seems to be even more problem with action obsolescence.
If an action is decorated with `ActionObsolete` attribute with `ObsoleteIndex` of $n$, `IsObsolete()` considers an action obsolete starting from index $n + 1$ not $n$. As a result, we have the following for the current implementation:

- Let $X$ be an action with `ActionObsolete` attribute with `ObsoleteIndex` of $n$.
- When `IsObsolete()` is called with $X$ and some index $k$, it returns true only if $n < k$.
- $X$ can be included in block $n$ as `IsObsolete()` is called with $k = n - 1$ and obviously we have $n \not< k = n - 1$.
- $X$ can be included in block $n + 1$ as `IsObsolete()` is called with $k = n$ and we still have $n \not < k = n$.
- $X$ cannot be included in block $n + 2$ as `IsObsolete()` is called with $k = n + 1$ and we finally have $n < k = n + 1$.

I was thinking about renaming `ActionObsolete` attribute to `ActionValidUntil` to get around the edge case issue, I've discovered that this is not viable since we are off by two. 😥

greymistcube:
https://github.com/planetarium/lib9c/commit/ef8c3dccba754941dff3882c6bf283714562b556

Tried to put up warning signs here and there and reduce the scope of possible confusion to `IsObsolete()`. 🙄

@moreal @longfin 
Can you double check the logic?

moreal (approved):
It seems okay to me.

PR #1746, opened 2023-02-23
Introduce `ManipulateState` action for easier test environment setting

U-lis:
## Update
- `ManipulateState` action force sets given state to given address.
- This function assumes given address and state are in pair.
- This function assumes given state is valid.

### Further work
- Add state editor to unity editor and use this action to save changed state

---

## Features
- `InitializeUtil` now can handle and create inits for DevEx tests.
- `UnlockCraftAction` can unlock craft related action by clearing required stage.
- `UnlockRecipe` can unlock Item reipe to craft.
- `UnlockStage` can unlock stages to test specific stage easily.

### Comment Needed
I wonder I did right thing to set new WorldInformation to clear stage at `ClearStage.cs` workaround.

### DIFF
--- a/Lib9c.DevExtensions/Action/Craft/UnlockCraftAction.cs
+++ b/Lib9c.DevExtensions/Action/Craft/UnlockCraftAction.cs
@@ -1,0 +1,78 @@
+using System;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using Bencodex.Types;
+using Libplanet;
+using Libplanet.Action;
+using Nekoyume;
+using Nekoyume.Action;
+using Nekoyume.Exceptions;
+using Nekoyume.Model;
+using Nekoyume.Model.State;
+using Nekoyume.TableData;
+using static Lib9c.SerializeKeys;
+
+namespace Lib9c.DevExtensions.Action.Craft
+{
+    [Serializable]
+    [ActionType("unlock_craft_action")]
+    public class UnlockCraftAction : GameAction
+    {
+        public Address AvatarAddress { get; set; }
+        public ActionTypeAttribute ActionType { get; set; }
+
+        public override IAccountStateDelta Execute(IActionContext context)
+        {
+            if (context.Rehearsal)
+            {
+                return context.PreviousStates;
+            }
+
+            var states = context.PreviousStates;
+            int targetStage;
+
+            if (ActionType.TypeIdentifier.Contains("combination_equipment"))
+            {
+                targetStage = GameConfig.RequireClearedStageLevel.CombinationEquipmentAction;
+            }
+            else if (ActionType.TypeIdentifier.Contains("combination_consumable"))
+            {
+                targetStage = GameConfig.RequireClearedStageLevel.CombinationConsumableAction;
+            }
+            else if (ActionType.TypeIdentifier.Contains("item_enhancement"))
+            {
+                targetStage = GameConfig.RequireClearedStageLevel.ItemEnhancementAction;
+            }
+            else
+            {
+                throw new InvalidActionFieldException(
+                    $"{ActionType.TypeIdentifier} is not valid action");
+            }
+
+            var worldInformation = new WorldInformation(
+                context.BlockIndex,
+                states.GetSheet<WorldSheet>(),
+                targetStage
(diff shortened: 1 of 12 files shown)


Lib9c.DevExtensions/Action/Craft/UnlockCraftAction.cs
```diff
@@ -0,0 +1,78 @@
...
+using static Lib9c.SerializeKeys;
+
+namespace Lib9c.DevExtensions.Action.Craft
+{
+    [Serializable]
+    [ActionType("unlock_craft_action")]
+    public class UnlockCraftAction : GameAction
+    {
+        public Address AvatarAddress { get; set; }
+        public ActionTypeAttribute ActionType { get; set; }
```
boscohyun, line 22: Can we change the type of `ActionType` to `string`? Because this action needs only the `ActionTypeAttribute.TypeIdentifier`.🙂

Lib9c.DevExtensions/Action/Stage/ClearStage.cs
```diff
@@ -0,0 +1,54 @@
...
+
+            var states = context.PreviousStates;
+            var worldInformation = new WorldInformation(
+                context.BlockIndex,
+                states.GetSheet<WorldSheet>(),
+                TargetStage
+            );
+            return states.SetState(AvatarAddress.Derive(SerializeKeys.LegacyWorldInformationKey),
+                worldInformation.Serialize()
+            );
```
boscohyun, line 36: It's enough to clear the stage.👍 One thing in the real scenario, Avatar's quest list will be updated with the inventory. But that's OK, because the actions in DevEx are not as strict as in reality.🙂

U-lis:
We will use `ManipulateState` action to change data, but I'll leave individual actions for historical reason :bookmark_tabs:

Lib9c.DevExtensions/Action/ManipulateState.cs
```diff
@@ -0,0 +1,41 @@
...
+        public override IAccountStateDelta Execute(IActionContext context)
+        {
+            var state = context.PreviousStates;
+            foreach (var (address, value) in StateList)
+            {
+                state = state.SetState(address, value);
+            }
+
+            return state;
+        }
```
boscohyun, line 27: Could you handle the case where `context.Rehearsal` is `true`?

Lib9c/Model/Quest/QuestList.cs
```diff
@@ -82,7 +82,7 @@ EquipmentItemSubRecipeSheet equipmentItemSubRecipeSheet
         )
         {
             _quests = new List<Quest>();
-            foreach (var questData in questSheet.OrderedList)
+            foreach (var questData in questSheet.OrderedList ?? new List<QuestSheet.Row>())
```
boscohyun, line 85: This changes can occur a non-deterministic error.🥲

U-lis: > This changes can occur a non-deterministic error.smiling_face_with_tear

Oh, I see. But in that case, is there are general way to handle empty input is incoming?
This is because an error occurs if we use `new QuestSheet()` to remove all the quests.

U-lis:
rebase makes commit history dirty, but just [c700cf1](https://github.com/planetarium/lib9c/pull/1746/commits/c700cf16e7ff67572169187734e820e92352b462) and [3284319](https://github.com/planetarium/lib9c/pull/1746/commits/3284319d993b123f50ef739b1dd7c845f6413c6b) are real changes.

Lib9c/Model/Quest/QuestList.cs
```diff
@@ -82,21 +82,24 @@ EquipmentItemSubRecipeSheet equipmentItemSubRecipeSheet
...
+
+                    var quest = CreateQuest(questData, reward, equipmentItemRecipeSheet);
+                    if (quest is null)
+                    {
+                        continue;
+                    }
 
-                _quests.Add(quest);
+                    _quests.Add(quest);
+                }
```
boscohyun, line 102: This change can cause inconsistency issues.🥲

boscohyun (approved):

PR #1757, opened 2023-02-27
Check whether the build passed or failed in several configurations and frameworks

moreal:
I opened this pull request to check it is compatible with Unity and the `DevEx` environment automatically without a situation to hold developers accountable. 🙄 

## Check ***Unity*** compatibility

The job enforces the `netstandard2.1` target for the `Lib9c*/*.csproj` and `Libplanet*/*.csproj` projects. It became a .net 6 project but the planetarium/NineChronicles project is still using `Lib9c` as a submodule. So it should check Unity(netstandard2.1) compatibility. See #1752 case.

## Check "DevEx" build also

Maybe, since #1734, the `LIb9c` build with the `DevEx` configuration was broken but not caught in CI. This pull request makes CI do it.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -51,6 +51,26 @@ jobs:
     - name: Test
       run: dotnet test --no-restore --verbosity normal --configuration ${{ matrix.configuration }}
 
+  build-for-unity:
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          submodules: true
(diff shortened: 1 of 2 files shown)


moreal:
I updated how to calculate the default table sheet directory in the `ImportSheets` method. And it is okay to fail the `main / build-and-test (DevEx) (pull_request)` job at this step.

@boscohyun Could you review this pull request again?

boscohyun:
The new path calculate logic looks good.🙂

PR #1768, opened 2023-03-02
obsolete transfer asset2

ipdae:
- related with #1718 .
- Even if the set block height (622,000) is reached, crystal transfer with TransferAsset2 is still possible. to avoid this problem, prevent the execution of TransferAsset2 from the set height.


sonohoshi (approved):

moreal (approved):
If it is merged and planned to release, it should be released ASAP.

PR #1786, opened 2023-03-08
Suppress warnings while collecting action type ids

moreal:
Since v100370, the `v*/all_action_type_ids.txt` begins to include a .NET command warning message. (I'm not sure why the warning messages are in stdout, not stderr 🙄 ...) This pull request makes the job suppress the warning message by controlling the [`WarningLevel`][warning-level] property as zero.

(I asked for reviews from those who know this context.)

[warning-level]: https://learn.microsoft.com/ko-kr/dotnet/api/microsoft.build.tasks.csc.warninglevel?view=netframework-4.8.1#microsoft-build-tasks-csc-warninglevel

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -112,7 +112,7 @@ jobs:
     - name: Collect available action type ids
       run: |
         mkdir publish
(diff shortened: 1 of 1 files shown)


moreal:
I merge this pull request because the change is simple.

PR #1800, opened 2023-03-14
[DevEx] feat: Apply balance list to ManipulateState action

boscohyun:
SSIA

### DIFF
--- a/.Lib9c.DevExtensions.Tests/Action/ManipulateStateTest.cs
+++ b/.Lib9c.DevExtensions.Tests/Action/ManipulateStateTest.cs
@@ -2,6 +2,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Collections.Immutable;
 using System.Linq;
 using Bencodex.Types;
 using Lib9c.DevExtensions.Action;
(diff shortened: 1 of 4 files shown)


.Lib9c.DevExtensions.Tests/Action/ManipulateStateTest.cs
```diff
@@ -503,14 +579,21 @@ public void SetMultipleStates()
             var questData = FetchQuest().Last();
             var targetQuestIdList = (List<int>)questData[0];
             var questList = (QuestList)questData[1];
+            var balanceList = FetchBalance()
+                .Select(objects => (addr: (Address)objects[0], fav: (FungibleAssetValue)objects[1]))
+                .Where(tuple => !(tuple.fav.Currency.Ticker == "NCG" && tuple.fav.MajorUnit > 100))
```
U-lis, line 584: Should we ignore balance more than 100 NCG?

boscohyun: The return values of `FetchBalance()` method has the `FungibleAssetValue` which contains the maximum quantity of `NCG` the `GoldBalanceState` can has.😅
So this line avoid that because `IAccountStateDelta.TransferAsset()` can throw an exception.

U-lis (approved):

PR #1803, opened 2023-03-15
development merge prepare-v100380

ipdae:


.Lib9c.Tests/Action/ItemEnhancementTest.cs
sonohoshi, line 1: This test use `ItemEnhancement11`. it's ok?
I think it seems to use `ItemEnhancement` or ~12.

sonohoshi (changes requested):

.Lib9c.Tests/Action/ItemEnhancementTest.cs
ipdae: there is no problem because the enum value is the same, but I will fix it.

sonohoshi (approved):

Unengine:
CI is broken.

PR #1855, opened 2023-04-06
feat: Introduce `Currencies`

boscohyun:
## Currencies

- The class `Currencies` summarizes all currencies like `Addresses`.
- Use `Currencies.Crystal` from `CrystalCalculator.CRYSTAL`.
- Use `Currencies.StakeRune` and `Currencies.DailyRewardRune` from `RuneHelper.StakeRune` and `RuneHelper.DailyRewardRune`.
- Use `Currencies.GetRune(ticker)` from `RuneHelper.ToCurrency(runeRow)` and `RuneHelper.ToFungibleAssetValue(runeRow, quantity)`.
- Use `Currencies.GetSoulStone(ticker)` from `PetHelper.GetSoulstoneCurrency(ticker)`.

## RuneHelper

- Remove the `decimalPlaces` and `minters` parameters as they are always `0` and `null`.

### DIFF
--- a/Lib9c/Currencies.cs
+++ b/Lib9c/Currencies.cs
@@ -1,0 +1,76 @@
+#nullable enable
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Libplanet;
+using Libplanet.Assets;
+using Nekoyume.TableData;
+
+namespace Nekoyume
+{
+    public static class Currencies
+    {
+        public static readonly Currency NCG = Currency.Legacy(
+            "NCG",
+            2,
+            new Address("0x47D082a115c63E7b58B1532d20E631538eaFADde"));
+
+        public static readonly Currency Crystal = Currency.Legacy(
+            "CRYSTAL",
+            18,
+            minters: null);
+
+        public static readonly Currency StakeRune = Currency.Legacy(
+            "RUNE_GOLDENLEAF",
+            0,
+            minters: null);
+
+        public static readonly Currency DailyRewardRune = Currency.Legacy(
+            "RUNE_ADVENTURER",
+            0,
+            minters: null);
+
+        public static Currency GetRune(string? ticker) =>
+            string.IsNullOrEmpty(ticker)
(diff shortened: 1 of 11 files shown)


Lib9c/Currencies.cs
```diff
@@ -0,0 +1,70 @@
...
+using Nekoyume.TableData;
+
+namespace Nekoyume
+{
+    public static class Currencies
+    {
+        public static readonly Currency NCG = Currency.Legacy(
+            "NCG",
+            2,
+            new Address("0x47D082a115c63E7b58B1532d20E631538eaFADde"));
```
moreal, line 17: It should be configurable because when the action logic uses `Currencies.NCG` it will not work in another forked network (because the minter will be different)

Lib9c/Currencies.cs
```diff
@@ -0,0 +1,70 @@
+#nullable enable
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Libplanet;
+using Libplanet.Assets;
+using Nekoyume.TableData;
+
+namespace Nekoyume
```
moreal, line 10: In a new file, isn't it okay to use the `Lib9c` namespace? 🤔 (I don't have a strong opinion for this suggestion).

moreal (approved):

Lib9c/Currencies.cs
boscohyun: Oh, thanks! I'll change.🙂

Lib9c/Currencies.cs
boscohyun: @moreal Yes, you are right. And some other forked network can hard-fork the NCG's minters in this line.🙂
After the pluggable-lib9c is applied, use `Currencies.NCG` in another context. There are no uses of `Currencies.NCG` now.

moreal (approved):

boscohyun:
@moreal Please check the c78b1fd commit.🙂

moreal (approved):

PR #1860, opened 2023-04-17
Introduce `ClaimStakeReward4` action

U-lis:
### Breaking changes
- `Addresses` has TableData version control data as Dictionary.
    - Sheet name and block index finalizes target TableData version.
    - All sheet retrieving logic should pass current block index to get latest version of sheet
- Calculates stake reward change with keeping past unclaimed reward
    - Calculate reward steps before/after update as `v1Step`, `v2Step`.
    - Use same `ProcessReward` action to give rewards.


U-lis:
https://github.com/planetarium/lib9c/pull/1861 is merged.

PR #1861, opened 2023-04-17
ClaimStakeReward4 action with hard-coded old data

U-lis:
### This is same change as [former PR](https://github.com/planetarium/lib9c/pull/1860). This is for comparison.

Old data is not from newly derived address, but hard-coded data.
Another core logic (splitting v1, v2 step, using `ProcessReward` function) are same.


Lib9c/TableData/TableDataVersionConfig.cs
```diff
@@ -0,0 +1,9 @@
+namespace Nekoyume.TableData
+{
+    public static class TableDataVersionConfig
+    {
+        // PatchTableSheet version update without changing model version
+        // FIXME: Set block index to real block index to change reward version
+        public const long StakeRewardSheetV2Index = 7_000_000L;
```
ipdae, line 7: it will be better move to place in `StakeState`

ipdae (commented):
there are no difference reward calculation before the designated block index, right? it seems possible without hard fork `ClaimStakeReward`.

Lib9c/Action/ClaimStakeReward.cs
```diff
@@ -12,25 +16,131 @@
...
+    /// Hard forked at https://github.com/planetarium/lib9c/pull/1458
     /// </summary>
     [ActionType(ActionTypeText)]
     public class ClaimStakeReward : GameAction, IClaimStakeReward, IClaimStakeRewardV1
     {
-        public const long ObsoletedIndex = 5_549_200L;
-        private const string ActionTypeText = "claim_stake_reward2";
+        private const string ActionTypeText = "claim_stake_reward4";
+
+        private const string StakeRegularRewardSheetV1Data =
```
ipdae, line 26: previous reward data should be shown before the specified block index in the client side, maybe it will be better to change it to public.

ipdae (approved):

PR #1876, opened 2023-04-28
Add stat reference in buff value

Unengine:
## Features
- Make `Skill.Power` decide the value of equipment's buff.
- `StatBuffSheet.Row.ReferencedStatType`  specifies the referenced `StatType` of the value.

## How it works
- When referenced `StatType` is specified : `buff value = original value + (referenced stat * (Skill.Power / 10000))`
- When referenced `StatType` not specified : `buff value = original value + Skill.Power`


Lib9c/TableData/Skill/StatBuffSheet.cs
```diff
@@ -33,8 +33,11 @@ public class Row : SheetRow<int>
             public int Duration { get; private set; }
 
             public SkillTargetType TargetType { get; private set; }
-            public StatModifier StatModifier { get; private set; }
-            public string IconResource { get; private set; }
```
ipdae, line 37: is this correct change?

ipdae (commented):
it might be occur exception when migrate sheet data. it needs to be fixed with pre-patch data.

Lib9c/TableData/Skill/StatBuffSheet.cs
Unengine: `IconResource` is not used in state, it's parse is replaced with `TryParse`. And `StatModifier` field is replaced into this.
```
            public int Value { get; private set; }
            public StatType ReferencedStatType { get; private set; }
```

ipdae (approved):

PR #1877, opened 2023-04-28
Calculate UpdatedAddresses on Lib9c side

longfin:
# Summary

This PR adds `.CalculateUpdatedAddresses()` to `IEnumerable<PolymorphicAction<ActionBase>>` and `IEnumerable<ActionBase>` to get updated addresses handy.

# Background

## `Transaction<T>.UpdatedAddresses` is no longer auto-calculated.

- This feature isn't necessary for Libplanet 1.0.* yet, because [it will be calculated on `Transaction<T>.Create()`](https://github.com/planetarium/libplanet/blob/1.0.0/Libplanet/Tx/Transaction.cs#L431-L432). 
- But also, Libplanet has planned to remove auto calculation since it isn't realistic to compute all through previous states to make a transaction, and complicates the implementation. 
- see also: 
  - https://github.com/planetarium/libplanet/issues/368
  - https://github.com/planetarium/libplanet/pull/3122
 
## Then, why do we still need it?

- However even now, 'UpdatedAddresses' are automatically written in transactions on 9c mainnet. 
  - This is possible because most of the actions so far are writing compatible code assuming the simple state base based on the `IActionContext.Rehearsal`.
  - E.g., https://github.com/planetarium/lib9c/blob/6912e58ceb7cc29d3ac9ef9fa4b5667af0158192/Lib9c/Action/TransferAsset.cs#L92-L98
- Sadly, related systems in 9c mainnet already seems to be relying `Transaction<T>.UpdatedAddresses`.
  - E.g., https://github.com/planetarium/9cscan-cloud/blob/0760e14f4ed07997d96784fcbc3841cd6790083f/project/sync/src/sync.js#L34-L40
- Thus, even if Libplanet decides to make the block layout compatible, the related system compatibility may be broken against empty `UpdatedAddresses`.
- As one of Libplanet devs, I really want to (re)organize tx/action evaluation flow without the auto-filling assumption of updated addresses. 
- but also as one of 9c devs, I don't want to face an ironic situation like, updating Libplanet (having "good design") kills related services in the ecosystem.

# Plan (9c side only)

- [ ] Add methods to calculate `.UpdatedAddresses` leveraging existing `Rehearsal` related codes to Lib9c (this PR)
- [ ] Add `updatedAddresses` argument explicitly for `Transaction<T>.Create()` when creating transactions (NineChronicles / NineChronicles.Headless).
- [ ] (Optional) Provide a migration path from relying on `Transaction<T>.UpdatedAddresses` to using `ActionEvaluation.UpdatedAddresses` for systems involving 9c mainnet.

### DIFF
--- a/.Lib9c.Tests/Action/ActionBaseExtensionsTest.cs
+++ b/.Lib9c.Tests/Action/ActionBaseExtensionsTest.cs
@@ -1,0 +1,33 @@
+namespace Lib9c.Tests.Action
+{
+    using System.Collections.Immutable;
+    using Libplanet;
+    using Libplanet.Action;
+    using Nekoyume.Action;
+    using Xunit;
+
+    public class ActionBaseExtensionsTest
+    {
+        [Fact]
+        public void CalculateUpdateAddresses()
+        {
+            var actions = new PolymorphicAction<ActionBase>[]
+            {
+                new TransferAsset(
+                    sender: default,
+                    recipient: default,
+                    amount: Currencies.DailyRewardRune * 1
+                ),
+            };
+
+            IImmutableSet<Address> actual = actions.CalculateUpdateAddresses();
+            Assert.Equal(
+                new Address[]
+                {
+                    default,
+                },
+                actual
+            );
+        }
+    }
+}
--- a/Lib9c/Action/ActionBaseExtensions.cs
+++ b/Lib9c/Action/ActionBaseExtensions.cs
@@ -1,0 +1,139 @@
+using System;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Linq;
+using System.Security.Cryptography;
+using Bencodex.Types;
+using Libplanet;
+using Libplanet.Action;
+using Libplanet.Assets;
+using Libplanet.Blocks;
+using Libplanet.Tx;
+
+namespace Nekoyume.Action
+{
+    public static class ActionBaseExtensions
+    {
+        public static IImmutableSet<Address> CalculateUpdateAddresses(
+            this IEnumerable<PolymorphicAction<ActionBase>> actions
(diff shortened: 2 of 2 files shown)


moreal:
> - Sadly, related systems in 9c mainnet already seems to be relying Transaction<T>.UpdatedAddresses.
>     - E.g., https://github.com/planetarium/9cscan-cloud/blob/0760e14f4ed07997d96784fcbc3841cd6790083f/project/sync/src/sync.js#L34-L40

I think it is not an issue related to `Transaction<T>.UpdatedAddresses`. If they want to know what states are updated by actions, they should use `TxSuccess.UpdatedStates` but they and we don't want to store the updated states too so they cannot use them. I think that is the real problem.

moreal (approved):
At the code level, I think it can be merged.

longfin:
> I think it is not an issue related to `Transaction<T>.UpdatedAddresses`. If they want to know what states are updated by actions, they should use `TxSuccess.UpdatedStates` but they and we don't want to store the updated states too so they cannot use them. I think that is the real problem.

It's true. but the point I wanted to mention is, many systems are already using `Transaction<T>.UpdatedAddresses` with assuming previous behavior (i.e., autofill with dry-run). in this situation, https://github.com/planetarium/libplanet/pull/3122 would be a breaking change if we don't have any other polyfill (like this PR), to mimic.

But on the other hand, I believe that we need to introduce a legit way to access "updated addresses", like as you mentioned. 😅

moreal (approved):

longfin:
Just for note) At least in v200010, many of the actions in Lib9c already seem to stop filling `UpdatedAddresses` using `IActionContext.Rehearsal`. 😅 (e.g., [HAS20](https://github.com/planetarium/lib9c/blob/v200010/Lib9c/Action/HackAndSlash.cs#L100-L103C14))

PR #1878, opened 2023-05-02
fix `SellCancellation9` invalidstateroothashexception from #6166598

area363:
image
https://9cscan.com/tx/9f736e20b2ce8dd923898e0be0b95c582defd0a76156368ffe976bdfacf1d1db

This `SellCancellation9` action failed in the mainnet cannonical chain but it actually succeeds when executing it with our current lib9c version and returns an `InvalidStateRootHashException`. This is a fix to manually skip this action and resolve the `InvalidStateRootHashException` error (I've used this to move on to the next block evaluation while doing other work).

### DIFF
--- a/Lib9c/Action/SellCancellation.cs
+++ b/Lib9c/Action/SellCancellation.cs
@@ -88,11 +88,6 @@ namespace Nekoyume.Action
 
         public override IAccountStateDelta Execute(IActionContext context)
         {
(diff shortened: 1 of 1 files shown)


moreal (approved):
I checked it failed. (If the 9c-main-full-state doesn't lie)

area363:
This PR is updated to resolve multiple incorrect `SellCancellation9` executions which should have returned either `InvalidTradableIdException` or `OrderIdDoesNotExistException` according to the mainnet chain. These errors occur from `#6166598` so the PR modifies the action to use `CancelV2` method beginning from `#6166598`.

ipdae (approved):

PR #1923, opened 2023-05-19
♻️ Remove `PolymorphicAction<T>` and use `ActionBase`

greymistcube:
This removes the usage of `PolymorphicAction<T>` in preparation for removing `PolymorphicAction<T>` from [libplanet](https://github.com/planetarium/libplanet) completely. This only pertains to actions with an `ActionTypeAttribute` attached.

Changes made:
- Use a `TypedActionLoader` with action type `ActionBase` instead of a `SingleActionLoader` with action type `PolymorphicAction<ActionBase>`.
- For those inheriting `GameAction`, automatic transformation of "inner" `PlainValue`s to the form `{ "type_id": <type_id>, "values": <values> }` has been moved from `PolymorphicAction<T>` to `GameAction`.
  - This means, there is no way to retrieve "inner" `PlainValue` (that corresponds to `<values>` above) directly.
- For those not inheriting `GameAction`, i.e. those inheriting `ActionBase` directly, each individual `IAction` is responsible for making its "new" `PlainValue` compatible with the old format of `{ "type_id": <type_id>, "values": <values> }`.
- All `LoadPlainValues()` expects the "full" `PlainValue` of the form `{ "type_id": <type_id>, "values": <values> }`.


moreal (approved):
If we keep this up, we may see a lot of mistakes in the future about modifying `PlainValue["type_id"]` and `ActionTypeAttribute.TypeId` non-equally. @planetarium/9c-dev has also to confirm this.

greymistcube:
I agree that we might need to further refine the interface (by introducing a new `interface` such as `ITypedAction`)/API (by implementing more functionality in `ActionBase`). I haven't quite decided which is better. 🙄

PR #1949, opened 2023-06-01
Introduce garages

boscohyun:
- Introduce `FungibleItemGarage` model.
- Introduce `LoadIntoMyGarages` action.
- Introduce `DeliverToOthersGarages` action.
- Introduce `UnloadFromMyGarages` action.
- Introduce `GARAGE` token.

### DIFF
--- a/Lib9c.Abstractions/ILoadIntoMyGarages.cs
+++ b/Lib9c.Abstractions/ILoadIntoMyGarages.cs
@@ -1,0 +1,20 @@
+#nullable enable
+
+using System.Linq;
+using System.Security.Cryptography;
+using Libplanet;
+using Libplanet.Assets;
+
+namespace Lib9c.Abstractions
+{
+    public interface ILoadIntoMyGarages
+    {
+        IOrderedEnumerable<(Address balanceAddr, FungibleAssetValue value)>? FungibleAssetValues
+        {
+            get;
+        }
+
+        Address? InventoryAddr { get; }
+        IOrderedEnumerable<(HashDigest<SHA256> fungibleId, int count)>? FungibleIdAndCounts { get; }
+    }
+}
--- a/Lib9c/Action/Garages/DeliverToOthersGarages.cs
+++ b/Lib9c/Action/Garages/DeliverToOthersGarages.cs
@@ -1,0 +1,265 @@
+#nullable enable
+
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Linq;
+using System.Security.Cryptography;
+using Bencodex.Types;
+using Lib9c.Abstractions;
+using Libplanet;
+using Libplanet.Action;
+using Libplanet.Assets;
+using Libplanet.State;
+using Nekoyume.Exceptions;
+using Nekoyume.Model.Garages;
+using Nekoyume.Model.State;
+
+namespace Nekoyume.Action.Garages
+{
+    [ActionType("deliver_to_others_garages")]
+    public class DeliverToOthersGarages : GameAction, IDeliverToOthersGarages, IAction
+    {
+        public Address RecipientAgentAddr { get; private set; }
+        public IOrderedEnumerable<FungibleAssetValue>? FungibleAssetValues { get; private set; }
+
+        public IOrderedEnumerable<(HashDigest<SHA256> fungibleId, int count)>?
+            FungibleIdAndCounts { get; private set; }
+
+        protected override IImmutableDictionary<string, IValue> PlainValueInternal =>
+            new Dictionary<string, IValue>
+            {
+                {
+                    "l",
+                    new List(
+                        RecipientAgentAddr.Serialize(),
+                        FungibleAssetValues is null
+                            ? (IValue)Null.Value
+                            : new List(FungibleAssetValues.Select(fav => fav.Serialize())),
+                        FungibleIdAndCounts is null
+                            ? (IValue)Null.Value
+                            : new List(FungibleIdAndCounts.Select(tuple => new List(
+                                tuple.fungibleId.Serialize(),
+                                (Integer)tuple.count))))
+                }
+            }.ToImmutableDictionary();
+
+        public DeliverToOthersGarages(
+            Address recipientAgentAddr,
+            IEnumerable<FungibleAssetValue>? fungibleAssetValues,
+            IEnumerable<(HashDigest<SHA256> fungibleId, int count)>? fungibleIdAndCounts)
+        {
+            RecipientAgentAddr = recipientAgentAddr;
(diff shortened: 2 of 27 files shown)


Lib9c/Action/Garages/UnloadFromMyGarages.cs
```diff
@@ -0,0 +1,196 @@
...
+                if (count < 0)
+                {
+                    throw new InvalidActionFieldException(
+                        $"[{addressesHex}] Count of fungible id must be positive." +
+                        $" {fungibleId}, {count}");
+                }
+            }
+        }
+
+        private IAccountStateDelta TransferFungibleAssetValues(
```
U-lis, line 150: Do we have to write down this function in load/unload separately? (same as TransferFungibleItems)
In my short sight, just switching sender and recipient do the thing right.
If I'm missing something, feel free to let me know.

Lib9c/Currencies.cs
```diff
@@ -36,6 +37,41 @@ public static class Currencies
             0,
             minters: null);
 
+        public static readonly Currency Garage = Currency.Uncapped(
+            "Garage",
```
U-lis, line 41: AFAIK, we have `NCG` and `CRYSTAL` for now.
Is it ok to name `Garage` not `GARAGE`?
I think all capital ticker is good to recognize and can reduce mistakes.

Lib9c/Model/Garages/FungibleItemGarage.cs
```diff
@@ -0,0 +1,111 @@
...
+
+        public IValue Serialize() => Count == 0
+            ? (IValue)Null.Value
+            : new List(
+                Item?.Serialize() ?? Null.Value,
+                (Integer)Count);
+
+        public void Add(int count)
+        {
+            if (Count + count < 0)
```
U-lis, line 51: Is it possible for `count` to be negative? It seems a bit unusual to pass a negative value to the `Add` function, which is typically used for adding positive amounts.

Lib9c.Abstractions/ILoadIntoMyGarages.cs
```diff
@@ -0,0 +1,20 @@
...
+namespace Lib9c.Abstractions
+{
+    public interface ILoadIntoMyGarages
+    {
+        IOrderedEnumerable<(Address balanceAddr, FungibleAssetValue value)>? FungibleAssetValues
+        {
+            get;
+        }
+
+        Address? InventoryAddr { get; }
```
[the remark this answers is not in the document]
boscohyun: Good point. I considered using the avatar index for this part. Because it may be more convenient than using the avatar address.
However, if you look at only the action in isolation, the fact that you have to derive and use the inventory address from the avatar address requires additional information.
In this regard, I thought that what the action requires should be the inventory address itself, and it would be better for the side that creates this action to accept the avatar's inventory or address.

Lib9c/Action/Garages/DeliverToOthersGarages.cs
```diff
@@ -0,0 +1,265 @@
...
+using Libplanet.Action;
+using Libplanet.Assets;
+using Libplanet.State;
+using Nekoyume.Exceptions;
+using Nekoyume.Model.Garages;
+using Nekoyume.Model.State;
+
+namespace Nekoyume.Action.Garages
+{
+    [ActionType("deliver_to_others_garages")]
```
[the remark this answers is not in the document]
boscohyun: Good point. I thought about how well the name of the action should indicate the function.
Agents can actually have many garages. This is because individual garages are used depending on the type of currency or item.
So the `DeliverToOthersGarages` action can change the state of the other agent's various garages, which is reflected in the action's name.

Shouldn't the name of an action reflect the specifics of the state it changes? Conversely, if you omit the `s`, it is likely to create a misunderstanding that only one garage state is changed.

I think like this. If the garage is written as a function module rather than a state, and the unit of state stored by currency or item is a box rather than a garage, it would be more natural to subtract `s` from the action name. What is this direction like?

Lib9c/Action/Garages/UnloadFromMyGarages.cs
boscohyun: You are right. Thanks!

Lib9c/Currencies.cs
boscohyun: Oh I didn't think so. I'll capitalize it for consistency. Thanks!

Lib9c/Model/Garages/FungibleItemGarage.cs
boscohyun: It was written simply to provide the addition operation in the class.
If you have a good name, please suggest it.
Or would it be better to change it to `SetCount(int value)` to do the summation outside?

U-lis: I simply thought form view of garage. On the garage's side, there are only `in` and `out` action makes change storage amount.
So If you don't have special meaning for the name `Add()` and `TransferTo()`, how about `Inbound()` and `Outbound()`? (those names are recommended by ChatGPT haha)

Lib9c.Abstractions/ILoadIntoMyGarages.cs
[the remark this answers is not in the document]
U-lis: Hmm :thinking: that makes sense. OK I'll keep this in mind and reflect that to IAP backend service.
Maybe some extra effort will be needed to answer `who's this?` but I understood what's on focus.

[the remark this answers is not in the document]
boscohyun: How about using the `Memo` property like `TransferAsset` action.

Lib9c/Model/Garages/FungibleItemGarage.cs
boscohyun: Oh, that's nice. Or... `Load()` and `Unload()`? And maybe replace `TransferTo()` with `DeliverTo()`?
I'll think about it a bit more when I get back to the review! Thanks!

Lib9c.Abstractions/ILoadIntoMyGarages.cs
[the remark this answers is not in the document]
boscohyun: I applied the `Memo` idea and you can figure out at #1959.

Lib9c/Model/Garages/FungibleItemGarage.cs
boscohyun: I applied `Load()`, `Deliver()`, `Unload()` idea and you can figure out at #1959.

boscohyun:
Replace with #1959.

PR #1954, opened 2023-06-12
Set GARAGE currency as a reword of the monster collection system

boscohyun:
- Introduce `GARAGE` currency.
- Introduce `Currencies.GetMinterlessCurrency(ticker)` method.
- Introduce `ActionObsoleteConfig.V200030ObsoleteIndex`.(7,200,000)
- Obsolete `stake`, `claim_stake_reward` action.
- Expand `StakeRegularRewardSheet` class and `claim_stake_reward3` action.


Lib9c/Model/State/StakeState.cs
```diff
@@ -54,6 +54,7 @@ public void Achieve(int level, int step)
         public const long RewardInterval = 50400;
         public const long LockupInterval = 50400 * 4;
         public const long StakeRewardSheetV2Index = 6_700_000L;
+        public const long CurrencyAsRewardStartIndex = 6_910_000L;
```
moreal, line 57: I cannot find the source of this index. If this is new, it should be bigger than current mainnet index (6974002).

moreal (changes requested):

Lib9c/Model/State/StakeState.cs
boscohyun: My understanding is that the context in which this value is used has not yet been executed.🤔

Lib9c/Model/State/StakeState.cs
```diff
@@ -154,7 +155,27 @@ public int CalculateAccumulatedRuneRewards(long blockIndex, out int v1Step, out
...
+            var targetBlockIndex = Math.Max(
+                Math.Max(StartedBlockIndex, ReceivedBlockIndex),
+                CurrencyAsRewardStartIndex);
+            v1Step = 0;
+            v2Step = (int)Math.DivRem(
+                blockIndex - targetBlockIndex,
+                RewardInterval,
+                out _
+            );
+            return v1Step + v2Step;
```
moreal, line 172: It seems there is no step interval check (like `CalculateStep` does). If the `RewardInterval` is `100` and `CurrencyAsRewardStartIndex` is `0`, it will receive 1 the first time and 2 the second time, 3 the third time, 6 at all 🤔

boscohyun: Ah, `ClaimStakeReward` is updating `StakeState.ReceivedBlockIndex`, so it should be fine.😉

.Lib9c.Tests/Model/State/StakeStateTest.cs
```diff
@@ -70,5 +70,56 @@ public void SerializeV2()
...
+            var addr = new PrivateKey().ToAddress();
+            var stakeState = new StakeState(
+                addr,
+                startedBlockIndex);
+            Assert.Equal(addr, stakeState.address);
+            Assert.Equal(startedBlockIndex, stakeState.StartedBlockIndex);
+            Assert.Equal(0, stakeState.ReceivedBlockIndex);
+            Assert.Equal(
+                startedBlockIndex + LockupInterval,
+                stakeState.CancellableBlockIndex);
```
moreal, line 118: They don't seem to be what this test is looking for.

.Lib9c.Tests/Model/State/StakeStateTest.cs
```diff
@@ -70,5 +70,56 @@ public void SerializeV2()
...
+        [InlineData(0L, 0L, RewardInterval * 9, RewardInterval + 1L, 7)]
+        // control complex. reward start block index is inside of reward range.
+        [InlineData(1L, 0L, RewardInterval * 9, RewardInterval, 7)]
+        [InlineData(1L, 0L, RewardInterval * 9 + 1L, RewardInterval, 8)]
+        [InlineData(RewardInterval, 0L, RewardInterval * 9, RewardInterval, 8)]
+        // reward start block index is greater than reward range.
+        [InlineData(0L, 0L, RewardInterval * 9, RewardInterval * 9 + 1L, 0)]
+        // reward start block index is less than reward range.
+        [InlineData(1L, 0L, RewardInterval * 9 + 1L, 0L, 9)]
+        public void GetRewardStep(
```
moreal, line 102: I think it will be better to test the public method `StakeState.CalculateAccumulatedCurrencyRewards` instead of the internal method. 🤔 But it is not critical 🤔

Lib9c/Model/State/StakeState.cs
```diff
@@ -160,18 +160,90 @@ public int CalculateAccumulatedRuneRewards(long blockIndex, out int v1Step, out
...
+                    .ToList();
+                if (stepBlockIndexes.Count == 0)
+                {
+                    return 0;
+                }
+            }
+
+            if (rewardStartBlockIndex.HasValue)
+            {
+                var validStepBlockIndex = rewardStartBlockIndex + RewardInterval;
```
moreal, line 243: I wonder if the below case is intended. In other words, what means `rewardStartBlockIndex`?

```python
# Python
RewardInterval = 100
rewardStartBlockIndex = 250
StartedBlockIndex = 0
ReceivedBlockIndex = 100
currentBlockIndex = 300
stepBlockIndexes = []

# After loop
stepBlockIndexes = [100, 200, 300]

# After "ReceivedBlockIndex > 0" condition
stepBlockIndexes = [200, 300]

# After "rewardStartBlockIndex.HasValue" condition if statement
validStepBlockIndex = rewardStartBlockIndex + RewardInterval  # 350
stepBlockIndexes = [x for x in stepBlockIndexes if x >= validStepBlockIndex]. # []
```

moreal: Confirmed by the game designer.

moreal (approved):

.Lib9c.Tests/Model/State/StakeStateTest.cs
boscohyun: Yes it is. I'll separate this.

.Lib9c.Tests/Model/State/StakeStateTest.cs
boscohyun: I wanted to use the GetRewardStep() method to see the return value based on the rewardStartBlockIndex argument.
It's true that I didn't add more tests for the public method.😅⏰

PR #1959, opened 2023-06-15
Introduce `LoadIntoMyGarages` action

boscohyun:
- Introduce `LoadIntoMyGaragesCostSheet`.
- Introduce `FungibleItemGarage`.
- Introduce `LoadIntoMyGarages` action.

### DIFF
--- a/Lib9c/Model/Item/Inventory.cs
+++ b/Lib9c/Model/Item/Inventory.cs
@@ -275,6 +275,49 @@ namespace Nekoyume.Model.Item
             return nonFungibleItem;
         }
 
+        /// <summary>
+        /// Add tradable fungible item.
+        /// If there is same item which is not locked, add count to it.
+        /// The same item is item which has same item sheet id, fungible id, tradable id and
+        /// same required block index.
+        /// If there is no same item, add new item.
+        /// </summary>
+        /// <exception cref="ArgumentNullException" />
+        /// <exception cref="ArgumentOutOfRangeException" />
+        public Item AddTradableFungibleItem(
+            ITradableFungibleItem tradableFungibleItem,
+            int count)
+        {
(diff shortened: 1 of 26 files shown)


Lib9c/Model/Item/Inventory.cs
```diff
@@ -275,6 +275,49 @@ public Item AddNonFungibleItem(ItemBase itemBase, ILock iLock = null)
...
+            item = outItems.First();
+            if (item.count > int.MaxValue - count)
+            {
+                throw new ArgumentOutOfRangeException(
+                    nameof(count),
+                    count,
+                    $"Aborted because {nameof(item.count)} is overflow");
+            }
+
+            item.count += count;
```
U-lis, line 317: I'm not sure we've validated `item.count` must be positive value before.

Lib9c/TableCSV/Garages/LoadIntoMyGaragesCostSheet.csv
```diff
@@ -0,0 +1,11 @@
+id,_memo,cost_kind,currency_ticker,item_id,garage_cost_per_unit
+_id,int
+_cost_kind,CURRENCY|ITEM
+_currency_ticker,(empty)|NCG|CRYSTAL...
+_item_id,(empty)|400000|500000...
+_garage_cost_per_unit,1|100|0.001...
```
U-lis, line 6: Is it OK to put extra info into plain CSV?

Lib9c/Model/Item/Inventory.cs
boscohyun: As far as I understand, inventory items cannot have a count less than 1.🤔

U-lis (approved):
LGTM

PR #1960, opened 2023-06-16
Introduce mead

ipdae:
merge [mead branch](https://github.com/planetarium/lib9c/tree/test/mead) in development

- check admin permission in `PrepareRewardAssets, CreatePledge`
- update mainnet patron address
- delete MeadConfig temp blockIndex & use V200030ObsoleteIndex


Lib9c/Action/RequestPledge.cs
```diff
@@ -0,0 +1,58 @@
...
+    {
+        public const string TypeIdentifier = "request_pledge";
+
+        public RequestPledge()
+        {
+        }
+
+        // Value from tx per block policy.
+        // https://github.com/planetarium/lib9c/blob/b6c1e85abc0b93347dae8e1a12aaefd767b27632/Lib9c.Policy/Policy/MaxTransactionsPerSignerPerBlockPolicy.cs#L29
+        public const int RefillMead = 4;
```
longfin, line 21: It might be too late but how about allowing this value as configurable?

Lib9c/Action/RequestPledge.cs
```diff
@@ -0,0 +1,58 @@
...
+        public Address AgentAddress;
+        public int Mead;
```
longfin, line 23: This field doesn't seem to be used.

Lib9c/Action/RewardGold.cs
```diff
@@ -296,5 +299,34 @@ public IAccountStateDelta MinerReward(IActionContext ctx, IAccountStateDelta sta
 
             return states;
         }
+
+        public static IAccountStateDelta TransferMead(IAccountStateDelta states)
+        {
+#pragma warning disable LAA1002
+            var targetAddresses = states
+                .UpdatedFungibleAssets
```
longfin, line 307:
```suggestion
                .TotalUpdatedFungibles
```

longfin (changes requested):

Lib9c/Action/RequestPledge.cs
ipdae: this field using configure refill mead value. i'll rename `RefillMead`

Lib9c/Action/RequestPledge.cs
ipdae: i'll rename `DefaultRefillMead`

longfin (approved):

PR #1970, opened 2023-06-20
⬆️ Bump libplanet to 2.1.0

greymistcube:
⬆️ 
`IncompleteBlockStatesException` has been removed from libplanet. I'm not sure the old patchwork made in https://github.com/planetarium/lib9c/pull/151#discussion_r506039478 is still relevant. 🤨

### DIFF
--- a/Lib9c.Policy/Policy/BlockPolicySource.Utils.cs
+++ b/Lib9c.Policy/Policy/BlockPolicySource.Utils.cs
@@ -74,16 +74,9 @@ namespace Nekoyume.Blockchain.Policy
 
         internal static AdminState GetAdminState(BlockChain blockChain)
         {
-            try
-            {
-                return blockChain.GetState(AdminState.Address) is Dictionary rawAdmin
-                    ? new AdminState(rawAdmin)
-                    : null;
-            }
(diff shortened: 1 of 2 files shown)


longfin:
> ⬆️ `IncompleteBlockStatesException` has been removed from libplanet. I'm not sure the old patchwork made in [#151 (comment)](https://github.com/planetarium/lib9c/pull/151#discussion_r506039478) is still relevant. 🤨

Basically, it might be okay since we replaced confirming activation by just checking the gas price.
 
@greymistcube Just to clarify, if there was an `IncompletedStateException` before, how does it work now?

longfin (approved):

greymistcube:
https://github.com/planetarium/libplanet/blob/4bf65f7ffc823f2b934f4a8e9af32ab42b8c38ea/Libplanet/Blockchain/BlockChainStates.cs#L69-L96

@longfin The relevant piece of code is as above. It throws a simple `ArgumentException` instead. Since `PreloadAsync()` has changed, I don't think policy check will be invoked without actually invoking `Append()` (or more broadly without its previous block already existing in storage) anymore. If that is the case, this should be fine, but I'm not so sure. 😐

PR #1974, opened 2023-06-21
Update Item Upgrade cost Change

NineSyu:
Update Item Upgrade Cost Change
- EnhancementCostSheetV2


tyrosine1153 (approved):

Unengine (changes requested):
Tests are failed. I think we should make this PR draft and make it ready when test codes are fixed.

PR #1981, opened 2023-06-27
Add new column IsEnhanceable to StatBuffSheet

tyrosine1153:
+ Add new Buffs 800001, 800002

please refer StatBuffSheet.cs diff


Unengine (approved):

Lib9c/TableData/Skill/StatBuffSheet.cs
```diff
@@ -49,6 +51,12 @@ public override void Set(IReadOnlyList<string> fields)
                 StatType = (StatType)Enum.Parse(typeof(StatType), fields[5]);
                 OperationType = (StatModifier.OperationType)Enum.Parse(typeof(StatModifier.OperationType), fields[6]);
                 Value = ParseInt(fields[7]);
+
+                if (fields.Count > 8)
+                {
+                    IsEnhanceable = !bool.TryParse(fields[8], out var isEnhanceable) || isEnhanceable;
+                    // bool.TryParse(fields[8], out var isEnhanceable) ? isEnhanceable : true
+                }
```
boscohyun, line 59: How about below?

```suggestion
                if (fields.Count == 8)
                {
                    IsEnhanceable = true;
                    return;
                }
                
                IsEnhanceable = bool.TryParse(fields[8], out var isEnhanceable)
                    ? isEnhanceable
                    : false;
```

boscohyun (approved):

PR #1995, opened 2023-07-04
♻️ Refactor test classes

greymistcube:
As there are `IAccountStateDelta`, `IAccountState`, and `IAccountDelta` as separate interfaces, the name of `State` class has been changed to `StateDelta` to avoid future confusion. This is mainly a preparatory work to smooth over introducing a different and new `State` class in a follow-up PR.

----

Addendum: On second thought, to be safe, I'll be naming things `MockStateDelta`, `MockDelta`, and `MockState` to avoid mistakes.

### DIFF
--- a/.Lib9c.Tests/Action/BattleArena5Test.cs
+++ b/.Lib9c.Tests/Action/BattleArena5Test.cs
@@ -45,7 +45,7 @@ namespace Lib9c.Tests.Action
                 .WriteTo.TestOutput(outputHelper)
                 .CreateLogger();
 
-            _initialStates = new State();
+            _initialStates = new MockStateDelta();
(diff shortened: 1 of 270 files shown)


.Lib9c.Tests/Action/BattleArena5Test.cs
```diff
@@ -45,7 +45,7 @@ public BattleArena5Test(ITestOutputHelper outputHelper)
                 .WriteTo.TestOutput(outputHelper)
                 .CreateLogger();
 
-            _initialStates = new State();
```
riemannulus, line 48: 👀

riemannulus (approved):

PR #1996, opened 2023-07-04
✨ Added `MockState` for testing

greymistcube:
- ~~`IAction.Execute()` should now run with clean `IAccountStateDelta`s~~
- It was allowed to add FAVs that didn't make sense. This is no longer the case.
  - For example, it was possible to "add" `(addressA, currencyB, favC)` where `favC.Currency` was not `currencyB`.
 
----

Addemdum: The first point isn't true yet. I'll bring `MockStateDelta` up to date with a separate PR.


boscohyun (approved):

greymistcube:
Eh, seems like I need to fix a bit more stuff. 😥

PR #2000, opened 2023-07-07
Added optional parameter for `MockState` mutations

greymistcube:
Allows

```csharp
MockState state = MockState.Empty;
if (condition)
{
    state.SetState(addressA, valueA);
}
else
{
    state.SetState(addressB, valueB);
}
```

to be written as

```csharp
MockState state = MockState.Empty
    .SetState(addressA, valueA, condition)
    .SetState(addressB, valueB, !condition);
```

I suppose there is some potential for abuse somewhere in there, but since it's only a mock class for testing, I've decided "eh, whatever". 🙄

### DIFF
--- a/.Lib9c.Tests/Action/MockState.cs
+++ b/.Lib9c.Tests/Action/MockState.cs
@@ -47,6 +47,10 @@ namespace Lib9c.Tests.Action
     ///         </list>
     ///     </description></item>
     /// </list>
+    /// Additionally, all mutating method accepts optional <see langword="bool"/> argument, set to
+    /// <see langword="true"/> as default, whether to actually run the method or not.
+    /// This is purely for syntactic purpose to help avoid if-else branching of code at
+    /// a higher level.
     /// </remarks>
     public class MockState : IAccountState
     {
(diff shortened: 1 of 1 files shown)


greymistcube:
#2000, yay? 😛

boscohyun:
Wow.. I'm not sure I'd use it.
I'm under the impression that using `IAccountStateDelta.SetState(some, params)' as a function to store state and reading it is a long time in the past, so putting a condition as the last argument would make the code more likely to be misread.

```cs
MockState state = MockState.Empty;
if (condition)
{
    state.SetState(addressA, valueA);
}
else
{
    state.SetState(addressB, valueB);
}
```

```cs
var state = condition
    ? MockState.Empty.SetState(addressA, valueA)
    : MockState.Empty.SetState(addressB, valueB);
```

I also prefer to use this trinomial representation.

greymistcube:
I agree, it's very abusable API and not that intuitive. 😕
I was thinking more of complex conditionals. Single `bool` condition can be turned into a ternary statement, but you can't chain it further. I suppose its probably not really worth it.

PR #2018, opened 2023-07-14
Optimize HackAndSlash.Execute #1

ipdae:
image

image

reduce Simulator.Simulate() from 123ms -> 39ms

- Optimize seriialization
- Reduce call duplicate object in loop
- Improve speed HitHelper, AttackCountHelper
- Disable client replay logic(skillInfo) in Execute

### DIFF
--- a/Lib9c/Battle/HitHelper.cs
+++ b/Lib9c/Battle/HitHelper.cs
@@ -64,101 +64,52 @@ namespace Nekoyume.Battle
 
         public static int GetHitStep1(int attackerLevel, int defenderLevel)
         {
-            var correction = 0;
             var diff = attackerLevel - defenderLevel;
 
             if (diff <= GetHitStep1LevelDiffMin)
             {
-                correction = GetHitStep1CorrectionMin;
-            }
-            else if (diff >= GetHitStep1LevelDiffMax)
-            {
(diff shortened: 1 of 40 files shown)


Lib9c/Battle/HitHelper.cs
```diff
@@ -64,101 +64,52 @@ public static class HitHelper
...
+                _ => 0
+            };
         }
 
         public static int GetHitStep2(int attackerHit, int defenderHit)
         {
             attackerHit = Math.Max(1, attackerHit);
             defenderHit = Math.Max(1, defenderHit);
-            var additionalCorrection = (int) ((attackerHit - defenderHit / 3m) / defenderHit * 100);
+            var additionalCorrection = (attackerHit * 100 - defenderHit * 100 / 3) / defenderHit;
```
boscohyun, line 112: This change is not backwards compatible.

```cs
[Fact]
public void Test()
{
    const int attackerHit = 1;
    const int defenderHit = 1;
    var result1 = (int)((attackerHit - defenderHit / 3m) / defenderHit * 100); // 66
    var result2 = (attackerHit * 100 - defenderHit * 100 / 3) / defenderHit; // 67
    Assert.Equal(result1, result2); // failed. 
}
```

boscohyun (approved):

PR #2071, opened 2023-08-10
feat: expand staking rewards

boscohyun:
# Changes

### Expand `StakeRegularRewardSheet.RewardInfo`

- Deprecate `int Rate` field.
  - This field is used from `ClaimStakeReward.V2` or earlier. Use `DecimalRate` instead.
- Introduce `int? CurrencyDecimalPlaces` field.
- Introduce `decimal? DecimalRate` field.
  - This field is used from `V3`(in blockchain state) or later.
  - This field is set to `Rate` when `ClaimStakeReward.V2` or earlier.

### Upgrade `ClaimStakeReward` action from 5 to 6

- Deprecate `claim_stake_reward5` at `ActionObsoleteConfig.V200060ObsoleteIndex`.
- Hard code a couple of csv strings in blockchain state as `ClaimStakeReward.V2`.
- Apply to `ClaimStakeRewardFactory`.

### Etc.

- (test)Improve `InitializeUtil`.
- (asset)Update `StakeRegularRewardSheet.csv`.

---
  - https://app.asana.com/0/0/1205238116149028


Lib9c/Action/ClaimStakeReward.cs
```diff
@@ -247,81 +349,138 @@ public override IAccountStateDelta Execute(IActionContext context)
...
                         states = states.MintAsset(
                             context,
                             context.Signer,
-                            rewardCurrencyQuantity * currencyRewardStep * rewardCurrency);
+                            rewardFav);
                         break;
+                    }
                     default:
-                        break;
+                        throw new ArgumentOutOfRangeException();
```
longfin, line 473:
```suggestion
                        throw new ArgumentException(nameof(regularReward), $"Can't handle reward type: {reward.Type}");
```

Lib9c/Extensions/DecimalExtensions.cs
```diff
@@ -0,0 +1,22 @@
...
+            var majorUnit = decimal.Parse(strings[0], CultureInfo.InvariantCulture);
+            if (strings.Length == 1)
+            {
+                return (new BigInteger(majorUnit), BigInteger.Zero);
+            }
+
+            var minorUnit = decimal.Parse(strings[1], CultureInfo.InvariantCulture);
+            return (new BigInteger(majorUnit), new BigInteger(minorUnit));
+        }
+    }
```
longfin, line 21: Major/Minor unit concept seems too specific concepts about Libplanet's FAV. thus how about returning `FungibleAssetValue` with `Currency` here?

```suggestion
        public static FungibleAssetValue ToFungibleAssetValue(this decimal value, Currency currency)
        {
            byte decimalPlaces = currency.decimalPlaces;
            decimal rounded = Math.Round(value, decimalPlaces, MidpointRounding.ToZero); // or need to receive MidpointRounding instead.
            BigInteger majorUnit = (BigInteger)Math.Floor(rounded);
            BigInteger minorUnit = (BigInteger)Math.Floor((rounded - majorUnit) * (decimal)BigInteger.Pow(10, decimalPlaces));

            return new FungibleAssetValue(currency, majorUnit, minorUnit);
        }
   }
```

Lib9c/Action/ClaimStakeReward.cs
```diff
@@ -247,81 +349,138 @@ public override IAccountStateDelta Execute(IActionContext context)
...
+                    throw new NullReferenceException(
+                        "decimal rate is null." +
+                        $" \"{ActionTypeText}\" action requires to set decimal rate.");
+                }
+
+                var stakedQuantity = decimal.Parse(
+                    stakedFav.GetQuantityString(),
+                    CultureInfo.InvariantCulture);
+                var rewardQuantityForSingleStep =
+                    Math.Truncate(stakedQuantity / reward.DecimalRate.Value);
```
longfin, line 388: I think that converting `reward.DecimalRate` to FAV with proper currency (atm NCG) would be better than converting FAV to decimal...

longfin (approved):
I've made some comments about the style and structure, but they aren't deal breakers since I haven't found any fatal flaws yet, and I'm preparing another proposal for this action.

Lib9c/TableData/StakeRegularRewardSheet.cs
```diff
@@ -29,36 +36,75 @@ public class RewardInfo
...
+            /// This field is only used when <see cref="Type"/> is <see cref="StakeRewardType.Currency"/>.
+            /// </summary>
+            public readonly int? CurrencyDecimalPlaces;
+
+            /// <summary>
+            /// The decimal rate of reward.
+            /// This field is used from `V3`(in blockchain state) or later.
+            /// This field is set to `Rate` when `ClaimStakeReward.V2` or earlier.
+            /// </summary>
+            public readonly decimal? DecimalRate;
```
longfin, line 50:
```suggestion
            public readonly decimal DecimalRate;
```

Lib9c/TableData/StakeRegularRewardSheet.cs
```diff
@@ -29,36 +36,75 @@ public class RewardInfo
...
                 {
                     CurrencyTicker = null;
+                    CurrencyDecimalPlaces = null;
+                    DecimalRate = Rate;
                     return;
                 }
 
                 CurrencyTicker = fields[3];
+
+                if (fields.Length == 4)
```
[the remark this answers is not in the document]
longfin: Ah, I understood that. it seems extension fields parser and... I think that we might need a versioning scheme for these sheets... 😢

Lib9c/Extensions/DecimalExtensions.cs
boscohyun: I've fix some code where to use this extension method and it can be remove.

PR #2078, opened 2023-08-18
Release/0.5.0

area363:
This is a backmerge before tagging.


area363:
Ignore nuget release for now.

PR #2132, opened 2023-09-06
chore: work a few unit tests

boscohyun:
SSIA

### DIFF
--- a/.Lib9c.Tests/Model/Stake/ContractTest.cs
+++ b/.Lib9c.Tests/Model/Stake/ContractTest.cs
@@ -118,6 +118,12 @@ namespace Lib9c.Tests.Model.Stake
                 lockupInterval));
         }
 
+        [Fact]
+        public void Constructor_With_StakePolicySheet_Throw_ArgumentNullException()
+        {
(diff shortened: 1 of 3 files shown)


.Lib9c.Tests/Model/Stake/ContractTest.cs
```diff
@@ -118,10 +118,16 @@ public class ContractTest
                 lockupInterval));
         }
 
+        [Fact]
+        public void Constructor_With_StakePolicySheet_Throw_ArgumentNullException()
+        {
+            Assert.Throws<ArgumentNullException>(() => new Contract((StakePolicySheet)null));
+        }
```
moreal, line 125: Since Unity used by NineChronicles project supports C# 9, it should use `nullable` checker too 🤔 Of course, it is good!

moreal (approved):

.Lib9c.Tests/Model/Stake/ContractTest.cs
boscohyun: Does this mean that I should enable the nullable reference type in the Contract class, and make sure to use the StakePolicySheet (non nullable) type when using the above constructor?

PR #2148, opened 2023-09-14
Add ClaimItems action

XxshiftxX:


Lib9c/Action/ClaimItem.cs
```diff
@@ -0,0 +1,94 @@
...
+        protected override void LoadPlainValueInternal(
+            IImmutableDictionary<string, IValue> plainValue)
+        {
+            AvatarAddress = plainValue[AvatarAddressKey].ToAddress();
+            Amount = plainValue[AmountKey].ToFungibleAssetValue();
+        }
+
+        public override IAccount Execute(IActionContext context)
+        {
+            var states = context.PreviousState;
```
boscohyun, line 45:
```suggestion
            context.UseGas(1);
            var states = context.PreviousState;
```
😉

moreal:
As a note, it needs a benchmark code to know how many execution times this action needs.

Lib9c/Action/ClaimItems.cs
```diff
@@ -0,0 +1,105 @@
...
+            }).GetItemSheet();
+
+            foreach (var fungibleAssetValue in Amounts)
+            {
+                var ticker = fungibleAssetValue.Currency.Ticker;
+                if (!ticker.StartsWith("it_") ||
+                    !int.TryParse(ticker.Replace("it_", string.Empty), out var itemId))
+                {
+                    throw new ArgumentException($"Format of Amount currency's ticker is invalid");
+                }
```
boscohyun, line 77: We have been setting the ticker for `Currency` to capitalize only(e.g., `NCG`, `CRYSTAL`), is there a specification(e.g., NCIP) for this ticker?

Lib9c/Action/ClaimItems.cs
```diff
@@ -0,0 +1,105 @@
...
+
+                var item = itemSheet[itemId] switch
+                {
+                    MaterialItemSheet.Row materialRow =>
+                        ItemFactory.CreateTradableMaterial(materialRow),
+                    var itemRow => ItemFactory.CreateItem(itemRow, context.Random)
+                };
+
+                avatarState.inventory.AddItem(item, (int)fungibleAssetValue.RawValue);
```
boscohyun, line 95: If we use the `FungibleAssetValue.RawValue` as the count of item, it will be better to make sure the `FungibleAssetValue.Currency.DecimalPlacese` equals to 0.

Lib9c/Action/ClaimItems.cs
```diff
@@ -0,0 +1,105 @@
...
+
+                states = states
+                    .BurnAsset(context, context.Signer, fungibleAssetValue)
+                    .SetState(avatarState.address, avatarState.Serialize());
```
boscohyun, line 99: After we get the `AvatarState` via `IAccountState.TryGetAvatarStateV2()`, we should set states(AvatarState, Inventory, WorldInformation, QuestList) separately depends on `out bool migrateRequired`.

boscohyun (commented):
Nice your first action!🎉

Lib9c/Action/ClaimItems.cs
XxshiftxX: Nope. It's just temporary ticker. You can check it on draft NCIP by swen.

Lib9c/Action/ClaimItems.cs
```diff
@@ -0,0 +1,105 @@
...
+
+            var avatarState = states.GetAvatarState(AvatarAddress);
+            if (avatarState is null)
+            {
+                throw new FailedLoadStateException(
+                    ActionTypeText,
+                    addressesHex,
+                    typeof(AvatarState),
+                    AvatarAddress);
+            }
```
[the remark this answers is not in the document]
XxshiftxX: At first, I tried this way to get avatarState: `TryGetAvatarStateV2`. But we can't use signer address as agentAddress because the signer and avatarAddress can be different in this case. Is there some way to get agentAddress with avatarAddress?

Lib9c/Action/ClaimItems.cs
```diff
@@ -0,0 +1,113 @@
...
+                foreach (var inventory in inventories)
+                {
+                    inventory.AddItem(item, (int)fungibleAssetValue.RawValue);
+                }
+                states = states.BurnAsset(context, context.Signer, fungibleAssetValue * inventories.Count);
+            }
+
+            foreach (var (avatarAddress, i) in AvatarAddresses.Select((x, i) => (x, i)))
+            {
+                states = states.SetState(avatarAddress.Derive(LegacyInventoryKey), inventories[i].Serialize());
```
moreal, line 107: `avatarAddress.Derive(LegacyInventoryKey)` seems used twice. How about caching the addresses or making a function to derive inventory address in same way.

Lib9c/Action/ClaimItems.cs
```diff
@@ -0,0 +1,113 @@
...
+                    MaterialItemSheet.Row materialRow =>
+                        ItemFactory.CreateTradableMaterial(materialRow),
+                    var itemRow => ItemFactory.CreateItem(itemRow, context.Random)
+                };
+
+                foreach (var inventory in inventories)
+                {
+                    inventory.AddItem(item, (int)fungibleAssetValue.RawValue);
+                }
+                states = states.BurnAsset(context, context.Signer, fungibleAssetValue * inventories.Count);
```
moreal, line 102: 🧹 

```suggestion
                }

                states = states.BurnAsset(context, context.Signer, fungibleAssetValue * inventories.Count);
```

Lib9c/Action/ClaimItems.cs
```diff
@@ -0,0 +1,113 @@
...
+            var states = context.PreviousState;
+            var itemSheet = states.GetSheets(containItemSheet: true).GetItemSheet();
+
+            var inventories = AvatarAddresses.Select(address =>
+            {
+                var addressHex = GetSignerAndOtherAddressesHex(context, address);
+                var inventory = states.GetInventory(address.Derive(LegacyInventoryKey));
+                if (inventory is null)
+                {
+                    throw new FailedLoadStateException(ActionTypeText, addressHex, typeof(AvatarState), address);
```
moreal, line 59: Since the third argument is state's type, I think it should be `Inventory` instead of `AvatarState`.

Lib9c/Action/ClaimItems.cs
```diff
@@ -0,0 +1,113 @@
...
+                }
+
+                var balance = states.GetBalance(context.Signer, fungibleAssetValue.Currency);
+                if (balance < fungibleAssetValue)
+                {
+                    throw new NotEnoughFungibleAssetValueException(
+                        context.Signer.ToHex(),
+                        fungibleAssetValue.RawValue,
+                        balance);
+                }
```
moreal, line 89: Because it should burn `fungibleAssetValue * inventories.Count`, this condition should check with the same amount.

```
                states = states.BurnAsset(context, context.Signer, fungibleAssetValue * inventories.Count);
```

Lib9c/Action/ClaimItems.cs
```diff
@@ -0,0 +1,114 @@
...
```
boscohyun, line 89: How about this? We just burn the asset here.
```suggestion
                states = states.BurnAsset(context, context.Signer, fungibleAssetValue * inventories.Count);
```

Lib9c/Action/ClaimItems.cs
```diff
@@ -0,0 +1,110 @@
...
+
+                var balance = states.GetBalance(context.Signer, fungibleAssetValue.Currency);
+                if (balance < fungibleAssetValue * inventories.Count)
+                {
+                    throw new NotEnoughFungibleAssetValueException(
+                        context.Signer.ToHex(),
+                        fungibleAssetValue.RawValue,
+                        balance);
+                }
```
boscohyun, line 85: Yes we can!
```suggestion
                states = states.BurnAsset(context, context.Signer, fungibleAssetValue * inventories.Count);
```

.Lib9c.Tests/Action/ClaimItemsTest.cs
```diff
@@ -0,0 +1,179 @@
...
+    using Libplanet.Types.Assets;
+    using Nekoyume;
+    using Nekoyume.Action;
+    using Nekoyume.Model;
+    using Nekoyume.Model.State;
+    using Serilog;
+    using Xunit;
+    using Xunit.Abstractions;
+
+    public class ClaimItemsTest
```
moreal, line 17: Can it test `ClaimItems` action's serialization too?

Lib9c/Action/ClaimItems.cs
```diff
@@ -0,0 +1,101 @@
...
+
+                var decimalPlaces = fungibleAssetValue.Currency.DecimalPlaces;
+                if (decimalPlaces != 0)
+                {
+                    throw new ArgumentException(
+                        "DecimalPlaces of fungibleAssetValue for claimItems are not 0");
+                }
+
+                states = states.BurnAsset(context, context.Signer, fungibleAssetValue * inventories.Count);
```
moreal, line 78: In optimization, I think it can be processed in the front of all lines.

boscohyun: Wow, that's a tough approach.👍

Lib9c/Action/ClaimItems.cs
```diff
@@ -0,0 +1,134 @@
...
+using static Lib9c.SerializeKeys;
+
+namespace Nekoyume.Action
+{
+    public class ClaimItems : GameAction, IClaimItems
+    {
+        private const string ActionTypeText = "claim_items";
+
+        public IEnumerable<Address> AvatarAddresses { get; private set; }
+        public IEnumerable<FungibleAssetValue> Amounts { get; private set; }
```
moreal, line 25: I think it is not good to use  `IEnumerable<T>` as properties' type. Of course, it will be good for each case. 

`IEnumerable<T>` doesn't guarantee it returns the same value for every iteration.

https://dotnetfiddle.net/G3jdWx

```csharp
using System;
using System.Globalization;
using System.Collections.Generic;
using System.Linq;
					
public class Program
{
	public static void Main()
	{
		int i = 1;
		IEnumerable<string> enumerable = new string[] { "foo", "bar" }.Select(x => string.Concat(Enumerable.Repeat(x, i++)));
		foreach (var s in enumerable) {
			Console.WriteLine(s);
		}

		foreach (var s in enumerable) {
			Console.WriteLine(s);
		}
	}
}
```

The above example code's output...:

```
foo
barbar
foofoofoo
barbarbarbar
```

Lib9c/Action/ClaimItems.cs
```diff
@@ -0,0 +1,107 @@
...
+using static Lib9c.SerializeKeys;
+
+namespace Nekoyume.Action
+{
+    [ActionType(ActionTypeText)]
+    public class ClaimItems : GameAction, IClaimItems
+    {
+        private const string ActionTypeText = "claim_items";
+
+        public List<(Address address, List<FungibleAssetValue> fungibleAssetValues)> ClaimData { get; private set; }
```
moreal, line 23: In my thought, I want to suggest using `IReadOnlyList<T>` because it can be corrupted. (like JavaScript's `const` does.)

```javascript
const list = [];
list.push(111); // list != []
```

```csharp
var claimItems = ClaimItems();
claimItems.ClaimData.Add((new Address("..."), new List<FungibleAssetValue> { ... }));  // 🙀 🙀 🙀 🙀 🙀 🙀 
```

Lib9c/SerializeKeys.cs
```diff
@@ -174,5 +174,8 @@ public static class SerializeKeys
 
         // Grand Finale
         public const string GrandFinaleIdKey = "gfi";
+
+        // ClaimItem
```
moreal, line 178:
```suggestion
        // ClaimItems
```

Lib9c/Action/ClaimItems.cs
```diff
@@ -0,0 +1,107 @@
...
+                    {
+                        throw new ArgumentException(
+                            $"DecimalPlaces of fungibleAssetValue for claimItems are not 0: {fungibleAssetValue.Currency.Ticker}");
+                    }
+
+                    var parsedTicker = fungibleAssetValue.Currency.Ticker.Split("_");
+                    if (parsedTicker.Length != 3
+                        || parsedTicker[0] != "Item"
+                        || (parsedTicker[1] != "NT" && parsedTicker[1] != "T")
+                        || !int.TryParse(parsedTicker[2], out var itemId))
```
moreal, line 82: @longfin Could you leave your opinion about the item token's ticker? (because you're original NCIP-18's proposer)

moreal, line 82: I'll try to define a new type like `ItemTokenCurrency: Curreny` that validates `Currency.Ticker` when constructing instances, in working hours.

.Lib9c.Tests/Action/ClaimItemsTest.cs
```diff
@@ -0,0 +1,222 @@
...
+                action.ClaimData[0].fungibleAssetValues[0].Currency.Ticker,
+                deserialized.ClaimData[0].fungibleAssetValues[0].Currency.Ticker);
+            Assert.Equal(
+                action.ClaimData[0].fungibleAssetValues[1].Currency.Ticker,
+                deserialized.ClaimData[0].fungibleAssetValues[1].Currency.Ticker);
+
+            Assert.Equal(action.ClaimData[1].address, deserialized.ClaimData[1].address);
+            Assert.Equal(
+                action.ClaimData[1].fungibleAssetValues[0].Currency.Ticker,
+                deserialized.ClaimData[1].fungibleAssetValues[0].Currency.Ticker);
```
moreal, line 80: I guess it can directly check `FAV`s equality because FAV implement `IEquatable<T>` interface.

https://github.com/planetarium/libplanet/blob/7235c0f35fd8e586a91fbad87d43bc881610fb58/Libplanet.Types/Assets/FungibleAssetValue.cs#L25-L28

Lib9c/Action/ClaimItems.cs
XxshiftxX: I think I can handle this. Thank you for suggestion!

moreal: ***C# STRUCT CAN'T INHERIT STRUCT***

moreal (approved):

moreal (changes requested):
Because this action can have multiple tasks, it must have a window size to restrict the number of tasks. I guess 100 is enough. cc @planetarium/9c-dx

Lib9c/Action/ClaimItems.cs
longfin: LGTM

moreal (approved):

longfin (approved):

PR #2150, opened 2023-09-19
Use list for `AvatarState`

limebell:
And support automatic migration for old states.

### DIFF
--- a/Lib9c/Module/AgentModule.cs
+++ b/Lib9c/Module/AgentModule.cs
@@ -31,7 +31,8 @@ namespace Nekoyume.Module
                     return new AgentState(list);
                 }
 
-                return null;
+                throw new InvalidCastException(
+                    "Serialized agent state must be a dictionary or a list.");
             }
             catch (InvalidCastException e)
             {
--- a/Lib9c/Module/AgentModule.cs
+++ b/Lib9c/Module/AgentModule.cs
@@ -48,7 +48,7 @@ namespace Nekoyume.Module
 
         public static IWorld SetAgentState(IWorld world, Address agent, AgentState state)
         {
-            // TODO: Override legacy address to null state?
+            // TODO: Overwrite legacy address to null state?
             var account = world.GetAccount(Addresses.Agent);
             account = account.SetState(agent, state.SerializeList());
             return world.SetAccount(account);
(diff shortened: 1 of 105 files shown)


Lib9c/Module/AgentModule.cs
```diff
@@ -21,7 +22,16 @@ public static class AgentModule
...
+                if (serializedAgent is Dictionary dict)
+                {
+                    return new AgentState(dict);
+                }
+                else if (serializedAgent is List list)
+                {
+                    return new AgentState(list);
+                }
+
+                return null;
```
OnedgeLee, line 34: Throwing exception instead of returning null would be better for L34 IMHO.

Lib9c/Module/AvatarModule.cs
```diff
@@ -17,90 +17,98 @@ namespace Nekoyume.Module
...
-            try
-            {
-                enemyAvatarState = GetAvatarStateV2(worldState, avatarAddress);
-            }
-            // BackWard compatible.
-            catch (FailedLoadStateException)
-            {
-                enemyAvatarState = GetAvatarState(worldState, avatarAddress);
-            }
+            AvatarState enemyAvatarState = GetAvatarState(worldState, avatarAddress);
```
OnedgeLee, line 111: Seems weird for me, too.

Lib9c/Module/AvatarModule.cs
```diff
@@ -119,160 +127,44 @@ public static AvatarState GetEnemyAvatarState(IWorldState worldState, Address av
...
             Address avatarAddress,
-            out AgentState agentState,
-            out AvatarState avatarState,
-            out bool avatarMigrationRequired
-        )
+            AvatarState state,
+            bool setAvatar,
+            bool setInventory,
+            bool setWorldInformation,
+            bool setQuestList)
```
OnedgeLee, line 154: Actually, I think these boolean options are just used for tests. May be it's better to update tests to use the most recent version, and remove theses options, if we have enough time.

OnedgeLee (approved):

Lib9c/Module/AvatarModule.cs
OnedgeLee: Oops, my mistake. these have to be remained, to prevent values not updated to be set.

OnedgeLee: And these seems to be removed, since we're splitting those to other accounts.

PR #2169, opened 2023-10-06
Do not sort pairs

moreal:
Since `IValue` doesn't inherit `IComparable` interface, it cannot be used as a key to sort.

And I don't understand why they should be sorted 🤔.

### DIFF
--- a/Lib9c/Action/ClaimItems.cs
+++ b/Lib9c/Action/ClaimItems.cs
@@ -33,23 +33,24 @@ namespace Nekoyume.Action
 
         protected override IImmutableDictionary<string, IValue> PlainValueInternal =>
             ImmutableDictionary<string, IValue>.Empty
-                .Add(ClaimDataKey, ClaimData.Select(tuple =>
+                .Add(ClaimDataKey, ClaimData.Aggregate(List.Empty, (list, tuple) =>
                 {
                     var serializedFungibleAssetValues = tuple.fungibleAssetValues.Select(x => x.Serialize()).Serialize();
 
-                    return (tuple.address, serialized: serializedFungibleAssetValues);
-                }).Serialize());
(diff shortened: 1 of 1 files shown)


boscohyun (approved):

boscohyun:
In the context of using this extension method, I expected the enumeration passed as an argument to be serialized and stored in state, so I sorted it so that the result was deterministic.

moreal:
In inner communication, I and @boscohyun decided to recover the extension method sort correctly and to make the `ClaimItems` action not use the extension method. After the work, I'll request a review again.

moreal:
@boscohyun I'm trying to write a comparison function like the one below but it seems to take more time. I'll work for other tasks today. So I'll request a review to @XxshiftxX and merge this pull request by his approval.

```csharp
// WIP
 int BencodexValueComparison(IValue x, IValue y)
            {
                return (x, y) switch
                    {
                        (Null, Null) => 0,
                        (Binary binaryX, Binary binaryY) => new ByteArrayComparer().Compare(binaryX.ToByteArray(), binaryY.ToByteArray()),
                        (Text textX, Text textY) => StringComparer.Create(CultureInfo.InvariantCulture, false).Compare(textX.Value, textY.Value),
                        (Integer integerX, Integer integerY) => integerX.Value.CompareTo(integerY.Value),
                        (List listX, List listY) => listX.Count == listY.Count
                            ? listX.Zip(listY).A
                            : listX.Count - listY.Count,
                        (Dictionary dictionaryX, Dictionary dictionaryY) => 
                };
            }

            var result = new List();
            foreach (var v in value
                         .OrderBy(tuple => tuple.Item1)
                         .ThenBy(tuple => tuple.Item2, comparer: Comparer<IValue>.Create(BencodexValueComparison)))
```

XxshiftxX (approved):
LGTM!

PR #2174, opened 2023-10-10
refactor: use `GetRandom` instead of `Random`

riemannulus:
SSIA, for Unity


Lib9c/Action/CreateAvatar0.cs
```diff
@@ -199,7 +199,7 @@ public override IAccount Execute(IActionContext context)
             var consumableItemSheet = ctx.PreviousState.GetSheet<ConsumableItemSheet>();
             AddItemsForTest(
                 avatarState: avatarState,
-                random: ctx.Random,
+                random: ctx.GetRandom(),
```
greymistcube, line 202: As `ctx.GetRandom()` always returns an `IRandom` of the same state, `ctx.GetRandom()` should be extracted first in order to preserve the behavior. 😶

riemannulus: https://github.com/planetarium/lib9c/pull/2174/commits/630bc04bd209c7fb7d4ffeb5c1413e3f569ea67c amended

greymistcube (approved):

PR #2189, opened 2023-10-25
Introduce `Planet` and separate the action obsoletions by them

longfin:
---
  - https://app.asana.com/0/0/1205745461458878

### DIFF
--- a/Lib9c/Action/ActionObsoleteAttributeExtension.cs
+++ b/Lib9c/Action/ActionObsoleteAttributeExtension.cs
@@ -26,7 +26,7 @@ namespace Nekoyume
                 return attrs.Single().ObsoleteIndex + 2 <= blockIndex;
             }
 
-            return attrs.Any(attr => attr.ObsoleteIndex <= blockIndex && attr.AffectedPlanet == planet);
+            return attrs.Any(attr => (attr.AffectedPlanet == Planet.Odin ? attr.ObsoleteIndex + 2 : attr.ObsoleteIndex) <= blockIndex && attr.AffectedPlanet == planet);
         }
     }
 }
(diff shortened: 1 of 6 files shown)


Lib9c/Action/ActionObsoleteAttributeExtension.cs
```diff
@@ -0,0 +1,32 @@
...
+                {
+                    return true;
+                }
+
+                // Comparison with ObsoleteIndex + 2 is intended to have backward compatibility with
+                // a bugged original implementation.
+                return attrs.Single().ObsoleteIndex + 2 <= blockIndex;
+            }
+
+            return attrs.Any(attr => attr.ObsoleteIndex <= blockIndex && attr.AffectedPlanet == planet);
```
moreal, line 29: Does it need to check `ObsoleteIndex + 2 <= blockIndex` too when the `attr.AffectedPlanet` is `Planet.Odin`?

moreal (commented):
It needs follow-up PR.

Lib9c/Action/ActionObsoleteAttributeExtension.cs
longfin: AFIK, +2 offset was needed to aide previous bug and it doesn't be needed against new chains. am I understanding correctly? @greymistcube

moreal: Yes, `Planet.Heimdall` and `Heimdall.Idun` will not need the +2 offset. But I think `Planet.Odin` needs +2 offset as I said. Odin is not a new chain.

longfin:
```suggestion
            return attrs.Any(attr => (attr.AffectedPlaent == Planet.Odin ? attr.ObsoleteIndex + 2 : attr.ObsoleteIndex) <= blockIndex && attr.AffectedPlanet == planet);
```

@moreal Ah, I see. you mean this?

moreal: @longfin Ah yes!! That is what I mean.

moreal (approved):

PR #2190, opened 2023-10-25
Introduce MintAssets & BurnAssets

longfin:
---
  - https://app.asana.com/0/0/1205790778399752

### DIFF
--- a/Lib9c/Action/BurnAsset.cs
+++ b/Lib9c/Action/BurnAsset.cs
@@ -1,0 +1,99 @@
+using System.Collections.Generic;
+using Bencodex.Types;
+using Libplanet.Action;
+using Libplanet.Action.State;
+using Libplanet.Crypto;
+using Libplanet.Types.Assets;
+using Nekoyume.Exceptions;
+using Nekoyume.Model.State;
+
+namespace Nekoyume.Action
+{
+    [ActionType(TypeIdentifier)]
+    public class BurnAsset : ActionBase
+    {
+        private const int MemoMaxLength = 80;
+        public const string TypeIdentifier = "burn_asset";
+
+        private string _memo;
+
+        public BurnAsset()
+        {
+        }
+
+        public BurnAsset(Address owner, FungibleAssetValue amount, string memo)
+        {
+            Owner = owner;
+            Amount = amount;
+            Memo = memo;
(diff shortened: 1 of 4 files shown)


Lib9c/Action/MintAssets.cs
```diff
@@ -0,0 +1,140 @@
...
+                    Inventory inventory = state.GetInventory(inventoryAddr);
+                    MaterialItemSheet itemSheet = state.GetSheet<MaterialItemSheet>();
+                    if (itemSheet is null || itemSheet.OrderedList is null)
+                    {
+                        throw new InvalidOperationException();
+                    }
+
+                    foreach (MaterialItemSheet.Row row in itemSheet.OrderedList)
+                    {
+                        if (row.ItemId.Equals(items.Id))
```
ipdae, line 76:
```suggestion
                   var row = itemSheet[items.Id]
```

longfin: `items.Id` is `HashDigest<SHA256>` and it doesn't seem to be used for lookup like as `itemSheet[items.Id]` since `itemSheet[]` expects only `int`.

ipdae (approved):

Lib9c/Action/BurnAsset.cs
```diff
@@ -0,0 +1,99 @@
...
+        }
+
+        public BurnAsset(Address owner, FungibleAssetValue amount, string memo)
+        {
+            Owner = owner;
+            Amount = amount;
+            Memo = memo;
+        }
+
+        public override IValue PlainValue =>
```
XxshiftxX, line 31: How about move `PlainValue` to Above `LoadPlainValue` or move `LoadPlainValue` below `PlainValue`?
I think it's better to place deserialization/serialization codes in same place.

XxshiftxX (approved):

Lib9c/Action/BurnAsset.cs
longfin: I'll adjust them.

longfin: Applied. https://github.com/planetarium/lib9c/pull/2190/commits/6d32a1144180fc60deae216bba0cb75e5553732f

PR #2191, opened 2023-10-26
update acc access check logic

area363:
Use `GetAccessLevel` instead of `IsAccessDenied` to determine whether an address should be allowed or denied.

p.s. Headless PR coming up soon.

### DIFF
--- a/Lib9c.Policy/AccessControlService/IAccessControlService.cs
+++ b/Lib9c.Policy/AccessControlService/IAccessControlService.cs
@@ -5,5 +5,7 @@ namespace Nekoyume.Blockchain
     public interface IAccessControlService
     {
         public bool IsAccessDenied(Address address);
+
+        public int GetAccessLevel(Address address);
     }
 }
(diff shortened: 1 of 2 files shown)


Lib9c.Policy/AccessControlService/IAccessControlService.cs
```diff
@@ -5,5 +5,7 @@ namespace Nekoyume.Blockchain
     public interface IAccessControlService
     {
         public bool IsAccessDenied(Address address);
```
Atralupus, line 7: Is it going to be deprecated?

area363: Technically, yes. But, I wanted to preserve it just incase we might need to use it later 😅 Should we just remove it?

Atralupus: No... I agree with you, but we should write a deprecation comment for `IsAccessDenied`

area363:
Replaced by https://github.com/planetarium/lib9c/pull/2194.

PR #2194, opened 2023-10-31
add GetTxQuota to ACS

area363:
### Changes
- Add `GetTxQuota` method to `ACS` to check a signer's individual `txQuotaPerSigner` when staging txs.

Replaces https://github.com/planetarium/lib9c/pull/2191.

### DIFF
--- a/Lib9c.Policy/AccessControlService/IAccessControlService.cs
+++ b/Lib9c.Policy/AccessControlService/IAccessControlService.cs
@@ -4,7 +4,7 @@ namespace Nekoyume.Blockchain
 {
     public interface IAccessControlService
     {
-        public bool IsAccessDenied(Address address);
+        public bool IsListed(Address address);
 
         public int? GetTxQuota(Address address);
     }
--- a/Lib9c.Policy/AccessControlService/IAccessControlService.cs
+++ b/Lib9c.Policy/AccessControlService/IAccessControlService.cs
@@ -5,5 +5,7 @@ namespace Nekoyume.Blockchain
     public interface IAccessControlService
     {
         public bool IsAccessDenied(Address address);
+
+        public int? GetTxQuota(Address address);
     }
 }
--- a/Lib9c.Policy/AccessControlService/IAccessControlService.cs
+++ b/Lib9c.Policy/AccessControlService/IAccessControlService.cs
@@ -4,8 +4,6 @@ namespace Nekoyume.Blockchain
 {
     public interface IAccessControlService
     {
-        public bool IsListed(Address address);
-
         public int? GetTxQuota(Address address);
     }
 }
--- a/Lib9c.Policy/NCStagePolicy.cs
+++ b/Lib9c.Policy/NCStagePolicy.cs
@@ -59,7 +59,18 @@ namespace Nekoyume.Blockchain
                     }
 
                     s.Add(tx);
-                    if (s.Count > _quotaPerSigner)
+                    int txQuotaPerSigner = _quotaPerSigner;
+                    if (_accessControlService != null)
+                    {
+                        // update txQuotaPerSigner if ACS returns a value for the signer.
+                        int? acsTxQuota = _accessControlService.GetTxQuota(tx.Signer);
+                        if (acsTxQuota != null)
+                        {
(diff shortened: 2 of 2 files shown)


area363:
@Atralupus This PR replaces https://github.com/planetarium/lib9c/pull/2191 PTAL🙏

Lib9c.Policy/NCStagePolicy.cs
```diff
@@ -77,7 +88,8 @@ public IEnumerable<Transaction> Iterate(BlockChain blockChain, bool filtered = t
 
         public bool Stage(BlockChain blockChain, Transaction transaction)
         {
-            if (_accessControlService != null && _accessControlService.IsAccessDenied(transaction.Signer))
+            var acsTxQuota = _accessControlService?.GetTxQuota(transaction.Signer);
+            if (_accessControlService != null && acsTxQuota == 0)
```
moreal, line 92: The `IAccessControlService` seems to return `int?` so is it enough to use the last condition? `null == 0` is false. (ref: https://stackoverflow.com/a/15777768)

```suggestion
            if (acsTxQuota == 0)
```

Lib9c.Policy/NCStagePolicy.cs
```diff
@@ -59,7 +59,18 @@ public IEnumerable<Transaction> Iterate(BlockChain blockChain, bool filtered = t
...
-                    if (s.Count > _quotaPerSigner)
+                    int txQuotaPerSigner = _quotaPerSigner;
+                    if (_accessControlService != null)
+                    {
+                        // update txQuotaPerSigner if ACS returns a value for the signer.
+                        int? acsTxQuota = _accessControlService.GetTxQuota(tx.Signer);
+                        if (acsTxQuota.HasValue)
+                        {
+                            txQuotaPerSigner = (int)acsTxQuota;
+                        }
```
moreal, line 70: Not critical but in C# way:

```suggestion
                        if (_accessControlService.GetTxQuota(tx.Signer) is { } acsTxQuota)
                        {
                            txQuotaPerSigner = acsTxQuota;
                        }
```

Lib9c.Policy/AccessControlService/IAccessControlService.cs
```diff
@@ -4,6 +4,8 @@ namespace Nekoyume.Blockchain
 {
     public interface IAccessControlService
     {
-        public bool IsAccessDenied(Address address);
+        public bool IsListed(Address address);
```
Atralupus, line 7: Now, It seems to be unnecessary.

area363: Would we ever need to check if an address is in the db? If not, I can remove this too.

Atralupus: `GetTxQuota` can handle that functionality.

Lib9c.Policy/NCStagePolicy.cs
```diff
@@ -59,7 +59,17 @@ public IEnumerable<Transaction> Iterate(BlockChain blockChain, bool filtered = t
...
-                    if (s.Count > _quotaPerSigner)
+                    int txQuotaPerSigner = _quotaPerSigner;
+                    if (_accessControlService != null)
+                    {
+                        // update txQuotaPerSigner if ACS returns a value for the signer.
+                        if (_accessControlService.GetTxQuota(tx.Signer) is { } acsTxQuota)
+                        {
+                            txQuotaPerSigner = acsTxQuota;
+                        }
+                    }
```
moreal, line 70: It uses the `IAccessControlService.GetTxQuota` method only. I think it is simple to use *null propagation*.

```suggestion
                    // update txQuotaPerSigner if ACS returns a value for the signer.
                    if (_accessControlService?.GetTxQuota(tx.Signer) is { } acsTxQuota)
                    {
                        txQuotaPerSigner = acsTxQuota;
                    }
```

moreal (approved):

moreal (approved):
👍🏻

area363:
@Atralupus PTAL one last time before merging 🙏

Atralupus (approved):

PR #2197, opened 2023-11-02
PDX-19: Add UnloadFromGarages Action

XxshiftxX:
Add `UnloadFromGarages` action with data structure and serialize/deserialize code.
Implementation of execution will be added following PR.

### DIFF
--- a/Lib9c.Abstractions/IUnloadFromGaragesV1.cs
+++ b/Lib9c.Abstractions/IUnloadFromGaragesV1.cs
@@ -5,16 +5,15 @@ using Libplanet.Common;
 using Libplanet.Crypto;
 using Libplanet.Types.Assets;
 
-namespace Lib9c.Abstractions;
-
-public interface IUnloadFromGaragesV1
+namespace Lib9c.Abstractions
 {
-    public IReadOnlyList<(
-        Address recipientAvatarAddress,
-        IOrderedEnumerable<(Address balanceAddress, FungibleAssetValue value)>? fungibleAssetValues,
-        IOrderedEnumerable<(HashDigest<SHA256> fungibleId, int count)>? fungibleIdAndCounts,
-        string? memo)> UnloadData
+    public interface IUnloadFromGaragesV1
     {
-        get;
+        public IReadOnlyList<(
+            Address recipientAvatarAddress,
+            IOrderedEnumerable<(Address balanceAddress, FungibleAssetValue value)>?
+            fungibleAssetValues,
+            IOrderedEnumerable<(HashDigest<SHA256> fungibleId, int count)>? fungibleIdAndCounts,
+            string? memo)> UnloadData { get; }
     }
 }
--- a/Lib9c.Abstractions/IUnloadFromGaragesV1.cs
+++ b/Lib9c.Abstractions/IUnloadFromGaragesV1.cs
@@ -1,0 +1,20 @@
+using System.Collections.Generic;
+using System.Security.Cryptography;
+using Libplanet.Common;
+using Libplanet.Crypto;
+using Libplanet.Types.Assets;
+
+namespace Lib9c.Abstractions;
+
+public interface IUnloadFromGaragesV1
+{
+    public IReadOnlyList<(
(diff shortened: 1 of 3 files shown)


Lib9c.Abstractions/IUnloadFromGaragesV1.cs
```diff
@@ -0,0 +1,20 @@
+using System.Collections.Generic;
+using System.Linq;
+using System.Security.Cryptography;
+using Libplanet.Common;
+using Libplanet.Crypto;
+using Libplanet.Types.Assets;
+
+namespace Lib9c.Abstractions;
```
moreal, line 8: It seems failed to build. Because the `Lib9c.Abstractions` doesn't use C# 10.0 yet. It is because Unity doesn't support C# 10 😢 

https://docs.unity3d.com/2023.2/Documentation/Manual/CSharpCompiler.html

moreal (changes requested):

Lib9c/Action/Garages/UnloadFromGarages.cs
```diff
@@ -0,0 +1,143 @@
...
+                fungibleIdAndCounts, string? memo)>();
+        }
+
+        public UnloadFromGarages(
+            IReadOnlyList<(
+                Address recipientAvatarAddress,
+                IEnumerable<(Address balanceAddress, FungibleAssetValue value)>?
+                fungibleAssetValues,
+                IEnumerable<(HashDigest<SHA256> fungibleId, int count)>? fungibleIdAndCounts,
+                string? memo)> unloadData)
```
moreal, line 45: It is too sad to duplicate codes for type because of csharplang's limitation. 😢

Lib9c/Action/Garages/UnloadFromGarages.cs
```diff
@@ -0,0 +1,143 @@
...
+                                $"The type of {nameof(rawValue)} must be bencodex list.");
+                var recipientAvatarAddress = value[0].ToAddress();
+                var fungibleAssetValues =
+                    GarageUtils.MergeAndSort(
+                        (value[1] as List)?.Select(raw =>
+                        {
+                            var fungibleAssetValue = (List)raw;
+                            return (fungibleAssetValue[0].ToAddress(),
+                                fungibleAssetValue[1].ToFungibleAssetValue());
+                        }));
```
moreal, line 95: It will be treated as `null` instead of throwing an error, even if invalid items like `Dictionary` became.

moreal (approved):

Lib9c/Action/Garages/UnloadFromGarages.cs
XxshiftxX:
```
type Foo = (...)
```
😢

moreal (commented):
Also, we should consider `unload_from_garages` as good naming 🤔 Should it be `bulk_unload_from_my_garages`? 🤔 @boscohyun Please leave your opinion 🙏🏻

boscohyun:
> Also, we should consider `unload_from_garages` as good naming 🤔 Should it be `bulk_unload_from_my_garages`? 🤔 @boscohyun Please leave your opinion 🙏🏻

Thanks for mentioning me.
1. It's OK to remove `my` from action identifier. Because the action use signer address basically.
2. Choosing to add `bulk` to the action identifier is easy and reasonable.

XxshiftxX:
> > Also, we should consider `unload_from_garages` as good naming 🤔 Should it be `bulk_unload_from_my_garages`? 🤔 @boscohyun Please leave your opinion 🙏🏻
> 
> Thanks for mentioning me.
> 
> 1. It's OK to remove `my` from action identifier. Because the action use signer address basically.
> 2. Choosing to add `bulk` to the action identifier is easy and reasonable.

@boscohyun @moreal Thank you for your opinion. I think `build_unload_from_garages` is nice option to represent how it works.
It seems to hard to change names now because it has many following PR/branches. So I'll apply this after bunch of branches merge into a single branch so that there's no remaining branch dependency.

PR #2207, opened 2023-11-07
Rename UnloadFromGarages to BulkUnloadFromGarages

XxshiftxX:
SSID


Lib9c/Action/Garages/BulkUnloadFromGarages.cs
```diff
@@ -20,7 +20,7 @@
 namespace Nekoyume.Action.Garages
 {
     [ActionType("unload_from_garages")]
```
moreal, line 22: Shouldn't the action type be updated too?

moreal (changes requested):

moreal (approved):

PR #2208, opened 2023-11-07
Limit ClaimItems.ClaimData's size to 100

moreal:
See PDX-37 card and [internal slack conversation](https://planetariumhq.slack.com/archives/C03PJ5CG30C/p1698982821799689). The limitation prevents the `ClaimItems` to take too long evaluation time since 9c network doesn't have a fee model.


U-lis (approved):
LGTM

PR #2221, opened 2023-11-11
🧹 Remove forced rehearsal extension method

greymistcube:
In preparation for [libplanet](https://github.com/planetarium/libplanet) 4.0.0. Required changes to [9c](https://github.com/planetarium/NineChronicles) has already been merged [here](https://github.com/planetarium/NineChronicles/pull/3396).


riemannulus:
Is `Ninechronicles.Headless` does not using that methods?

greymistcube:
@riemannulus 
Nope. It was only being used in 9c client. 9c-headless isn't really responsible for actually creating `Transaction`s. 😶
I suppose it makes the problem worse since it should be possible to create transactions from 9c-headless, which would be different from the one made by 9c. All the more the reason to get rid of this. 🙄

riemannulus (approved):

PR #2226, opened 2023-11-13
Support wrapped fav burn & mint

ipdae:
- ClaimItems support wrapped fav mint & burn
- Add ClaimItems.memo field 
- Add ClaimItems mail type

---
  - https://app.asana.com/0/0/1205936060181170

### DIFF
--- a/Lib9c/Action/ClaimItems.cs
+++ b/Lib9c/Action/ClaimItems.cs
@@ -105,6 +105,8 @@ namespace Nekoyume.Action
                 }
 
                 var agentAddress = avatarDict[AgentAddressKey].ToAddress();
+                var favs = new List<FungibleAssetValue>();
+                var items = new List<(int id, int count)>();
                 foreach (var fungibleAssetValue in fungibleAssetValues)
                 {
                     var wrappedTicker = fungibleAssetValue.Currency.Ticker;
@@ -116,9 +118,11 @@ namespace Nekoyume.Action
                         var currency = Currencies.GetMinterlessCurrency(ticker);
                         var recipientAddress =
                             currency.DecimalPlaces > 0 ? agentAddress : avatarAddress;
+                        var fav = FungibleAssetValue.FromRawValue(currency, fungibleAssetValue.RawValue);
                         states = states
                             .BurnAsset(context, context.Signer, fungibleAssetValue)
-                            .MintAsset(context, recipientAddress, FungibleAssetValue.FromRawValue(currency, fungibleAssetValue.RawValue));
+                            .MintAsset(context, recipientAddress, fav);
+                        favs.Add(fav);
                     }
                     else
                     {
@@ -152,21 +156,30 @@ namespace Nekoyume.Action
                         // but we'll deal with it temporarily here.
                         // If Pluggable AEV ever becomes a reality,
                         // it's only right that this is fixed in Inventory.
+                        var itemCount = (int)fungibleAssetValue.RawValue;
                         if (item is INonFungibleItem)
                         {
-                            foreach (var _ in Enumerable.Range(0, (int)fungibleAssetValue.RawValue))
+                            foreach (var _ in Enumerable.Range(0, itemCount))
                             {
                                 inventory.AddItem(item, 1);
                             }
                         }
                         else
                         {
-                            inventory.AddItem(item, (int)fungibleAssetValue.RawValue);
+                            inventory.AddItem(item, itemCount);
                         }
+                        items.Add((item.Id, itemCount));
                     }
                 }
 
-                states = states.SetState(inventoryAddress, inventory.Serialize());
+                var mailBox = new MailBox((List)avatarDict[MailBoxKey]);
+                var mail = new ClaimItemsMail(context.BlockIndex, context.GetRandom().GenerateRandomGuid(), context.BlockIndex, favs, items);
+                mailBox.Add(mail);
+                mailBox.CleanUp();
+                avatarDict = avatarDict.SetItem(MailBoxKey, mailBox.Serialize());
+                states = states
+                    .SetState(inventoryAddress, inventory.Serialize())
+                    .SetState(avatarAddress, avatarDict);
             }
 
             return states;
--- a/Lib9c/Action/ClaimItems.cs
+++ b/Lib9c/Action/ClaimItems.cs
@@ -90,21 +90,21 @@ namespace Nekoyume.Action
             var random = context.GetRandom();
             var itemSheet = states.GetSheets(containItemSheet: true).GetItemSheet();
 
-            foreach (var (recipientAddress, fungibleAssetValues) in ClaimData)
+            foreach (var (avatarAddress, fungibleAssetValues) in ClaimData)
             {
-                var inventoryAddress = recipientAddress.Derive(LegacyInventoryKey);
-                Inventory inventory = null;
-                bool inventoryExist = false;
-                try
+                var inventoryAddress = avatarAddress.Derive(LegacyInventoryKey);
+                var inventory = states.GetInventory(inventoryAddress)
+                            ?? throw new FailedLoadStateException(
+                                ActionTypeText,
+                                GetSignerAndOtherAddressesHex(context, inventoryAddress),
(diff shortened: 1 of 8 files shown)


Lib9c/Action/ClaimItems.cs
```diff
@@ -70,62 +71,91 @@ public override IAccount Execute(IActionContext context)
...
-                var inventoryAddress = avatarAddress.Derive(LegacyInventoryKey);
-                var inventory = states.GetInventory(inventoryAddress)
-                                ?? throw new FailedLoadStateException(
-                                    ActionTypeText,
-                                    GetSignerAndOtherAddressesHex(context, inventoryAddress),
-                                    typeof(Inventory),
-                                    inventoryAddress);
+                var inventoryAddress = recipientAddress.Derive(LegacyInventoryKey);
+                Inventory inventory = null;
+                bool inventoryExist = false;
```
longfin, line 78: Could we determine `inventoryExist` with just null check instead of explicit variable?

Lib9c/Action/ClaimItems.cs
```diff
@@ -70,62 +71,91 @@ public override IAccount Execute(IActionContext context)
...
+                try
+                {
+                    inventory = states.GetInventory(inventoryAddress);
+                    inventoryExist = true;
+                }
+                catch (Exception)
```
longfin, line 84: I think that catching `Exception` instead of specific exception for inventory missing would make harder to debug in some circumstances.

Lib9c/Action/ClaimItems.cs
```diff
@@ -70,62 +71,91 @@ public override IAccount Execute(IActionContext context)
...
+                {
+                    // inventory not required if claim fav
+                }
 
                 foreach (var fungibleAssetValue in fungibleAssetValues)
                 {
-                    if (fungibleAssetValue.Currency.DecimalPlaces != 0)
+                    var wrappedTicker = fungibleAssetValue.Currency.Ticker;
+                    var separator = wrappedTicker.Contains("__") ? "__" : "_";
```
longfin, line 92: How about using simple prefix like `w` instead of separator?

Lib9c/Action/ClaimItems.cs
```diff
@@ -74,58 +94,92 @@ public override IAccount Execute(IActionContext context)
...
+                var agentAddress = avatarDict[AgentAddressKey].ToAddress();
+                var favs = new List<FungibleAssetValue>();
+                var items = new List<(int id, int count)>();
                 foreach (var fungibleAssetValue in fungibleAssetValues)
                 {
-                    if (fungibleAssetValue.Currency.DecimalPlaces != 0)
+                    var wrappedTicker = fungibleAssetValue.Currency.Ticker;
+                    var separator = wrappedTicker.Contains("FAV") ? "__" : "_";
+                    var parsedTicker = wrappedTicker.Split(separator);
+                    if (parsedTicker.Length == 2)
```
moreal, line 115:
```suggestion
                    if (parsedTicker.Length == 2 && parsedTicker[0] == "FAV")
```

Lib9c/Action/ClaimItems.cs
```diff
@@ -74,58 +94,92 @@ public override IAccount Execute(IActionContext context)
...
+                }
 
+                var agentAddress = avatarDict[AgentAddressKey].ToAddress();
+                var favs = new List<FungibleAssetValue>();
+                var items = new List<(int id, int count)>();
                 foreach (var fungibleAssetValue in fungibleAssetValues)
                 {
-                    if (fungibleAssetValue.Currency.DecimalPlaces != 0)
+                    var wrappedTicker = fungibleAssetValue.Currency.Ticker;
+                    var separator = wrappedTicker.Contains("FAV") ? "__" : "_";
```
moreal, line 113: Or:
```suggestion
                    var separator = wrappedTicker.StartsWith("FAV") ? "__" : "_";
```

Lib9c/Action/ClaimItems.cs
```diff
@@ -74,58 +94,92 @@ public override IAccount Execute(IActionContext context)
...
+                    var parsedTicker = wrappedTicker.Split(separator);
+                    if (parsedTicker.Length == 2)
                     {
-                        throw new ArgumentException(
-                            $"DecimalPlaces of fungibleAssetValue for claimItems are not 0: {fungibleAssetValue.Currency.Ticker}");
+                        var ticker = parsedTicker[1];
+                        var currency = Currencies.GetMinterlessCurrency(ticker);
+                        var recipientAddress =
+                            currency.DecimalPlaces > 0 ? agentAddress : avatarAddress;
```
moreal, line 120: Note: Maybe there is a promise to use `decimalPlaces` as 0. (e.g., `rune_`, `runestone_`, `soulstone_`)

Lib9c/Action/ClaimItems.cs
```diff
@@ -22,24 +25,37 @@ public class ClaimItems : GameAction, IClaimItems
...
                 .Add(ClaimDataKey, ClaimData.Aggregate(List.Empty, (list, tuple) =>
                 {
                     var serializedFungibleAssetValues = tuple.fungibleAssetValues.Select(x => x.Serialize()).Serialize();
 
                     return list.Add(new List(tuple.address.Bencoded, serializedFungibleAssetValues));
                 }));
+            if (!string.IsNullOrEmpty(Memo))
+            {
+                dict = dict.Add(MemoKey, Memo.Serialize());
+            }
```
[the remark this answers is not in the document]
ipdae: If memo is an empty string, I think there is any reason to serialize it and include it in the chain. (I think network & space are wasted with unnecessary information.) Since the chain compares the serialized state, there are no problem.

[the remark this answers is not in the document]
moreal: > If memo is an empty string, I think there is any reason to serialize it and include it in the chain. (I think network & space are wasted with unnecessary information.)

I think so. Then, I think it should restrict when the `plainValue[MemoKey] = ""` case as alternative.

Lib9c/Action/ClaimItems.cs
```diff
@@ -52,6 +68,10 @@ public ClaimItems(IReadOnlyList<(Address, IReadOnlyList<FungibleAssetValue>)> cl
                         new Address(pair[0]),
                         pair[1].ToList(x => x.ToFungibleAssetValue()) as IReadOnlyList<FungibleAssetValue>);
                 }).ToList();
+            if (plainValue.ContainsKey(MemoKey))
+            {
+                Memo = (Text) plainValue[MemoKey];
+            }
```
moreal, line 74:
```
            if (plainValue.ContainsKey(MemoKey))
            {
            	if (plainValue[MemoKey] is Text text && text.Value.Length == 0) throw new ArgumentException(nameof(plainValue));
                Memo = (Text) plainValue[MemoKey];
            }
```

Lib9c/Action/ClaimItems.cs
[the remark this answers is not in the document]
ipdae: That's why I used `string.isNullorEmpty`, is there better suggestion?

Lib9c/Action/ClaimItems.cs
```diff
@@ -74,58 +94,92 @@ public override IAccount Execute(IActionContext context)
...
-                    if (fungibleAssetValue.Currency.DecimalPlaces != 0)
+                    var wrappedTicker = fungibleAssetValue.Currency.Ticker;
+                    var separator = wrappedTicker.Contains("FAV") ? "__" : "_";
+                    var parsedTicker = wrappedTicker.Split(separator);
+                    if (parsedTicker.Length == 2)
                     {
-                        throw new ArgumentException(
-                            $"DecimalPlaces of fungibleAssetValue for claimItems are not 0: {fungibleAssetValue.Currency.Ticker}");
+                        var ticker = parsedTicker[1];
+                        var currency = Currencies.GetMinterlessCurrency(ticker);
```
moreal, line 118: Note: It means the `FAV_...` will allows only `FAV__GARAGE`, `FAV__CRYSTAL`.

moreal (approved):

PR #2227, opened 2023-11-13
Allow empty NCG vault

longfin:
This PR allows empty initial NCG supply, for planets governing under derived tokenomics.

### DIFF
--- a/.Lib9c.Tests/Model/State/GoldCurrencyStateTest.cs
+++ b/.Lib9c.Tests/Model/State/GoldCurrencyStateTest.cs
@@ -22,6 +22,7 @@ namespace Lib9c.Tests.Model.State
             GoldCurrencyState deserialized = new GoldCurrencyState(serialized);
 
             Assert.Equal(currency.Hash, deserialized.Currency.Hash);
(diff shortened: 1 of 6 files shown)


.Lib9c.Tests/Model/State/GoldCurrencyStateTest.cs
```diff
@@ -42,5 +43,20 @@ public void SerializeWithDotnetAPI()
 
             Assert.Equal(currency.Hash, deserialized.Currency.Hash);
         }
+
+        [Fact]
+        public void SerializeWithInitialSupply()
+        {
+            #pragma warning disable CS0618
```
moreal, line 50:
```suggestion
#pragma warning disable CS0618
```

Lib9c/Action/RewardGold.cs
moreal, line 1: It will be good to test in `RewardGoldTest`.

longfin: Added. https://github.com/planetarium/lib9c/pull/2227/commits/bb6d2a0de156f0fed0aad42cd7ece917df6fc9c1

moreal (approved):

PR #2230, opened 2023-11-14
Update AuraSummon with Sheet model

U-lis:
- Read CSV with loop, not hard code.
- Calculate cumulative ratio on-denamd.

---
  - https://app.asana.com/0/0/1205936092654008


Lib9c/TableData/Summon/SummonSheet.cs
```diff
@@ -31,48 +49,20 @@ public override void Set(IReadOnlyList<string> fields)
...
+                    var idx = 2 * i + 2;
+                    if (TryParseInt(fields[idx], out _) && TryParseInt(fields[idx + 1], out _))
+                    {
+                        Recipes.Add((ParseInt(fields[idx]), ParseInt(fields[idx+1])));
+                    }
+                    else
+                    {
+                        break;
+                    }
                 }
```
moreal, line 66: Is it okay not to check `fields`' length? If not, `fields[idx]` may throw an error.

moreal: It may use https://github.com/planetarium/lib9c/blob/development/Lib9c/TableCSV/Summon/SummonSheet.csv this CSV file, then it may not work as optional.

ipdae (approved):

moreal (approved):

PR #2237, opened 2023-11-15
Fit `MintAssets` to mail system

longfin:
This PR has two changes as below

1. Adds `MintAssets.Memo` for output mails
2. Makes `MintAssets.Execute()` to care about `AvatarState.MailBox` as like `UnloadFromMyGarages` action.
   * it reuses `UnloadFromMyGaragesRecipientMail` 😓 .
   * I'm not a big fan of current mail system, especially spoiling output states. but on the other hands, I don't know about current client's behavior which relying states. so I decided mimics `UnloadFromMyGarages` as possible.

### DIFF
--- a/Lib9c/Action/MintAssets.cs
+++ b/Lib9c/Action/MintAssets.cs
@@ -25,9 +25,10 @@ namespace Nekoyume.Action
         {
         }
 
-        public MintAssets(IEnumerable<MintSpec> specs)
+        public MintAssets(IEnumerable<MintSpec> specs, string? memo)
         {
             MintSpecs = specs.ToList();
+            Memo = memo;
         }
 
         public override IAccount Execute(IActionContext context)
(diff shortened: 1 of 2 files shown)


Lib9c/Action/MintAssets.cs
```diff
@@ -79,32 +119,54 @@ public override IAccount Execute(IActionContext context)
...
-                    )
+                    values.AddRange(mintSpecsNotNull.Select(s => s.Serialize()));
                 }
-            );
+
+                return new Dictionary(
+                    new[]
+                    {
+                        new KeyValuePair<IKey, IValue>((Text)"type_id", (Text)TypeIdentifier),
+                        new KeyValuePair<IKey, IValue>((Text)"values",new List(values))
```
moreal, line 156:
```suggestion
                        new KeyValuePair<IKey, IValue>((Text)"values", new List(values))
```

Lib9c/Action/MintAssets.cs
```diff
@@ -69,8 +84,33 @@ public override IAccount Execute(IActionContext context)
...
+                    state.GetState(recipient) is Dictionary dict &&
+                    dict.TryGetValue((Text)SerializeKeys.MailBoxKey, out IValue rawMailBox)
+                )
+                {
+                    var mailBox = new MailBox((List)rawMailBox);
+                    (List<FungibleAssetValue> favs, List<FungibleItemValue> fivs) = mailRecords[recipient];
+                    mailBox.Add(
+                        new UnloadFromMyGaragesRecipientMail(
+                            context.BlockIndex,
+                            context.GetRandom().GenerateRandomGuid(),
```
moreal, line 103: I think it should use only one random instance per one action context. Or maybe the mail id will be all the same.

https://github.com/planetarium/libplanet/blob/e555adddeac03765f7a4efd6015fb53cbb5231b7/Libplanet.Action/ActionContext.cs#L70-L71

moreal (commented):
I left some comments but it will be good to receive reviews from 9c-unity client developers about mails.

Lib9c/Action/MintAssets.cs
longfin: Amended. https://github.com/planetarium/lib9c/pull/2237/commits/8a5d8b431675752181724a0420ea66caa7e0c03c

Lib9c/Action/MintAssets.cs
longfin: Amended. https://github.com/planetarium/lib9c/commit/8a5d8b431675752181724a0420ea66caa7e0c03c

moreal (approved):
I approved at the action level. Please receive approval from planetarium/NineChronicles developers

sonohoshi (approved):

PR #2258, opened 2023-11-28
Introduce RuneSummon

ipdae:
- Introduce RuneSummon action


Lib9c/Action/RuneSummon.cs
```diff
@@ -0,0 +1,223 @@
...
+        public Address AvatarAddress;
+
+        public const string GroupIdKey = "gid";
+        public int GroupId;
+
+        public const string SummonCountKey = "sc";
+        public int SummonCount;
+
+        private const int SummonLimit = 10;
+        public const int RuneQuantity = 100;
```
U-lis, line 36: IMHO, it seems quite inconvenient to change quantity of rune summoned.
Can we separate it to Sheet data or somewhere easier to manage?

PR #2271, opened 2023-12-05
🧹  chore: cleanup lib9c service

riemannulus:
# Context
Since https://github.com/planetarium/lib9c/pull/2269, we use `Lib9c.Plugin` instead of `Lib9c.StateService` and `Libplanet.Extensions.RemoteActionEvaluator`. So we don't need `Lib9c.StateService` and `Libplanet.Extensions.RemoteActionEvaluator`.

# Rationale
Remove `Lib9c.StateService` and `Libplanet.Extensions.RemoteActionEvaluator` from the project. 👋


OnedgeLee (approved):

moreal (approved):
👋🏻

Issue #2277, opened 2023-12-06
🧹  Cleanup useless actions


riemannulus:
# Context
Since https://github.com/planetarium/lib9c/pull/2265, we don't need old actions anymore. We need to remove useless actions.

# Checklist
- [x] BattleArena
- [x] Buy
- [x] Sell
- [x] ClaimStakeReward
- [x] CombinationConsumeable
- [x] CombinationEquipment
- [x] CreateAvatar
- [x] DailyReward
- [x] HackAndSlash
- [x] HackAndSlashSweep
- [x] ItemEnhancement
- [x] MimisbrunnrBattle
- [x] Raid
- [x] RapidCombination
- [x] RankingBattle
- [x] SellCancellation
- [x] TransferAsset
- [x] UpdateSell
- [x] at la

riemannulus:
Clean complete. 🧹

PR #2280, opened 2023-12-06
🧹  cleanup: remove useless `Buy` actions

riemannulus:
relative #2277


moreal (approved):
I worried about removing some models like `Buy...Result` but I checked this PR doesn't remove the models (`Buy7.cs` file).

https://github.com/planetarium/lib9c/blame/e62f34bbc0572b7a80a1717bce88d42bf33c2f5a/Lib9c/Action/AttachmentActionResult.cs#L19-L20

PR #2281, opened 2023-12-06
🧹  ci: cleanup deprecated build

riemannulus:
Since #2271, we do not have `StateService` anymore. Remove build and push image from CI.


moreal (approved):
Maybe all jobs in this workflow can be deprecated. (this `push_docker_image.yaml` file can be deleted.)

PR #2318, opened 2024-01-02
Validate ticket count

ipdae:
Validate negative ticket count

---
  - https://app.asana.com/0/0/1206261370536137

### DIFF
--- a/.Lib9c.Tests/Action/BattleArenaTest.cs
+++ b/.Lib9c.Tests/Action/BattleArenaTest.cs
@@ -1037,6 +1037,37 @@
             }));
         }
 
+        [Theory]
+        [InlineData(8, null)]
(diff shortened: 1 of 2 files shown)


sonohoshi (approved):

jonny-jeahyunchoi (approved):

.Lib9c.Tests/Action/BattleArenaTest.cs
```diff
@@ -1037,6 +1037,37 @@ public void Execute_ValidateDuplicateTicketPurchaseException()
             }));
         }
 
+        [Theory]
+        [InlineData(8, null)]
+        [InlineData(100, null)]
```
U-lis, line 1042: Don't we block ticket usage greater than 8?

PR #2326, opened 2024-01-09
Adjust policy fields for multiplanetary

longfin:
This PR renames the fields on various policies, to represent more specific planets that will be applied.


Lib9c.Policy/Policy/MaxTransactionsPerSignerPerBlockPolicy.cs
```diff
@@ -17,7 +17,7 @@ private MaxTransactionsPerSignerPerBlockPolicy(int defaultValue)
         public static IVariableSubPolicy<int> Default =>
             new MaxTransactionsPerSignerPerBlockPolicy(int.MaxValue);
 
-        public static IVariableSubPolicy<int> Mainnet =>
+        public static IVariableSubPolicy<int> Odin =>
```
area363, line 20: Can we consider either raising or eliminating the SpannedSubPolicy value across all networks? The current maximum value of 4 is restricting the swift processing of our internal service's transactions.

longfin: I think so. (I'm not a fan of discriminating internal v. public services, but it's why that we need to replace this policy to MEAD based tx control. 😅 )

But even if so, it might be large work for this PR.

Lib9c.Policy/Policy/MaxTransactionsPerBlockPolicy.cs
```diff
@@ -17,7 +17,13 @@ private MaxTransactionsPerBlockPolicy(int defaultValue)
         public static IVariableSubPolicy<int> Default =>
             new MaxTransactionsPerBlockPolicy(int.MaxValue);
 
-        public static IVariableSubPolicy<int> Mainnet =>
+        public static IVariableSubPolicy<int> Odin =>
+            Default
+                .Add(new SpannedSubPolicy<int>(
+                    startIndex: 0,
+                    value: BlockPolicySource.MaxTransactionsPerBlock));
```
moreal, line 24: Just note: the `BlockPolicySource.MaxTransactionsPerBlock` value should be in this `MaxTransactionsPerBlockPolicy.cs` file. 🤔

moreal (approved):

greymistcube (approved):

area363 (approved):

PR #2332, opened 2024-01-16
Support decimal rates of 3 or more digits

moreal:
Before this pull request, it doesn't support `0.025` as decimal rates in `StakeRegularRewardSheet`. There was an error while parsing it as NCG FAV. But it is not necessary to parse the value as NCG FAV. I think it is just to use `FungibleAssetValue.DivRem` method. I replaced the logic with `new BigInteger(decimal / decimal)` to remove digits.

```
Unhandled exception. System.FormatException: The currency NCG (50778d9d92d8d76b325cc572dceb7d30d9887472) does not allow more than 2 decimal places
   at Libplanet.Types.Assets.FungibleAssetValue.Parse(Currency currency, String value) in /path/tolib9c/.Libplanet/Libplanet.Types/Assets/FungibleAssetValue.cs:line 480
   at Lib9c.Tools.Program.Main(String[] args) in /path/to/lib9c/.Lib9c.Tools/Program.cs:line 24
```

### DIFF
--- a/Lib9c/Model/State/StakeRewardCalculator.cs
+++ b/Lib9c/Model/State/StakeRewardCalculator.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Globalization;
+using System.Numerics;
 using Bencodex.Types;
 using Lib9c;
 using Libplanet.Action;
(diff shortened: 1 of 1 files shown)


boscohyun (approved):

moreal:
@boscohyun I amended like below becuase `FungibleAssetValue.ToString()` returns a string value including ticker string. (e.g. 50 NCG)

from:
```
            var stakedNcgDecimal = TableExtensions.ParseDecimal(stakedNcg.ToString());
```

to:
```
            var stakedNcgDecimal = TableExtensions.ParseDecimal(stakedNcg.GetQuantityString());
```

boscohyun:
@moreal Good job!

boscohyun (approved):

moreal:
After talking internally, we've decided to push the bug fix to the next lib9c release, so we're changing the target to the development branch.

PR #2360, opened 2024-01-30
Introduce CollectionState, ActivateCollection

### LINKED ISSUE #2341: 도감 등록 action 제작

ipdae:
- resolve #2341, #2336

### DIFF
--- a/Lib9c/Action/ActivateCollection.cs
+++ b/Lib9c/Action/ActivateCollection.cs
@@ -10,6 +10,7 @@ using Nekoyume.Extensions;
 using Nekoyume.Model.Collection;
 using Nekoyume.Model.Item;
 using Nekoyume.Model.State;
+using Nekoyume.Module;
 using Nekoyume.TableData;
 using static Lib9c.SerializeKeys;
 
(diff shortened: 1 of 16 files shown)


Lib9c/Action/ActivateCollection.cs
```diff
@@ -0,0 +1,126 @@
...
+
+                    materials.Remove(material);
+                }
+
+                if (materials.Any())
+                {
+                    throw new Exception();
+                }
+
+                var collectionAddress = CollectionState.Derive(AvatarAddress);
```
riemannulus, line 79: We do not more need any `Derive` something. Just like this

`state.GetAccount(CollectionState.StateAddress).GetState(AvatarAddress);`

Now `state.GetAccount(CollectionState.StateAddress)` has its storage system. So you can use the same address for each `Account`. (You can easily notice the `Account` is the same as the `RDBMS` table. but key-value table.)

We strongly recommend separating the state by the domain, like `Avatar` and `Inventory`. So you can search each data with `AvatarAddress` or `AgentAddress` without deriving. Just use it like PK.

ipdae: add commit https://github.com/planetarium/lib9c/pull/2360/commits/1d8adecb1e402fc12c5349d80fa5095ae686152f please check 🙏

riemannulus: Much better, but has it passed tests? 

Lib9c/Module/CollectionModule.cs
```
        public static IWorld SetCollectionState(this IWorld world, Address collection, CollectionState state)
        {
            var account = world.GetAccount(Addresses.Collection);
            account = account.SetState(collection, state.SerializeList());
            return world.SetAccount(Addresses.Avatar, account);
        }
```

This should be written state to `Addresses.Collection` I guess...

Lib9c/Model/State/CollectionState.cs
```diff
@@ -0,0 +1,33 @@
...
+        public CollectionState(List serialized)
+        {
+            var rawList = (List) serialized[0];
+            foreach (var value in rawList)
+            {
+                Ids.Add((Integer)value);
+            }
+        }
+
+        public IValue SerializeList()
```
riemannulus, line 27: We have `IBencodable` interface for this kind of type.

riemannulus (approved):

PR #2377, opened 2024-02-07
Update maxtransactionpersignerblockpolicy

area363:
This pull request continues the work started in [PR #2326 on the lib9c repository](https://github.com/planetarium/lib9c/pull/2326).

### Updates
- Integration of `HeimdallInternal` into the `Planet` attribute.
- Modification of the `MaxTransactionsPerSignerPerBlock` setting to a universal default across all networks.
-- This adjustment allows the chain to include more transactions from 9c's internal services (such as `IAP`, `SeasonPass`, `WorldBoss`, etc.) in a single block, enhancing processing speed. The internal service transaction limit is established at `100` within the `ACC service`.
-- Transactions from regular users will continue to be limited by the `txQuotaPerSigner` parameter, which remains at `1` for all networks.

### Todo
- Update `MaxTransactionsPerSignerPerBlockPolicy` to a `MEAD` based tx control.

### DIFF
--- a/Lib9c.Policy/Policy/BlockPolicySource.cs
+++ b/Lib9c.Policy/Policy/BlockPolicySource.cs
@@ -72,6 +72,12 @@ namespace Nekoyume.Blockchain.Policy
                     maxTransactionsPerBlockPolicy: MaxTransactionsPerBlockPolicy.Heimdall,
                     maxTransactionsPerSignerPerBlockPolicy: MaxTransactionsPerSignerPerBlockPolicy.Heimdall
                 ),
+                Planet.HeimdallInternal => GetPolicy(
+                    maxTransactionsBytesPolicy: MaxTransactionsBytesPolicy.Heimdall,
+                    minTransactionsPerBlockPolicy: MinTransactionsPerBlockPolicy.Heimdall,
+                    maxTransactionsPerBlockPolicy: MaxTransactionsPerBlockPolicy.Heimdall,
+                    maxTransactionsPerSignerPerBlockPolicy: MaxTransactionsPerSignerPerBlockPolicy.HeimdallInternal
+                ),
                 _ => throw new ArgumentException(
                     $"Can't retrieve policy for given planet ({planet})",
                     nameof(planet)
--- a/Lib9c.Policy/Policy/MaxTransactionsPerSignerPerBlockPolicy.cs
+++ b/Lib9c.Policy/Policy/MaxTransactionsPerSignerPerBlockPolicy.cs
@@ -18,27 +18,16 @@ namespace Nekoyume.Blockchain.Policy
             new MaxTransactionsPerSignerPerBlockPolicy(int.MaxValue);
 
         public static IVariableSubPolicy<int> Odin =>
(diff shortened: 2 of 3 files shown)


area363:
Hi @moreal Could you take a look at this PR? It's designed to match odin and heimdall's `MaxTransactionsPerSignerPerBlockPolicy` (odin currently restricts all signer's tx to 4 per block whereas heimdall does not so it's harder for odin's internal services to quickly process txs when there's a lot of user inflow). 🙏

Lib9c.Policy/Policy/MaxTransactionsPerSignerPerBlockPolicy.cs
```diff
@@ -18,27 +18,16 @@ private MaxTransactionsPerSignerPerBlockPolicy(int defaultValue)
...
-            Default
-                // Note: Introduced to prevent transactions spamming that may result in
-                // the chain grinding to a halt without meaningful state transitions happening.
-                // See https://github.com/planetarium/libplanet/issues/1449.
-                // Issued for v100086.
-                // FIXME: Starting index and value must be finalized accordingly before deployment.
-                .Add(new SpannedSubPolicy<int>(
-                    startIndex: 2_800_001,
-                    value: 4));
+            Default;
```
moreal, line 21: Is it okay to remove this restriction? I'm not sure how `MaxTransactionsPerSignerPerBlockPolicy` affects but many users doesn't have enough 'MEAD's now and their transactions will fail.

Lib9c.Policy/Policy/BlockPolicySource.cs
```diff
@@ -72,6 +72,12 @@ public IBlockPolicy GetPolicy(Planet planet)
                     maxTransactionsPerBlockPolicy: MaxTransactionsPerBlockPolicy.Heimdall,
                     maxTransactionsPerSignerPerBlockPolicy: MaxTransactionsPerSignerPerBlockPolicy.Heimdall
                 ),
+                Planet.HeimdallInternal => GetPolicy(
+                    maxTransactionsBytesPolicy: MaxTransactionsBytesPolicy.Heimdall,
+                    minTransactionsPerBlockPolicy: MinTransactionsPerBlockPolicy.Heimdall,
+                    maxTransactionsPerBlockPolicy: MaxTransactionsPerBlockPolicy.Heimdall,
+                    maxTransactionsPerSignerPerBlockPolicy: MaxTransactionsPerSignerPerBlockPolicy.HeimdallInternal
```
moreal, line 79: Why is it different with the Heimdall network? If the `maxTransactionsPerSignerPerBlockPolicy` is different with mainnet, a case when a service having not enough MEAD produces many txs without enough interval and their txs are failed, cannot be caught in internal testing. (If internal's `maxTransactionsPerSignerPerBlockPolicy` is bigger than mainnet's one.)

moreal (approved):
I've heard that there is a follow-up to include transactions to match MEAD holdings. We shared via internal messenger that it is okay because this is currently restricted at the staging level.

PR #2440, opened 2024-03-06
Implement NCIP-18

### LINKED ISSUE #2438: Implement NCIP-18
- implement https://github.com/planetarium/NCIPs/pull/13

ipdae:
- resolve #2438

### DIFF
--- a/.Lib9c.Tests/Action/RetrieveAvatarAssetsTest.cs
+++ b/.Lib9c.Tests/Action/RetrieveAvatarAssetsTest.cs
@@ -30,6 +30,42 @@ namespace Lib9c.Tests.Action
         private static readonly Dictionary<string, string>
             _csv = TableSheetsImporter.ImportSheets();
 
+        private readonly Address _signer;
+        private readonly IWorld _state;
+
+        public RetrieveAvatarAssetsTest()
+        {
+            var ca = new CreateAvatar
+            {
+                index = 0,
+                hair = 2,
+                lens = 3,
+                ear = 4,
+                tail = 5,
+                name = "JohnDoe",
+            };
+            _signer = new PrivateKey().Address;
+            IWorld state = new World(new MockWorldState());
+            foreach (var (key, value) in _csv)
+            {
+                state = state.SetLegacyState(Addresses.GetSheetAddress(key), (Text)value);
+            }
+
+            state = state
+                .SetLegacyState(
+                    Addresses.GameConfig,
+                    new GameConfigState(_csv[nameof(GameConfigSheet)]).Serialize())
+                .SetLegacyState(Addresses.GoldCurrency, new GoldCurrencyState(_currency).Serialize());
+
+            _state = ca.Execute(new ActionContext
+            {
+                PreviousState = state,
+                BlockIndex = 0,
+                Signer = _signer,
+                RandomSeed = 0,
+            });
+        }
+
         [Fact]
         public void PlainValue()
         {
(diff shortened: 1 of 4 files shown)


jckdotim:
@planetarium/9c-dev it might be sensitive code because it directly handles NCG. I will also take a quick look, but I hope at least 1 person need to review the whole execution code very carefully so that we can avoid potential vulnerability regarding it. thanks.

Lib9c/Action/RetrieveAvatarAssets.cs
```diff
@@ -0,0 +1,55 @@
...
+            Address signer = context.Signer;
+            var state = context.PreviousState;
+            if (state.TryGetAvatarState(signer, AvatarAddress, out _))
+            {
+                var currency = state.GetGoldCurrency();
+                var balance = state.GetBalance(AvatarAddress, currency);
+                if (balance > 0 * currency)
+                {
+                    return state.TransferAsset(context, AvatarAddress, signer, balance);
+                }
```
moreal, line 49: If there is no enough balance (i.e., no NCG), it should throw an exception like `Insufficient...Exception`. If not, it will throw `FailedLoadStateException` and it makes workers on debugging this action confused.

ipdae: If the balance is insufficient, it will be throw InsufficientBalanceException internally in TransferAsset interface, so there is no need to implement it separately. i'll check and add a test case.

moreal (approved):
It seems okay to me. We should care about `TryGetAvatarState` logic's avatar ownership check spec not change.

.Lib9c.Tests/Action/RetrieveAvatarAssetsTest.cs
```diff
@@ -0,0 +1,162 @@
...
+        [Fact]
+        public void Execute_Throw_InsufficientBalanceException()
+        {
+            var agentState = _state.GetAgentState(_signer);
+            Assert.NotNull(agentState);
+            var avatarAddress = agentState.avatarAddresses[0];
+            Assert.Equal(0 * _currency, _state.GetBalance(avatarAddress, _currency));
+
+            var action = new RetrieveAvatarAssets(avatarAddress);
+            Assert.Throws<InsufficientBalanceException>(() => action.Execute(new ActionContext
```
moreal, line 153:
```
Expected: typeof(Libplanet.Action.State.InsufficientBalanceException)
Actual:   typeof(System.ArgumentOutOfRangeException): The value to transfer has to be greater than zero. (Parameter 'value')
```

It seems to throw ArgumentOutOfRangeException.

https://docs.libplanet.io/4.0.6/api/Libplanet.Action.State.IAccount.html#Libplanet_Action_State_IAccount_TransferAsset_Libplanet_Action_IActionContext_Libplanet_Crypto_Address_Libplanet_Crypto_Address_Libplanet_Types_Assets_FungibleAssetValue_System_Boolean_

- `arg <= 0` => ArgumentOutOfRangeException
- `arg > 0` => InsufficientBalanceException

moreal (changes requested):

Lib9c/Action/RetrieveAvatarAssets.cs
```diff
@@ -0,0 +1,52 @@
...
+            var asDict = (Dictionary)((Dictionary)plainValue)["values"];
+            AvatarAddress = asDict["a"].ToAddress();
+        }
+
+        public override IWorld Execute(IActionContext context)
+        {
+            context.UseGas(1);
+            Address signer = context.Signer;
+            var state = context.PreviousState;
+            if (state.TryGetAvatarState(signer, AvatarAddress, out _))
```
moreal, line 42: If this line's purpose is only to check `AvatarAddress` is the address of `context.Signer`'s avatar, `TryGetAvatarState()` method can be too heavy. It should deserialize AvatarState from raw bencodex. It may affect to this action's performance.

moreal (approved):

Lib9c/Action/RetrieveAvatarAssets.cs
ipdae: it would be better to just call GetAgentState and then check the `agentState.AvatarAddresses` 🤔

PR #2474, opened 2024-03-20
Fix stat conversion debuff check

ipdae:
- fix `StatBuff.IsDebuff` check when stat conversion skill case (power is 0. but custom field has value)


U-lis (approved):

U-lis:
Test fails :cry:

PR #2515, opened 2024-04-05
Fix errors on DPoS

limebell:
- Fixed error that occurred transferring 0 asset.
- Fixed error that the unbonded validator still remains in the validator set.


Lib9c/Action/DPoS/Control/AllocateReward.cs
s2quake, line 1: The file name and class name are different, is that ok?

.Lib9c.Plugin/PluginActionEvaluator.cs
```diff
@@ -18,7 +20,8 @@ public PluginActionEvaluator(IPluginKeyValueStore keyValueStore)
         {
             var stateStore = new TrieStateStore(new WrappedKeyValueStore(keyValueStore));
             _actionEvaluator = new ActionEvaluator(
-                _ => new RewardGold(),
```
s2quake, line 21: Is the `RewardGold` action no longer used? I'm asking because I'm not sure. 🤔

limebell: Yes. In PoS, gold distribution will be executed by PoS logic. Missing logic will be moved to block action later.

PR #2531, opened 2024-04-16
Add more complex slashing test code with random NCG and delegators.

s2quake:
This PR includes the following.

* Fixed issue where exception is thrown if slash factor is 0.
* Added slashing test code for random NCGs and random delegators.

The reason why I created this PR is to make sure that slashing is working correctly in a variety of situations (such as random NCGs and delegates).

### DIFF
--- a/.Lib9c.Tests/Action/DPoS/Control/SlashCtrlComplexTest.cs
+++ b/.Lib9c.Tests/Action/DPoS/Control/SlashCtrlComplexTest.cs
@@ -4,28 +4,33 @@ namespace Lib9c.Tests.Action.DPoS.Control
     using System.Linq;
     using Libplanet.Action.State;
     using Libplanet.Crypto;
-    using Libplanet.Types.Assets;
     using Nekoyume.Action.DPoS.Control;
-    using Nekoyume.Action.DPoS.Exception;
     using Nekoyume.Action.DPoS.Misc;
     using Nekoyume.Action.DPoS.Model;
     using Xunit;
 
-    public class SlashCtrlTest : SlashCtrlTestBase
+    public class SlashCtrlComplexTest : SlashCtrlTestBase
     {
+        public static readonly object[][] TestData
+            = Enumerable.Range(0, Random.Shared.Next(5, 10))
+                        .Select(item => new object[]
+                        {
+                            Random.Shared.Next() % 2 == 0, // jailed
+                            Random.Shared.Next(1, 101), // slashFactor
+                        })
+                        .ToArray();
+
         private const int ValidatorCount = 2;
-        private const int DelegatorCount = 2;
+        private static readonly int DelegatorCount = Random.Shared.Next(2, 32);
 
         private readonly PublicKey[] _operatorPublicKeys;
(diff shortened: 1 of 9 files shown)


.Lib9c.Tests/Action/DPoS/Control/SlashCtrlComplexTest.cs
```diff
@@ -0,0 +1,588 @@
...
+        }
+
+        [Theory]
+        [MemberData(nameof(TestData))]
+        public void Slash_Test(bool jailed, int slashFactor)
+        {
+            // Given
+            var validatorNCG = NextNCG();
+            var validatorPower = PowerFromNCG(validatorNCG);
+            var validatorShare = ShareFromPower(validatorPower);
```
OnedgeLee, line 55: IMO, it's better to obtain `validatorPower` and `validatorShare` from states via `GetPower()` or `GetShare()` right after validator promotion, considering future modification on tests.
Actually, there are lots of lines like this, which seems to be fragile for me, considering test extension(for example, on the condition that inflation is added right after promotion) but only for this small unit test context, okay.

.Lib9c.Tests/Action/DPoS/Control/SlashCtrlTestBase.cs
```diff
@@ -0,0 +1,173 @@
...
+        {
+            if (!value.Currency.Equals(Asset.ConsensusToken))
+            {
+                throw new ArgumentException(
+                        message: $"'{value}' is not {nameof(Asset.ConsensusToken)}",
+                        paramName: nameof(value));
+            }
+
+            return FungibleAssetValue.FromRawValue(Asset.Share, value.RawValue);
+        }
```
OnedgeLee, line 134: `ShareFromPower` can make confusion, IMO.
`Power` is always equivalent to `GovernanceToken`, while `Share` isn't.
`Share` is conditionally equivalent to `GovernanceToken`, and existence of `ShareFromPower` method can make people to think those are exchangeable, though it's code for tests.

OnedgeLee (approved):
Left two comments, but I think those are trivial and matter of tastes for test, so approved.

limebell (approved):
LGTM

.Lib9c.Tests/Action/DPoS/Control/SlashCtrlComplexTest.cs
s2quake: Except for `validatorNCG`, `validatorPower` and `validatorShare` are defined as expected values to be used in tests. Therefore, if a test passes with the expected and actual values being equal, I believe there is no fragility.

.Lib9c.Tests/Action/DPoS/Control/SlashCtrlTestBase.cs
s2quake: Hmm, is there any way to fix this? I have no idea right now, or is there a proper method name?

s2quake: First, I think you're right.

PR #2543, opened 2024-04-22
Feature/fee collect

limebell:
libplanet: https://github.com/planetarium/libplanet/pull/3762

### DIFF
--- a/Lib9c/Action/DPoS/Model/ProposerInfo.cs
+++ b/Lib9c/Action/DPoS/Model/ProposerInfo.cs
@@ -1,0 +1,31 @@
+﻿using Bencodex.Types;
+using Libplanet.Crypto;
+using Nekoyume.Action.DPoS.Util;
+
+namespace Nekoyume.Action.DPoS.Model
+{
+    public class ProposerInfo
+    {
+        public ProposerInfo(long blockIndex, Address proposer)
+        {
+            BlockIndex = blockIndex;
+            Proposer = proposer;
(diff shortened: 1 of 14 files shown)


Lib9c/Action/DPoS/Model/ProposerInfo.cs
```diff
@@ -0,0 +1,31 @@
+﻿using Bencodex.Types;
+using Libplanet.Crypto;
+using Nekoyume.Action.DPoS.Util;
+
+namespace Nekoyume.Action.DPoS.Model
+{
+    public class ProposerInfo
```
s2quake, line 7:
```suggestion
    public class ProposerInfo : IBencodable
```
I'm not sure, but shouldn't `ProposerInfo` implement `IBencodable`?

Lib9c/Action/DPoS/Sys/AllocateReward.cs
```diff
@@ -34,12 +36,14 @@ public override IWorld Execute(IActionContext context)
             var states = context.PreviousState;
             var nativeTokens = ImmutableHashSet.Create(
                 Asset.GovernanceToken, Asset.ConsensusToken, Asset.Share);
-            states = Control.AllocateRewardCtrl.Execute(
+            var previousProposerInfo =
```
s2quake, line 39: It looks like this code will throw an exception if the `AllocateReward` action is executed when the genesis block is created.

PR #2631, opened 2024-06-17
♻️ Refactor state ncg

greymistcube:
♻️ Uses a randomly generated "NCG-type" currency each time tests are run.


Lib9c/Action/DPoS/Sys/AllocateReward.cs
```diff
@@ -35,9 +37,8 @@ public override void LoadPlainValue(IValue plainValue)
         public override IWorld Execute(IActionContext context)
         {
             var states = context.PreviousState;
-            var nativeTokens = ImmutableHashSet.Create(
-                Asset.GovernanceToken, Currencies.Mead);
-;
+            var nativeTokens = states.GetNativeTokens();
```
limebell, line 40: May be missing `Currencies.Mead`?

limebell (changes requested):

Lib9c/Action/DPoS/Sys/AllocateReward.cs
greymistcube: In that case, it is rather worrying that all the tests pass other than those in `BlockPolicyTest`. 😑
`GetNativeTokens()` is defined as an extension method in `LegacyModule.cs` as the following:

```csharp
public static ImmutableHashSet<Currency> GetNativeTokens(this IWorldState worldState) =>
    ImmutableHashSet.Create<Currency>(
        GetGoldCurrency(worldState),
        Asset.ConsensusToken,
        Asset.Share);
```

Would adding `Currencies.Mead` solve the issue?

greymistcube: Added `Currencies.Mead` to `GetNativeTokens()`. See 3521e1372ab0dfa5e3ad2ed905dbbd8dbe2fb723.

limebell (approved):

PR #2641, opened 2024-06-20
Apply code reviews and improve performances

U-lis:
- Use Weighted selector to pick reward
- Add missing conditions
    - Cannot explore(challenge) after reached max floor before unlock
    - Transfer remaining bounty once
- Optimization
    - Use avatar state to handle avatar and inventory, not call inventory state once again
    - Validate signer-avatar using AddressesHex, not calling state


jonny-jeahyunchoi (approved):

U-lis:
@ipdae idun setting 을 위해 merge 합니다. 리뷰 부탁드립니다.

PR #2668, opened 2024-07-04
Bugfix/fix mead

limebell:


.Lib9c.Tests/TempTest.cs
```diff
@@ -0,0 +1,72 @@
...
+            Log.Debug("ValidatorAddress: {validatorAddress}", validatorAddress);
+            Assert.True(true);
+        }
+
+        // Test that loads a chain from local file.
+        [Fact]
+        public void LoadChainFromLocalFile()
+        {
+            // Arrange
+            var storePath = "C:\\Users\\lime_\\planetarium\\local-test\\mainnet-snapshot";
```
boscohyun, line 46: Presumably this `TempTest` is not intended to be merged, right?

Lib9c/Action/RewardGold.cs
```diff
@@ -38,7 +38,7 @@ public override void LoadPlainValue(IValue plainValue)
 
         public override IWorld Execute(IActionContext context)
         {
-            context.UseGas(1);
```
boscohyun, line 41: Is `RewardGold` action no longer gassing?

limebell: Despite of the line, it was treated as a `BlockAction`, so it did not consume gas actually at the libplanet level before. With changes in how actions are processed in libplanet, keeping that line could potentially cause issues, which is why it has been removed to maintain consistency with previous behavior.

boscohyun: Thanks for the explanation!

PR #2717, opened 2024-07-31
Introduce MigrateFee action

### LINKED ISSUE #2704: Migration action for fee address

ipdae:
- resolve #2704 
- 마이그레이션을 실행할 대상 주소가 정해지면 현재사용되는 액션들의 fee address도 해당 주소로 변경해야합니다.

### DIFF
--- a/Lib9c/Action/MigrateFee.cs
+++ b/Lib9c/Action/MigrateFee.cs
@@ -1,9 +1,11 @@
 using System.Collections.Generic;
 using System.Linq;
+using System.Numerics;
 using Bencodex.Types;
 using Libplanet.Action;
 using Libplanet.Action.State;
 using Libplanet.Crypto;
+using Libplanet.Types.Assets;
 using Nekoyume.Model.State;
 using Nekoyume.Module;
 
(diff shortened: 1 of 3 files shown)


Lib9c/Action/MigrateFee.cs
```diff
@@ -0,0 +1,57 @@
...
+namespace Nekoyume.Action
+{
+    [ActionType(TypeIdentifier)]
+    public class MigrateFee : ActionBase
+    {
+        public const string TypeIdentifier = "migrate_fee";
+        // TODO Will change once target address is determined
+        public static readonly Address TargetAddress = new Address();
+
+        public List<Address> FeeAddresses;
```
moreal, line 19: Can we not use pre-calculated fee addresses? When I looked at the `Execute` logic, it seems to be able to take any NCG, even if the address is not the fee address. (e.g. StakeState's NCG, Agent's NCG, etc)

If it is impossible, the current logic is also okay.

ipdae:
@moreal refactoring `MigrateFee` action due to the following requirements:

- instead of sending the entire balance for each address, a specified amount needs to be sent.
- the action name should be more generalized (inspired by the lesson learned from using PrepareRewardAssets for actual asset minting).
- for the reasons above, recipient address will be added as a parameter.

moreal:
> @moreal refactoring `MigrateFee` action due to the following requirements:
> 
> * instead of sending the entire balance for each address, a specified amount needs to be sent.
> * the action name should be more generalized (inspired by the lesson learned from using PrepareRewardAssets for actual asset minting).
> * for the reasons above, recipient address will be added as a parameter.

@ipdae At first, thanks to your description 🙏🏻 It looks good but I think the action name should be continued as `MigrateFee` or only its behaviour should be continued. The reason is the same with the comment I left https://github.com/planetarium/lib9c/pull/2717/files#r1697805900. The action to transfer someone's assets in force MUST NOT be introduced.

ipdae:
> @ipdae At first, thanks to your description 🙏🏻 It looks good but I think the action name should be continued as MigrateFee or only its behaviour should be continued. The reason is the same with the comment I left https://github.com/planetarium/lib9c/pull/2717/files#r1697805900. The action to transfer someone's assets in force MUST NOT be introduced.

@moreal I'll keep the action name as it is. Since address validation requires checking the state, it might be better to perform this validation when preparing the PlainValue rather than during Execute. action also checking permissions, it seems there shouldn't be any particular issues with this approach. do you think this is acceptable?

moreal:
> @moreal I'll keep the action name as it is. Since address validation requires checking the state, it might be better to perform this validation when preparing the PlainValue rather than during Execute. action also checking permissions, it seems there shouldn't be any particular issues with this approach. do you think this is acceptable?

From the perspective of working at Planetarium, I think it's an acceptable direction, but I don't think it's very pleasant to have an action in the protocol that allows an admin to take property from an arbitrary address. If Planetarium can take it because they don't like someone's behavior, that alone is a red flag. Of course, as I said before, I work for Planetarium, so I trust them, so I think it's acceptable.

moreal:
@ipdae I don't want to delay this further, I think Nine Chronicles is a decentralized game, but I think it's still in the process of changing to become a decentralized game, and if it doesn't actually do that, then it might be okay to leave that possibility open. I apologize if my unnecessary concern has caused the PR to stall, I'll approve it. 🙏🏻

moreal (approved):

PR #2719, opened 2024-07-31
Implement Guild system

moreal:
This pull request implements several actions:
- `ApplyGuild`
- `CancelGuildApplication`
- `RejectGuildApplication`
- `AcceptGuildApplication`
- `MakeGuild`
- `RemoveGuild`
- `QuitGuild`
- `BanGuildMember`
- `UnbanGuildMember`

> [!NOTE]
> It started using *typed-addresses* to solve the problem of using only the `Address` type. This helps to avoid mistakes when taking arguments in the same order as `(AgentAddress, GuildAddress)`, such as `AgentAddress` for an `Agent`'s address and `GuildAddress` for a `Guild`'s address.

> [!NOTE]
> It reserved `Addresses.EmptyAccountAddress` account to remove banned addresses once by overwriting the banned account with the account (`EmptyAccountAddress`, `Trie(root = null)`).


Lib9c/Action/Guild/ApplyGuild.cs
```diff
@@ -0,0 +1,70 @@
...
+            var world = context.PreviousState;
+            var signer = context.GetAgentAddress();
+
+            if (world.GetJoinedGuild(signer) is not null)
+            {
+                throw new InvalidOperationException("The signer is already joined in a guild.");
+            }
+
+            // NOTE: Check there is such guild.
+            _ = world.GetGuild(GuildAddress);
```
s2quake, line 58: Wouldn't it be better to use TryGetGuild for consistency?

Lib9c/Action/Guild/BanGuildMember.cs
```diff
@@ -0,0 +1,88 @@
...
+            {
+                throw new InvalidOperationException("The signer is not a guild master.");
+            }
+
+            if (guild.GuildMasterAddress == Target)
+            {
+                throw new InvalidOperationException("The guild master cannot be banned.");
+            }
+
+            world = world.Ban(guildAddress, Target);
```
s2quake, line 73: It's best to run the Ban method last.
First do the guild reject, then leave the guild, then do the ban.
It shouldn't matter since it's done in one action anyway, but hmm...

moreal: It was just an afterthought, a feature that was added later. The order doesn't seem to matter much. ✅

Lib9c/Action/Guild/ApplyGuild.cs
moreal: I fixed it in [5cd8026](https://github.com/planetarium/lib9c/pull/2719/commits/5cd8026b64d9c0cf803e848666cb3af92adf3302)

Lib9c/Action/Guild/BanGuildMember.cs
moreal: I amended it in 48b20bd1237d35629b2dc4008c68a6594e0973e4

s2quake (approved):

s2quake:
I believe that the reusability of the module has improved significantly.
You shouldn't have to write any test code for Action 😀(although you should 😭).

moreal:
@ipdae When #2723 is merged after this pull request merged, I'll update the `QuitGuild` action because it requires those features.

PR #2720, opened 2024-07-31
Introduce IssueToken action

### LINKED ISSUE #2709: IssueToken
개러지 토큰을 비용으로 지불해서 Item, FungibleAssetValue 토큰을 발급합니다.

ipdae:
- resolve #2709 
- burn FungibleAssetValue, TradableMaterial and mint token for using ClaimItems action


jaeho0103 (approved):

Lib9c/TableData/Garages/LoadIntoMyGaragesCostSheet.cs
```diff
@@ -29,11 +30,12 @@ public class Row : SheetRow<int>
             public override void Set(IReadOnlyList<string> fields)
             {
                 Id = ParseInt(fields[0]);
-                CurrencyTicker = fields[1];
-                FungibleId = string.IsNullOrEmpty(fields[2])
+                ItemId = TryParseInt(fields[1], out var result) ? result : 0;
+                CurrencyTicker = fields[2];
+                FungibleId = string.IsNullOrEmpty(fields[3])
```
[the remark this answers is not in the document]
ipdae: Even if this sheet only using planetarium, changing the order of the column so concerning. I'll change the order of the ItemId fields.

.Lib9c.Tests/Model/Item/InventoryTest.cs
```diff
@@ -995,6 +995,30 @@ public void TryGetNonFungibleItem()
             Assert.False(inventory.TryGetNonFungibleItem(equipment.NonFungibleId, out Equipment _, 1L));
         }
 
+        [Fact]
+        public void RemoteTradableMaterial()
```
U-lis, line 999:
```suggestion
        public void RemoveTradableMaterial()
```

U-lis (approved):
Maybe there is typo, else are looks good to me.

Lib9c/Action/IssueToken.cs
```diff
@@ -0,0 +1,98 @@
...
+            var garageCost = sheet.GetGarageCost(FungibleAssetValues, Items);
+            state = state.TransferAsset(
+                context,
+                context.Signer,
+                Addresses.GarageWallet,
+                garageCost
+            );
+            foreach (var fungibleAssetValue in FungibleAssetValues)
+            {
+                var currency = fungibleAssetValue.Currency;
```
moreal, line 67: It should check the currency is `NCG`, or it has `minters` field. Because, to use issued *Item Token*s, it should use `ClaimItems` action. And the `ClaimItems` action burns the wrapped currency FAVs and mints unwrapped currency FAVs. But it cannot mint the FAVs when its currency has `minters` field. For instance, `NCG` has `minters` field so `FAV__NCG` wrapped currency FAVs cannot be unwrapped.

Lib9c/Action/IssueToken.cs
```diff
@@ -0,0 +1,98 @@
...
+                var wrappedCurrency = Currencies.GetWrappedCurrency(currency);
+                var recipient = GarageUtils.PickRecipient(context.Signer, AvatarAddress, currency);
```
moreal, line 69: It may not be important. But I think it should use other word instead of `recipient` because nobody sends to the recipient and the recipient doesn't receive anything.

moreal (changes requested):
I submitted ***Request changes*** because of https://github.com/planetarium/lib9c/pull/2720#discussion_r1701841543 comment. I think it should be fixed before merging.

Lib9c/Action/IssueToken.cs
ipdae: LoadIntoMyGaragesCostSheet already check `Currency` or `Item` can be minting.

moreal (approved):

Lib9c/Action/IssueToken.cs
moreal: It seems resolved by [9071233](https://github.com/planetarium/lib9c/pull/2720/commits/90712336e508662bb52cd340876d5e89392a29ef)

moreal:
The new added test seems failed. Please check it 🙏🏻

moreal (approved):

PR #2722, opened 2024-07-31
Join accounts with pledge to guild automatically

moreal:
> [!NOTE]
> #2719 must be merged before this pull request.

This pull request introduces `AutoJoinGuild` policy action to be used as `PolicyActionsRegistry.beginTxActions` and `MigratePledgeToGuild` action. The actions work like the below cases:

```
// type (PlanetariumGuild, UserGuild, UserPledge)
(true, null, Pledge) => JoinGuild,
(true, Guild, Pledge) => DoNothing,  // It has already guild.
(true, null, null) => DoNothing,  // It didn't contract pledge yet.
(false, _, _) => DoNothing,  // Planetarium didn't make guild yet.
```


Lib9c/PolicyAction/Tx/Begin/AutoJoinGuild.cs
```diff
@@ -0,0 +1,43 @@
...
+            {
+                throw new InvalidOperationException(
+                    "This action must be called when it is a policy action.");
+            }
+
+            var world = context.PreviousState;
+            var signer = context.GetAgentAddress();
+
+            var (resultWorld, _) = GuildMigrationCtrl.MigratePlanetariumPledgeToGuild(world, signer);
+            return resultWorld;
```
s2quake, line 40: In general (TryParse, TryGetValue, etc.), if the return type is bool, there should be a guarantee that this method will not throw any exceptions.
I think it would be a bit confusing for someone using this method to have a mix of successes, failures, and exceptions.
What I would suggest is to define and throw an exception class that can be used in the case of returning failure (true).
Then, in AutoJoinGuild, make sure that only that exception is handled by the catch.

Lib9c/Action/Guild/Migration/Controls/GuildMigrationCtrl.cs
```diff
@@ -0,0 +1,75 @@
...
+
+            var pledgeAddress = target.GetPledgeAddress();
+
+            // Patron contract structure:
+            // [0] = PatronAddress
+            // [1] = IsApproved
+            // [2] = Mead amount to refill.
+            if (!world.TryGetLegacyState(pledgeAddress, out List list) || list.Count < 3 || list[0].ToAddress() != MeadConfig.PatronAddress)
+            {
+                return (world, true);
```
s2quake, line 69: In general (TryParse, TryGetValue, etc.), if the return type is bool, there should be a guarantee that this method will not throw any exceptions.
I think it would be a bit confusing for someone using this method to have a mix of successes, failures, and exceptions.
What I would suggest is to define and throw an exception class that can be used in the case of returning failure (true).
Then, in AutoJoinGuild, make sure that only that exception is handled by the catch.

Lib9c/TypedAddress/PledgeAddress.cs
```diff
@@ -0,0 +1,59 @@
+using Bencodex;
+using Bencodex.Types;
+using Libplanet.Crypto;
+
+namespace Nekoyume.TypedAddress
+{
+    public readonly struct PledgeAddress : IBencodable
```
s2quake, line 7: If there is a class or struct that is repeated or similar to `AgentAddress`, it would be nice to use the C# Generic feature.

Lib9c/Action/Guild/Migration/Controls/GuildMigrationCtrl.cs
moreal: So are you saying that we should create an exception for every reason why it fails and throw it? If so, I think I'd need about 6 exception classes. I'm a little torn because I don't think there's anything extra to handle when I get an exception, or maybe it's not very C#-like to return something like (result, err) like Go, so I could fix it with something like `TryMigrate...`?

moreal: I tried to apply your opinion at [b4d361e](https://github.com/planetarium/lib9c/pull/2722/commits/b4d361e6d590aba3e87ae990fbd21b2ab00649f0). Please review again 🙏🏻

s2quake: LGTM 😀

Lib9c/TypedAddress/PledgeAddress.cs
moreal: What is the example to use C# Generic feature in this case? 🤔 

Maybe like...:

```csharp
public readonly struct TypedAddress<T> : IBencodable;
public readonly AgentAddress : TypedAddress<SomeType>;
```

But C# doesn't allow to inherit struct type. 🤔

s2quake: Aha. My mistake, 😀

ipdae (approved):

PR #2723, opened 2024-07-31
Use IDelegatee, IDelegator on guild system

OnedgeLee:
# Purpose
- Implement delegation system that can be applied to guild system, and validator system
- Use implemented delegation system on guild system

# Changes
### Introduce interfaces and abstract / concrete classes for delegation
- `IDelegatee` : Interface for target of delegation.
- `IDelegator` : Interface for source of delegation.
- `Delegatee` : Abstract class for `IDelegatee`; Any model can be a delegatee by inherit this class.
- `Delegator`: Abstract class for` IDelegator`; Any model can be a delegator by inherit this class.
- `Bond` : Bonding information model.
- `UnbondLockIn` : Maintains information of locked-in FAVs from undelegation.
- `RebondGrace` : Maintains information of grace period from redelegation.
- `UnbondingSet` : Set that contains every UnbondLockIn & RebondGrace. This will call release on every unbonding contracts each block.
- `IDelegationRepository` : Repository that injected to `Delegatee` and `Delegator`. It defines required CRUDs for delegation relations.
- `DelegationRepository` : Concrete implementation of `IDelegationRepository`.

### Use Delegatee, Delegator on guild system
- `Guild` inherits `Delegatee`.
- `GuildParticipant` inherits `Delegator`.


Lib9c/Delegation/Bond.cs
```diff
@@ -0,0 +1,57 @@
...
+        public BigInteger Share { get; private set; }
+
+        public long LastDistributeHeight { get; private set; }
+
+        public IValue Bencoded => List.Empty
+            .Add(Share)
+            .Add(LastDistributeHeight);
+
+        public void AddShare(BigInteger share)
+        {
```
s2quake, line 43: Shouldn't the parameter be verified for negative numbers?

Lib9c/Delegation/RebondGrace.cs
```diff
@@ -0,0 +1,164 @@
...
+
+        public int MaxEntries { get; }
+
+        public bool IsFull => Entries.Values.Sum(e => e.Count) >= MaxEntries;
+
+        public bool IsEmpty => Entries.IsEmpty;
+
+        public ImmutableSortedDictionary<long, ImmutableList<RebondGraceEntry>> Entries { get; private set; }
+
+
```
s2quake, line 59: Hmm, two blank lines. 😀

Lib9c/Delegation/Delegation.cs
```diff
@@ -0,0 +1,94 @@
...
+
+        public RebondGrace RebondGrace { get; }
+
+        public UnbondingSet UnbondingSet { get; }
+
+        public FungibleAssetValue? IncompleteBond { get; private set; }
+
+        public FungibleAssetValue? IncompleteUnbond { get; private set; }
+
+        public void AddBond(FungibleAssetValue fav, BigInteger share)
```
s2quake, line 34: If this method is only used in the `Delegatee` class, I think that access-modifier of this method must be `internal`.

Lib9c/Delegation/Delegation.cs
```diff
@@ -0,0 +1,94 @@
+using System;
+using System.Numerics;
+using Libplanet.Crypto;
+using Libplanet.Types.Assets;
+
+namespace Nekoyume.Delegation
+{
+    public class Delegation
```
s2quake, line 8: If there are no properties or methods to override, I think it's a good idea to seal the class. 😀

Lib9c/Delegation/UnbondingSet.cs
```diff
@@ -0,0 +1,64 @@
+using System.Collections.Immutable;
+using System.Linq;
+using Bencodex.Types;
+using Libplanet.Crypto;
+
+namespace Nekoyume.Delegation
+{
+    public class UnbondingSet
```
s2quake, line 8: Is there a reason why this class is not IBencodable?

Lib9c/Delegation/Bond.cs
OnedgeLee: Added argument sanity check for this method!

Lib9c/Delegation/Delegation.cs
OnedgeLee: 👍

Lib9c/Model/Guild/Guild.cs
moreal, line 1: I'm preparing 1.16.0 release which will be released at 2024.08.07. So this change will be included in 1.17.0 release and it will have breaking-change because the `Guild`'s schema seems changed. As note, I should check this de/serialization part after releasing 1.16.0.

OnedgeLee: Okay, I'll convert it as a draft, and fix after 1.16.0 release.

OnedgeLee:
This changes will be applied with full DPoS features, Closing.

PR #2736, opened 2024-08-06
IssueToken allow non-tradable materials

ipdae:


Lib9c/Model/Item/Inventory.cs
```diff
@@ -442,7 +442,33 @@ public static bool IsMaterialRemovable(Item item, int id, long blockIndex)
...
+                if (material is TradableMaterial)
+                {
+                    return false;
+                }
+
+                return true;
+            }
+
+            return false;
+        }
```
moreal, line 468: It seems not to access any other methods, properties of `Inventory`. Shouldn't it be a static method and extracted from `Inventory` class?

Lib9c/Model/Item/Inventory.cs
```diff
@@ -442,7 +442,33 @@ public static bool IsMaterialRemovable(Item item, int id, long blockIndex)
...
+        /// <param name="id">The ID of the material to remove.</param>
+        /// <returns>Returns a boolean indicating whether the item is removable.</returns>
+        public static bool IsMaterialRemovable(Item item, int id)
+        {
+            if (item.Locked)
+            {
+                return false;
+            }
+
+            if (item.item is Material material && material.Id == id)
```
moreal, line 457: `material.Id == id` seems not related to *'is removable'*. I think it should be like the below example because `id` is defined in `ItemBase`, not only for `Material`.

```csharp
List<Item> targetItems = _items.Where(item => item.Id == id && IsRemovableNonTradableMaterial(item)).ToList();
```

Or more..

```csharp
bool IsNonTradableMaterial(ItemBase item) => item is Material && item is not TradableMaterial;
List<Item> targetItems = _items.Where(item => !item.Locked && item.item.Id == id && IsNonTradableMaterial(item.item)).ToList();
```

moreal (commented):
Just code review, not about bugs.

Lib9c/Model/Item/Inventory.cs
ipdae: Since it is already defined as a static method, relocating it should not cause any significant issues.

PR #2756, opened 2024-08-13
Fix reward logic

OnedgeLee:
SSIA


.Lib9c.Tests/Action/DPoS/RedelegateTest.cs
```diff
@@ -1,4 +1,4 @@
-﻿namespace Lib9c.Tests.Action.DPoS
+namespace Lib9c.Tests.Action.DPoS
```
limebell, line 1: What's the difference?

OnedgeLee: May be EOL type...

PR #2757, opened 2024-08-14
Check InsufficientBalance case in MigrateFee

ipdae:


Lib9c/Action/MigrateFee.cs
```diff
@@ -80,6 +80,11 @@ public override IWorld Execute(IActionContext context)
                 {
                     states = states.TransferAsset(context, sender, recipient, amount);
                 }
+                else
+                {
+                    throw new InsufficientBalanceException(
+                        $"required {amount} but {sender} balance is {balance}", sender, balance);
+                }
```
moreal, line 87: Hmm, if the `sender` doesn't have enough assets, it will occur `InsufficientBalanceException`

https://docs.libplanet.io/5.2.2/api/Libplanet.Action.State.IWorldExtensions.html#Libplanet_Action_State_IWorldExtensions_TransferAsset_Libplanet_Action_State_IWorld_Libplanet_Action_IActionContext_Libplanet_Crypto_Address_Libplanet_Crypto_Address_Libplanet_Types_Assets_FungibleAssetValue_

moreal (commented):
When I run `MigrateFeeTest` with `states = states.TransferAsset(context, sender, recipient, amount);` without condition, it passes the cases.

Lib9c/Action/MigrateFee.cs
ipdae: If the balance is insufficient after checking the balance above, TransferAsset is currently not executed at all. 
Due to performance considerations, we're pre-checking the balance. As a result, if the balance is insufficient, we skip calling TransferAsset altogether, which is why we've implemented separate handling for this scenario.

moreal: According to Libplanet team, they removes validation logics from Libplanet library and recommend to validate itself in application level. So these lines are needed.

moreal (approved):

PR #2685, opened 2024-07-15
Remove unused actions

moreal:
This pull request removes unused actions:

 - `transfer_asset`
 - `stake`
 - `activate_account2`
 - `add_activated_account2`

PR #2780, opened 2024-08-27
Refactor `AvatarState` constructor

moreal:
## Description

This pull request only adjusts the constructor of `AvatarState` and introduces a new static factory method, `AvatarState.Create`. With some huge changes (almost from test code) by replacing the `AvatarState` constructor with the factory method.

Before this pull request, it was hard to create an `AvatarState` instance because its constructor was heavily dependent on `AvatarSheets`. Even if I want to create a simple custom `AvatarState` instance, it requires `AvatarSheets` (CSV data).

So it will become to provide a simple constructor (just setting properties and fields) and handle something more with the factory method.

## Reviewers

I requested reviews for @ipdae, @sonohoshi, @eugene-doobu who are actively working on Lib9c. If you know someone else who would be good for this pull request, please nominate them.

## Future work plan

If this pull request is accepted, I'll refactor `WorldInformation` and `QuestList` too if I need. These works will be helpful for #2685 and https://github.com/planetarium/NineChronicles.Headless/pull/2540.

### DIFF
--- a/Lib9c/Model/State/AvatarState.cs
+++ b/Lib9c/Model/State/AvatarState.cs
@@ -68,10 +68,31 @@ namespace Nekoyume.Model.State
             return key.PublicKey.Address;
         }
 
-        public AvatarState(Address address,
+        public static AvatarState Create(Address address,
             Address agentAddress,
             long blockIndex,
             AvatarSheets avatarSheets,
+            Address rankingMapAddress,
+            string name = null)
+        {
+            var worldInformationVar = new WorldInformation(blockIndex, avatarSheets.WorldSheet,
+                GameConfig.IsEditor, name);
+            var questListVar = new QuestList(
+                avatarSheets.QuestSheet,
+                avatarSheets.QuestRewardSheet,
+                avatarSheets.QuestItemRewardSheet,
+                avatarSheets.EquipmentItemRecipeSheet,
+                avatarSheets.EquipmentItemSubRecipeSheet
+            );
+            return new AvatarState(
+                address, agentAddress, blockIndex, questListVar, worldInformationVar, rankingMapAddress, name);
+        }
+
+        public AvatarState(Address address,
+            Address agentAddress,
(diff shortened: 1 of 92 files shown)


ipdae (approved):
What is the scope of the update changes for relate repositories(NineChronicles, etc) that use AvatarState?

Lib9c/Model/State/AvatarState.cs
```diff
@@ -65,11 +65,32 @@ public static Address CreateAvatarAddress()
...
+                avatarSheets.QuestRewardSheet,
+                avatarSheets.QuestItemRewardSheet,
+                avatarSheets.EquipmentItemRecipeSheet,
+                avatarSheets.EquipmentItemSubRecipeSheet
+            );
+            return new AvatarState(
+                address, agentAddress, blockIndex, questListVar, worldInformationVar, rankingMapAddress, name);
+        }
+
+        public AvatarState(Address address,
```
eugene-doobu, line 88: What if we take the constructor private and depend only on "AvatarState.Create"?

moreal: The purpose of this pull request was to easily create an `AvatarState` instance without using `AvatarSheets`. We can change the constructor's access modifier to `internal` and allow to access to `internal` elements to the `Lib9c.Tests` project ([it already opened](https://github.com/planetarium/lib9c/blob/776c9113f7fd8f99b5645222d45b57dd7ad2cb82/Lib9c/AssemblyInfo.cs#L3C38-L3C43)), but it cannot open to `NIneChronicles.Headless.Tests` and other external test projects.

moreal:
> What is the scope of the update changes for relate repositories(NineChronicles, etc) that use AvatarState?

When I searched `planetarium/NineChronicles` (tool, tests), `planetarium/www.nine-chronicles.dev` (docs), `planetarium/NineChronicles.DataProvider` (tests)

PR #2797, opened 2024-08-29
Backmerge 1.17.3

U-lis:
Backmerge 1.17.3 to development
Idonno why conflict


moreal:
@U-lis Because I bumped `Lib9c` version to 1.17.2 in main branch, so it has conflicts. When releasing 1.17.3, it needs to bump projects' version to 1.17.3 to upload to NuGet successfully.

PR #2803, opened 2024-09-02
Test/exceptionformatter

sky1045:
In order to upgrade dotnet version from 6 to 8, we cannot use `BinaryFormatter` any more
this PR includes the replacement for ExceptionFormatter using `BinaryFormatter` currently
[ref](https://learn.microsoft.com/ko-kr/dotnet/core/compatibility/serialization/8.0/binaryformatter-disabled)


ipdae (changes requested):
Exception 직렬화 테스트가 실패하는것같습니다. 확인 부탁드립니다.

PR #2816, opened 2024-09-10
Use dotnet tool-manifest

moreal:
This pull request introduces dotnet [`tool-manifest`](https://learn.microsoft.com/en-us/dotnet/core/tools/local-tools-how-to-use) with some revisions. WIth tool-manifest, it can specify tool's version and manage it with version-control (Git).


boscohyun (approved):
Great!

PR #2818, opened 2024-09-10
Exclude BTAI from docs

moreal:
It excludes elements under `BTAI` namespace from DocFX docs. See https://dotnet.github.io/docfx/docs/dotnet-api-docs.html#filter-by-type for DocFX configuration.


boscohyun (approved):
👍

PR #2838, opened 2024-09-22
Introduce `IWorld.RemoveLegacyState` extension method

moreal:
This pull request introduces the `public static IWorld LegacyModule.RemoveLegacyState(this IWorld world, Address address)` method, which removes states instead of setting them to `Null`.

I replace `SetLegacyState(_, Null)` in `Stake`, with `RemoveLegacyState` as an example.

### DIFF
--- a/Lib9c/Action/Stake.cs
+++ b/Lib9c/Action/Stake.cs
@@ -153,7 +153,7 @@ namespace Nekoyume.Action
             if (Amount == 0)
             {
                 return states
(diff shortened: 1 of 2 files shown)


boscohyun (approved):

boscohyun:
This change will cause the `StakeState` deleted via the `Stake` action to be `null`, rather than of type `Bencodex.Types.Null`.

PR #2886, opened 2024-10-14
🔧 fix: pluggable ActionEvaluator

riemannulus:
- Add `IActionLoader` on `IPluginActionEvaluator`
- Add `IPolicyActionRegistry` on `IPluginActionEvaluator`
- Fix a bug when unmarshaling `ICommitedActionEvaulation`

@moreal  Please check 925f0e6. I've implemented it when we talked about it before. (use reflection)

### DIFF
--- a/.Lib9c.Plugin.Shared/IPluginActionEvaluator.cs
+++ b/.Lib9c.Plugin.Shared/IPluginActionEvaluator.cs
@@ -1,7 +1,14 @@
+using Libplanet.Action;
+using Libplanet.Action.Loader;
+
 namespace Lib9c.Plugin.Shared
 {
     public interface IPluginActionEvaluator
     {
         byte[][] Evaluate(byte[] blockBytes, byte[]? baseStateRootHashBytes);
+
+        IActionLoader ActionLoader { get; }
+
+        IPolicyActionsRegistry PolicyActionsRegistry { get; }
     }
 }
--- a/.Lib9c.Plugin/Lib9c.Plugin.csproj
+++ b/.Lib9c.Plugin/Lib9c.Plugin.csproj
@@ -15,5 +15,48 @@
       <ExcludeAssets>runtime</ExcludeAssets>
     </ProjectReference>
   </ItemGroup>
-  
+
+  <ItemGroup>
+    <ProjectReference Include="..\.Libplanet\src\Libplanet.Action\Libplanet.Action.csproj">
+      <Private>false</Private>
+      <ExcludeAssets>runtime</ExcludeAssets>
+    </ProjectReference>
+
+    <ProjectReference Include="..\.Libplanet\src\Libplanet\Libplanet.csproj">
+      <Private>false</Private>
+      <ExcludeAssets>runtime</ExcludeAssets>
+    </ProjectReference>
+    <ProjectReference Include="..\.Libplanet\src\Libplanet.Types\Libplanet.Types.csproj">
+      <Private>false</Private>
+      <ExcludeAssets>runtime</ExcludeAssets>
+    </ProjectReference>
+    <ProjectReference Include="..\.Libplanet\src\Libplanet.Common\Libplanet.Common.csproj">
+      <Private>false</Private>
+      <ExcludeAssets>runtime</ExcludeAssets>
+    </ProjectReference>
+
+    <ProjectReference Include="..\.Libplanet\src\Libplanet.Crypto\Libplanet.Crypto.csproj">
+      <Private>false</Private>
+      <ExcludeAssets>runtime</ExcludeAssets>
+    </ProjectReference>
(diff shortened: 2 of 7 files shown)


.Lib9c.Plugin.Shared/IPluginActionEvaluator.cs
```diff
@@ -1,7 +1,14 @@
...
+
 namespace Lib9c.Plugin.Shared
 {
     public interface IPluginActionEvaluator
     {
         byte[][] Evaluate(byte[] blockBytes, byte[]? baseStateRootHashBytes);
+
+        IActionLoader ActionLoader { get; }
+
+        IPolicyActionsRegistry PolicyActionsRegistry { get; }
```
moreal, line 12: I think the `IActionLoader` and `IPolicyActionsRegistry` types are too detailed types to include in the plugin system.

Please check the following types will never be changed? Or Libplanet will care type backward-compatibility for Lib9c pluggable system?

 - `IActionLoader`
 - `IPolicyActionsRegistry`
 - `IAction`
   - `IActionContext`
     - `IWorld`
       - `IWorldDelta`
       - `IAccount`
     - `ITransaction`
     - `EvidenceBase`
     - `IRandom`
     - `Address`
     - `TxId`

.Libplanet.Extensions.ActionEvaluatorCommonComponents/ActionEvaluationMarshaller.cs
```diff
@@ -42,4 +43,27 @@ public static ICommittedActionEvaluation Deserialize(byte[] serialized)
...
+    {
+        if (dictionary["exception"] is not Text typeName)
+        {
+            return null;
+        }
+
+        Type? type = Assembly.GetCallingAssembly().GetType(typeName.Value);
+        if (type is null)
+        {
+            throw new ArgumentException($"Type {typeName.Value} not found in the calling assembly.");
```
moreal, line 57: This line may need to return `null` or an alternative `Exception` type instance because if there is no type corresponding to `typeName`, it cannot unmarshal all `ActionEvaluation`, since `ActionEvaluationMarshaller.Unmarshal` will fail.

.Libplanet.Extensions.ActionEvaluatorCommonComponents/ActionEvaluationMarshaller.cs
```diff
@@ -42,4 +43,27 @@ public static ICommittedActionEvaluation Deserialize(byte[] serialized)
...
+        }
+
+        object? instance = Activator.CreateInstance(type);
+
+        if (instance is null)
+        {
+            throw new ArgumentException($"Failed to create an instance of {type}.");
```
moreal, line 64: Same with the above comment.

.Libplanet.Extensions.ActionEvaluatorCommonComponents/ActionEvaluationMarshaller.cs
```diff
@@ -42,4 +43,27 @@ public static ICommittedActionEvaluation Deserialize(byte[] serialized)
...
+            return null;
+        }
+
+        Type? type = Assembly.GetCallingAssembly().GetType(typeName.Value);
+        if (type is null)
+        {
+            throw new ArgumentException($"Type {typeName.Value} not found in the calling assembly.");
+        }
+
+        object? instance = Activator.CreateInstance(type);
```
moreal, line 60: It will be better to check the type is inherited `Exception`
```csharp
        if (!typeof(Exception).IsAssignableFrom(type))
        {
            // Do something.
        }

        object? instance = Activator.CreateInstance(type);
```

.Lib9c.Plugin/Lib9c.Plugin.csproj
```diff
@@ -15,5 +15,48 @@
...
+      <ExcludeAssets>runtime</ExcludeAssets>
+    </ProjectReference>
+    <ProjectReference Include="..\.Libplanet\src\Libplanet.Types\Libplanet.Types.csproj">
+      <Private>false</Private>
+      <ExcludeAssets>runtime</ExcludeAssets>
+    </ProjectReference>
+    <ProjectReference Include="..\.Libplanet\src\Libplanet.Common\Libplanet.Common.csproj">
+      <Private>false</Private>
+      <ExcludeAssets>runtime</ExcludeAssets>
+    </ProjectReference>
```
moreal, line 36: These assemblies need to be included because it will try to use shared types if these assemblies are excluded and it will cause a type mismatch error. If not, in my opinion, Libplanet must guarantee type backward-compatibility by SemVer strategy, and Lib9c have to be able to ignore new Libplanet major updates. (no bump)

.Lib9c.Plugin.Shared/IPluginActionEvaluator.cs
moreal, line 12: Why these fields become needed?

PR #2887, opened 2024-10-14
feat: implement `Addresses.GetAccountAddress(currency)` method

boscohyun:
SSIA

### DIFF
--- a/Lib9c/Addresses.cs
+++ b/Lib9c/Addresses.cs
@@ -4,6 +4,7 @@ using System.Linq;
 using System.Security.Cryptography;
 using Libplanet.Common;
 using Libplanet.Crypto;
+using Libplanet.Types.Assets;
 using Nekoyume.Action;
 using Nekoyume.Model.State;
 using Nekoyume.TableData;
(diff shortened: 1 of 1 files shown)


moreal (commented):
What is the expected use case in Lib9c project for the new method?

boscohyun:
> What is the expected use case in Lib9c project for the new method?

I was expecting it to be used by those using `Lib9c`, and thought it would be easier to simply provide the account address for the call.

moreal:
> I was expecting it to be used by those using `Lib9c`, and thought it would be easier to simply provide the account address for the call.

I understand your intention, but I was wondering who exactly will be using the API. In another thought, I think it should be provided by the "Libplanet" library. How about suggesting this to them?

boscohyun:
> I think it should be provided by the "Libplanet" library. How about suggesting this to them?

I agree. Hmm.. I'll request this to libplanet.👍

boscohyun:
Replace with this issue:
https://github.com/planetarium/libplanet/issues/3987

PR #2923, opened 2024-10-22
fix try parse WorldBossRankingRewardSheet

tyrosine1153:
- 0 또는 null같은 빈 데이터가 들어갈 경우 Material List에서 제외되도록 수정

### DIFF
--- a/.Lib9c.Tests/TableData/WorldBossRankingRewardSheetTest.cs
+++ b/.Lib9c.Tests/TableData/WorldBossRankingRewardSheetTest.cs
@@ -81,6 +81,7 @@ namespace Lib9c.Tests.TableData
             var rewards = row.GetRewards(tableSheets.RuneSheet, tableSheets.MaterialItemSheet);
 
             Assert.Equal(expected, rewards.Count);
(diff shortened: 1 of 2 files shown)


U-lis (changes requested):
I think we should test `sheet.rows.Count` to clarify 17 and 18 rows are ignored.

.Lib9c.Tests/TableData/WorldBossRankingRewardSheetTest.cs
```diff
@@ -68,16 +68,19 @@ public void FIndRow_Throw_ArgumentException()
             Assert.Throws<ArgumentException>(() => sheet.FindRow(900003, 0, 0));
         }
 
-        [Fact]
-        public void GetRewards()
+        [Theory]
+        [InlineData(900001, 1, 0, 6)]
+        [InlineData(900002, 600, 60, 5)]
+        [InlineData(900002, 800, 80, 4)]
```
tyrosine1153, line 74: @U-lis  I added new test cases for 17, 18 rows

PR #3013, opened 2024-11-18
Fix stake without guild not to send guild gold

OnedgeLee:
Stake without sent guild gold to `NonValidatorDelegatee`,
but this makes impossible to use guild gold later.
So, fixed not to send guild gold to `NonValidatorDelegatee`.


Lib9c/Action/Stake.cs
s2quake, line 1: Line: 240
```csharp
         state = state.BurnAsset(context, Addresses.NonValidatorDelegatee, gg);
```
Is the above code necessary? If so, I think it should have an if condition.
For example, if gg is 0, or if BurnAssets throws an exception, or if ...

s2quake (approved):

PR #3186, opened 2025-04-22
Resolve Battle.cs not using ToDotnetString for the memo field.

FioX0:
Fixes the issue where GQL Battle Actions get their MEMO loaded into the wrong format. Due to an invalid memo format, all GQL Battle Actions are unreadable by the Arena Endpoints.

Before the change:

prefix

Post the change:
postfix


ipdae (approved):
LGTM

Atralupus (approved):

PR #3288, opened 2026-05-19
Preserve original exception type in ActionEvaluationMarshaller

ipdae:
## Summary

When `ActionEvaluationMarshaller.Unmarshal` rebuilt a deserialized
`ICommittedActionEvaluation`, it constructed `new Exception(typeName)`,
which collapsed the original action-exception type into a generic
`System.Exception`. The original `FullName` survived only as the
`Message` of that generic exception.

Downstream this caused [`Libplanet.Types.Tx.TxExecution`](https://github.com/planetarium/libplanet/blob/main/src/Libplanet.Types/Tx/TxExecution.cs)
— which derives exception names via `e.GetType().FullName` — to record
every PAEV-evaluated failure as `"System.Exception"`. That value is then
exposed through NineChronicles.Headless's `transactionResult.exceptionNames`
GraphQL field, where users observed every failure showing up as
`System.Exception` regardless of the real action exception.

The fix resolves the recorded `FullName` against the host's loaded
assemblies and instantiates the matching `Exception` subtype with
`FormatterServices.GetUninitializedObject`, so `GetType().FullName`
round-trips correctly. The serialization side is unchanged, so this is
backward compatible with existing stored data.

When a type cannot be resolved (e.g. an exception type not loaded in
the host process), it falls back to the previous behavior of returning
a generic `Exception` carrying the type name as its message.

## Test plan

- [x] `dotnet test .Lib9c.ActionEvaluatorCommonComponents.Tests/` — 4 passed
- [x] Added `ExceptionTypeNamePreservedAfterRoundTrip` theory covering
  `InvalidOperationException`, `ArgumentException`, and the action-specific
  `UnexpectedlyTerminatedActionException`
- [ ] Verify on a headless build using PAEV that `transactionResult.exceptionNames`
  now returns the real action exception name (e.g. `Nekoyume.Action.InvalidAddressException`)
  instead of `System.Exception` for a known-failing transaction

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ipdae:
Follow-up: added a second commit ([a129628](https://github.com/planetarium/lib9c/commit/a12962881b)) to also serialize the `InnerException` chain.

Without this, the first commit alone correctly preserved the outer exception's type name, but in the real PAEV path the outer is always `Libplanet.Action.UnexpectedlyTerminatedActionException` wrapping the actual action exception. Libplanet's `TxExecution` constructor unwraps one level (`InnerException ?? exception`), and since our deserialized outer had a null inner the wrapper leaked through to `transactionResult.exceptionNames`.

This commit:
- `Marshal` now writes the outer→inner chain as a Bencodex `List` of type names.
- `Unmarshal` reconstructs the chain and writes each inner into `Exception._innerException` via reflection.
- Falls back to the legacy single-`Text` format on read so a new headless can still parse messages from an older plugin DLL.

Verified on heimdall-rpc-2 reproduction: a newly failed tx surfaced `Libplanet.Action.UnexpectedlyTerminatedActionException` after only the first commit was deployed; expected behavior after this commit is the real inner action exception name.

Tests added: `InnerExceptionChainPreservedAfterRoundTrip`, `UnmarshalBackwardCompatibleWithLegacyTextFormat` (6/6 passing in `Lib9c.ActionEvaluatorCommonComponents.Tests`).

ipdae:
recheck

PR #3299, opened 2026-08-27
feat: adjust market and synthesis restrictions by sheet (PLD-1483)

ipdae:
## Why

Two restrictions are code constants today:

- `RegisterProduct.NonTradableTickerCurrencies` — five currencies that may not be listed on the market
- `Synthesize.InvalidMaterialItemId` — seven item ids that may not be used as a synthesis material

Adding one entry takes a lib9c release, a headless redeploy and a client release, because the
client references `InvalidMaterialItemId` directly from its bundled lib9c. The list has been
amended twice since it was introduced in 2024-12 (most recently PLD-731), and an incident found
today cannot be contained today.

This adds a sheet that `patch_table_sheet` can adjust, and applies it at the points that decide
whether an item or a currency may leave its owner.

Closes PLD-1483 (making the non-tradable rune list adjustable by sheet); the same sheet covers
the item axis, which is what the portal web shop needs for account bound costumes.

## The sheet

```
RestrictionSheet.csv
key,market_registrable,synthesize_material
10660004,,false
CRYSTAL,false,
```

One key column spells an item id for items and a ticker for currencies. The call site already
knows which one it holds — the market action has an item id or a ticker — so no discriminator
column is needed.

Parsing is deliberately split in two:

- **Runtime (`Row.Set`) is lenient.** `Sheet<TKey, TValue>` does not guard this call, so a row
  that throws aborts the whole sheet and takes every restriction with it. An empty cell is
  unspecified, an unparsable cell restricts (a typo must fail toward blocking, which is
  recoverable, not toward leaking), and duplicate keys merge toward the restrictive value.
- **`ValidateCsv` is strict** and runs inside `PatchTableSheet`, so a malformed policy fails the
  transaction instead of landing on the chain. It rejects a key no lookup can ever match
  (`010660004`, `crystal`, a Cyrillic homoglyph) — such a row would sit there silently and read
  as "unrestricted".

`ValidateCsv` mirrors `Sheet.Set`'s grammar, including comment rows and `_` prefixed columns.
The two disagreeing is the dangerous failure, so their agreement is pinned by tests.

## Where it is applied

| Point | What it decides |
| --- | --- |
| `RegisterProduct.Register` | item and fungible asset registration. `RegisterProduct` and `ReRegisterProduct` are the only market registration actions still live — the rest are `[ActionObsolete]` — and both reach the market through this one static, including `ReRegisterProduct`'s legacy order branch |
| `Synthesize` | material eligibility, on top of `InvalidMaterialItemId` |
| `PatchTableSheet`, `PatchTableSheetCompressed` | validate the CSV before writing it |

## Design decisions

**The sheet adds to the hardcoded lists, it does not replace them.** A patched sheet that forgets
an entry must not lift a restriction the code already enforces. The cost is that lifting one of
the twelve existing entries still takes a release. Note the two are not enforced side by side:
`InvalidMaterialItemId` is checked next to the sheet inside `Synthesize`, but
`NonTradableTickerCurrencies` still lives in `AssetInfo.Validate()`, which is the caller's
responsibility to invoke. Moving it beside the ticker check is worth a follow-up.

**A patch stops new registrations; it does not retract listings that are already up.**
`BuyProduct` does not consult the sheet, so restricting an item after the fact leaves existing
products on sale until they are cancelled. Restricting something already in circulation therefore
needs a scan for surviving products, not just a patch.

**Loading uses a new `LegacyModule.TryGetPatchedSheet`.** The existing `TryGetSheet` catches every
exception, so one typo would report a malformed policy as absent and switch every restriction
off. Only genuine absence may fall back.

**Absence is the pre-patch behavior, exactly.** No new throw fires while the sheet is missing, in
any of the four actions. That is what keeps re-evaluation of blocks older than the patch
identical, and it is why the sheet is loaded optionally rather than as a required sheet — a
required load would fail every affected action on a chain that has not been patched yet.

## Also fixes: item currency products

An item currency (`Item_*`, or `FAV__Item_*` once wrapped) can be registered as a fungible asset
product today. `BuyProduct` sends the asset to the buyer's **avatar** address, and an item currency has no way
out of one: `ClaimItems` burns from the signer, and `TransferAsset` requires `Sender == signer`,
which an avatar address can never be. So the buyer pays NCG for a balance they can never spend.
(Other currencies do leave an avatar address — rune and soulstone actions burn from it, and
market registration itself moves from it — this is specific to item currencies.)

A wrapped ticker is unwrapped before both this check and the sheet lookup, because wrapping does
not make it a different asset: `claim_items` unwraps it straight back. Without that, a `CRYSTAL`
row would be walked past by `FAV__CRYSTAL`, which is reachable — CRYSTAL has a row in
`LoadIntoMyGaragesCostSheet`, so it can go through a garage and come back wrapped.

The ticker exists per item id and cannot be enumerated as sheet rows, so the rule lives in code
with the sheet's presence as its switch, which keeps older blocks unaffected.

## Also changes: what an unlisted item weighs in synthesis

`SynthesizeWeightSheet` returned `DefaultWeight = 10000` for an item it does not list, so adding a
costume to `CostumeItemSheet` dropped it into every result pool of its grade and sub type
immediately, and it stayed there until someone remembered to write a 0. That is how four grade 8
costumes were handed out in 2026-08. The default is now 0: a pool is what the sheet spells out.

**This needs the sheet backfilled, which is in this PR.** The sheet only listed exceptions and let
everything else ride on the implicit 10000, so flipping the default alone would leave
`FullCostume` grades 2–6 and 8, every `Title` grade, `Grimoire` 7–8 and `Aura` 8 with a total
weight of 0 — synthesis into them would have nothing to draw. `Title` had none of its 53 items
listed. Every item that can be drawn today is written out with an explicit 10000 (108 rows), and
every pool's total weight is unchanged, so the odds for existing items stay exactly as they are.
A test asserts no synthesizable pool is left without weight.

The draw also moves from `random.Next(totalWeight + 1)` to `random.Next(totalWeight)`. The
inclusive bound could land on `totalWeight`, walk past every item and throw, at a rate of
1/(total+1) — small while pools carried implicit weights, not small once a pool is only what the
sheet lists. A pool with no weight at all now throws saying so.

This changes evaluation, so it belongs behind a fork boundary: add a `ForkableActionEvaluator`
entry pinning the current evaluator up to the deploy block, with the new lib9c from the next one.
The sheet backfill must be patched **before** that deploy — it is a no-op under the current code,
and skipping the order leaves a window where synthesis is dead.

## Testing

- `dotnet test .Lib9c.Tests` — 3,753 passed, 0 failed, 6 skipped (pre-existing skips)
- New coverage: sheet parsing and its agreement with `ValidateCsv` (including `_` prefixed
  columns), parity with both hardcoded lists (so removing a constant later breaks a test), the
  shipped CSV passing its own validator, absence keeping the old behavior in every applied
  action, an unreadable sheet not passing silently, wrapped and unwrapped item currency tickers,
  and compressed patch validation.

## Follow-ups (not in this PR)

- Client: `ShopSell`, `Inventory`, `ItemTooltipDetail`, `Synthesis`, `SynthesisInventory` decide
  sellability and material eligibility on their own, so a restricted item still shows the button
  and fails at execution until they read the sheet.
- CI: a set-difference check across `CostumeItemSheet` / `SynthesizeWeightSheet` /
  `CostumeSummonSheet` / `RestrictionSheet`, since a new bound costume needs an entry in three
  of them and a missing key is the one typo `ValidateCsv` cannot catch.
- The garage and tokenization paths are untouched, so a row here does not make a **fungible**
  item account bound by itself. The sheet doc says so explicitly.

## Deployment order

1. Roll out to every node first, and confirm the rollout actually took: a headless image bump
   alone can leave the previous lib9c plugin DLL in the node's cached plugin directory. Once a
   patch lands, an older node diverges on ordinary traffic — it lets a restricted registration
   through while a new node throws — and on a malformed CSV it diverges on block validity itself.
2. Patch the policy sheet **before** the item sheet when adding a restricted item, so there is no
   window where the item exists without its policy. `PatchTableSheet.TableName` is scalar, so the
   two cannot travel in one transaction.
3. Both Odin and Heimdall.

Before the first patch, two things are worth checking because they cannot be undone by another
patch. Any `Item_*` balance sitting at an avatar address loses its only exit when item currency
products are refused, so count those first. And restricting an item id applies to every copy in
existence, including ones players bought earlier — a restricted item should be issued under a new
id rather than an id already in circulation.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ipdae:
## 배포 전 준비 목록 (머지 시점에는 불필요, 릴리즈에 실릴 때 필요)

이 PR의 마지막 두 커밋(`SynthesizeWeightSheet` 기본 weight 10000 → 0)은 **조건 없는 평가 변경**입니다. 나머지 커밋은 `RestrictionSheet`가 체인에 없으면 이전과 완전히 동일하게 동작하므로 언제 배포해도 무해하지만, 기본값 전환은 아래가 갖춰지기 전에 릴리즈에 실리면 노드가 갈라집니다.

### 1. 시트 백필을 상류 스프레드시트에 반영

`Lib9c/TableCSV/SynthesizeWeightSheet.csv`에 54행(`<itemId>,10000`)을 추가했습니다. 이 파일은 기획이 Google Sheets에서 내보내는 산출물이므로(직전 커밋들이 `Update ... from Google Sheets`), **54행이 상류에 없으면 다음 export가 이 변경을 지웁니다.** 기본값이 0인 상태에서 그 export가 들어오면 합성 결과 풀 전체가 총 weight 0이 되어 합성이 전면 실패합니다.

`EveryReachableResultPoolHasWeight` 테스트가 PR 단계에서 막아주지만, 그것은 파이프라인이 멈추는 것이지 데이터가 고쳐지는 것은 아닙니다.

### 2. `40100060` / `40100061` 확인

두 아이템(Heimdall/Odin 26 2nd Champion)은 weight 시트에 없어 암묵 기본값 10000으로 뽑히고 있었습니다. 백필이 이를 **명시 10000으로 고정**합니다. 현재 동작을 그대로 옮긴 것이지만, 0이어야 할 값이라면 patch 전에 정정해야 합니다(온체인 patch 후 되돌리는 비용이 큽니다).

### 3. fork 경계 배선

현재 `charts/all-in-one/scripts/common/appsettings.json`에 `Headless:ActionEvaluator` 키가 없어 모든 validator/RPC가 `ActionEvaluatorType.Default`(이미지에 컴파일된 in-process lib9c)로 동작합니다. 즉 fork 항목 추가는 한 줄 추가가 아니라 **평가기 종류 변경**입니다:

- `ForkableActionEvaluator` + 구간별 `PluggedActionEvaluator`
- 구/신 lib9c 플러그인 zip 2종 발행, **URL은 버전 핀**(`latest` 류는 `/data/plugins/` 캐시 때문에 구 DLL로 조용히 평가됨)
- odin / heimdall 전 노드에 적용

`charts/all-in-one/templates/remote-action-evaluator-headless.yaml`은 참고용으로 쓸 수 없습니다 — `replicas: 0`이고 `"type": "RemoteActionEvaluator"`는 enum(`Default` / `ForkableActionEvaluator` / `PluggedActionEvaluator`)에 없는 값입니다.

### 4. 순서 불변식

```
시트 백필 patch_table_sheet  <  forkStart
```

fork 시작 인덱스가 백필 patch 블록보다 앞이면, 신 평가기가 54행이 없던 구간을 재평가하면서 총 weight 0으로 예외를 던집니다 → 과거에 성공했던 synthesize 트랜잭션이 실패 → state root 불일치. 제네시스 동기, 리오그, DataProvider 재평가 모두 해당됩니다.

백필 patch 자체는 현행 코드에서 무해합니다(명시 10000 = 기존 암묵 기본값).

### 5. 클라이언트

- `ActionRenderHandler`가 `LocalRandom(eval.RandomSeed)`으로 결과를 재시뮬레이션해 결과창에 띄웁니다. 서브모듈이 구 lib9c에 핀돼 있으면 새 추첨 상한(`Next(totalWeight)`)과 어긋나 버킷 경계에서 결과창과 실제 획득이 달라질 수 있습니다.
- `Synthesis.GetSynthesizeResultPool` 프리뷰가 "미등재 = 포함" 규칙을 자체 재구현하고 있어, 앞으로 weight 없이 아이템만 추가되면 체인은 제외하고 클라는 노출합니다. `SynthesizeSimulator.GetWeight`를 쓰도록 고치는 편이 좋습니다.

### 6. `RestrictionSheet` 쪽 게이트 (첫 patch 전, 되돌릴 수 없음)

- 아바타 주소에 `Item_*` / `FAV__Item_*` 잔고를 보유한 계정 실측 — 아이템 통화 상품 등록이 막히면 그 잔고의 유일한 출구가 닫혀 영구 동결됩니다.
- 지급 레일이 `Item_NT_<itemId>` 토큰을 **유저 주소에 앉히지 않는지** 확인. `ClaimItems`에 서명자–아바타 소유 검증이 없어, 토큰이 유저 주소에 놓이면 타인에게 송금하거나 타인 아바타로 claim할 수 있고 시트로는 막을 수 없습니다.
- 시트 patch는 **신규 등록만** 막습니다. `BuyProduct`는 이 시트를 보지 않으므로, 이미 등록된 상품은 취소되기 전까지 계속 팔립니다.
- 제한 아이템에도 클라의 판매/합성 버튼은 그대로 노출되고 트랜잭션에서 실패합니다(`InvalidItemIdException`은 이 액션에서 처음 발생하는 타입). CS 사전 공유가 필요합니다.

### 후속 PR 후보

- `RegisterProduct.NonTradableTickerCurrencies` 검사를 `AssetInfo.Validate()`에서 `Register`의 티커 검사 옆으로 이동 (지금은 두 정책이 다른 지점에서 강제됨)
- `BuyProduct`에도 정책 게이트 (사후 봉쇄용)
- `CostumeItemSheet` ↔ `SynthesizeWeightSheet` ↔ `CostumeSummonSheet` ↔ `RestrictionSheet` 교차 검사 CI
- `SynthesizeWeightSheet.Row.Set`을 전함수로 (지금은 빈 줄 하나로 시트 전체 로드가 실패합니다 — 이 PR 작업 중 실제로 발생)

ipdae:
recheck

